Issue #9, opened 2020-02-06
Coverage isn't collected or reported to codecov


exarkun:
The test suite jobs are running the `testenv` environment and not the `coverage` environment.  No coverage is collected or reported by them.

PR #10, opened 2020-02-06
Try smashing coverage into testenv

exarkun:
Fixes: #9


exarkun:
The NixOS failure is related to an Eliot issue via Tahoe-LAFS.  There's a fix in Tahoe-LAFS waiting to be landed.  So not blocking this on that issue.  Everything else looks good, merging.

Issue #11, opened 2020-02-07
Coverage is only collected from allmydata package


exarkun:
Our code is in the magic_folder package, though.

PR #13, opened 2020-02-07
Correct coverage collection

exarkun:
Fixes: #11


exarkun:
The Eliot failure on the nixos job is being address by Tahoe-LAFS upstream.

PR #31, opened 2020-02-13
Rewrite list tests

### LINKED ISSUE #30: Eliminate no_network from magic_folder.test.cli.test_magic_folder.ListMagicFolder
Child of #22

exarkun:
Fixes #30

### DIFF
--- a/src/magic_folder/frontends/magic_folder.py
+++ b/src/magic_folder/frontends/magic_folder.py
@@ -328,10 +328,6 @@ def save_magic_folders(node_directory, folders):
         yamlutil.safe_dump({u"magic-folders": folders}),
     )
 
-    config = configutil.get_config(os.path.join(node_directory, u"tahoe.cfg"))
-    configutil.set_config(config, "magic_folder", "enabled", "True")
-    configutil.write_config(os.path.join(node_directory, u"tahoe.cfg"), config)
-
 
 class MagicFolder(service.MultiService):
 
--- a/src/magic_folder/scripts/magic_folder_cli.py
+++ b/src/magic_folder/scripts/magic_folder_cli.py
@@ -3,8 +3,10 @@ from __future__ import print_function
 import os
 import sys
 import urllib
+from six.moves import (
+    StringIO as MixedIO,
+)
 from types import NoneType
-from six.moves import cStringIO as StringIO
 from datetime import datetime
 from ConfigParser import SafeConfigParser
 import json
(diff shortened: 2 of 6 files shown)


exarkun:
Did this actually end up as a win?  It's certainly a lot more lines. :(  But the line count change misses the fact that these tests depend on a lot less code from Tahoe-LAFS now...

I really wanted to avoid having any subprocesses at all here but it's not possible to *really* create a new magic folder without Tahoe-LAFS storage capabilities.  The tests could directly mash `magic_folder.yaml` and perhaps test some things without a running Tahoe-LAFS node ... but I think that might go against the spirit of _these_ tests.  Perhaps there is room to exercise more cases in a different test suite that operates directly on `magic_folder.yaml` instead of a running Tahoe-LAFS node?

src/magic_folder/frontends/magic_folder.py
```diff
@@ -328,10 +328,6 @@ def save_magic_folders(node_directory, folders):
         yamlutil.safe_dump({u"magic-folders": folders}),
     )
 
-    config = configutil.get_config(os.path.join(node_directory, u"tahoe.cfg"))
-    configutil.set_config(config, "magic_folder", "enabled", "True")
-    configutil.write_config(os.path.join(node_directory, u"tahoe.cfg"), config)
-
```
exarkun: The `magic-folder list` tests invoke `magic-folder create` and other mutate-y commands.  Those commands all want to upgrade and rewrite the magic-folder configuration.  Previously this made sense but now the `[magic_folder]enabled` setting in `tahoe.cfg` is worse than useless.  We don't want to enable Tahoe-LAFS' built-in magic-folders support if we're running the stand-alone magic-folder.  So, the stand-alone magic-folder should never rewrite the Tahoe-LAFS config to turn on this toggle.

src/magic_folder/scripts/magic_folder_cli.py
```diff
@@ -3,8 +3,10 @@
 import os
 import sys
 import urllib
+from six.moves import (
+    StringIO as MixedIO,
+)
```
exarkun: Should not be a behavior change but should be more clear to readers.  This is not a byte stream.  This is not a text stream.  This is a whatever-you-feel-like stream.  This is bad and should be fixed but the implementation makes no consistent decision about whether it wants to write out bytes or text  and I didn't want to fix every string literal in the whole project in this PR.

src/magic_folder/scripts/magic_folder_cli.py
```diff
@@ -1059,7 +1061,7 @@ def do_magic_folder(options):
     try:
         return f(so)
     except Exception as e:
-        print("Error: %s" % (e,), file=options.stderr)
+        print(u"Error: {}".format(e), file=options.stderr)
```
exarkun: I did get tricked into *starting* to change every string literal in the whole project before I realized that was a trap, though...

src/magic_folder/test/cli/test_magic_folder.py
```diff
@@ -290,48 +316,162 @@ def get_Bob_magicfolder(result):
         return d
 
 
-class ListMagicFolder(MagicFolderCLITestMixin, AsyncTestCase):
-
+@defer.inlineCallbacks
+def create_introducer(testcase, introducer_directory):
+    """
+    Make an run a Tahoe-LAFS introducer node.
```
vu3rdd: Typo here? "Make a Tahoe-LAFS introducer node? ?

PR #32, opened 2020-02-14
.circleci/run-tests.sh: double quote shell variables

### LINKED ISSUE #136: The CircleCI configuration includes some incorrectly unquoted shell variable expansions
`.circleci/run-tests.sh` expands `BOOTSTRAP_VENV` and `PROJECT_ROOT` without quoting.  If these paths have specials in them, stuff will break.

vu3rdd:
Fixes a few a few minor missing double quotes reported by shellcheck.

Fixes #136


.circleci/run-tests.sh
```diff
@@ -74,14 +74,14 @@ else
...
 fi
 
-${TIMEOUT} ${BOOTSTRAP_VENV}/bin/tox \
-    -c ${PROJECT_ROOT}/tox.ini \
+${TIMEOUT} "${BOOTSTRAP_VENV}"/bin/tox \
+    -c "${PROJECT_ROOT}"/tox.ini \
     --workdir /tmp/magic-folder.tox \
     -e "${MAGIC_FOLDER_TOX_ENVIRONMENT}" \
-    ${MAGIC_FOLDER_TOX_ARGS} || "${alternative}"
+    "${MAGIC_FOLDER_TOX_ARGS}" || "${alternative}"
```
exarkun: This quoting change is the only one I'm concerned about here.  I think that the point was to be able to pass multiple positional arguments to tox here - perhaps mostly for pass-through.  For example, `["--", "--dry-run", "magic_folder.tests.cli"]`.  I think the added quoting turns these all into a single argument in tox's argv and I don't know what the consequences of that are.

Maybe it's fine because the single string then expands to multiple strings when tox uses it to construct a shell command?  Because I don't think there's consistent quoting inside tox.ini for handling these...  But maybe there should be?

Is there a better way to handle multiple arguments in shell than this?  Should it be a shell array instead?

exarkun:
> Thanks. I like the idea of running a shell lint tool over all the shell as part of CI. In practice I don't recall ever using this tool so I'm not sure how often the suggestions are right or wrong. Do you have a sense of whether this is usable in a CI configuration?

I looked into this further.  shellcheck is mad at a bunch of our correctly unquoted shell expanses so I guess we shouldn't add it to CI.  Too bad.  Maybe there's a different way to write some of our shell to avoid this problem, I dunno, but I guess we won't try to deal with this right now.

PR #37, opened 2020-02-19
Rewrite status

### LINKED ISSUE #33: Eliminate no_network from magic_folder.test.cli.test_magic_folder.StatusMagicFolder
Child of #22

exarkun:
Fixes #33

### DIFF
--- a/src/magic_folder/web/magic_folder.py
+++ b/src/magic_folder/web/magic_folder.py
@@ -1,9 +1,6 @@
 import json
 import cgi
 
-from twisted.internet.endpoints import (
-    serverFromString,
-)
 from twisted.application.internet import (
     StreamServerEndpointService,
 )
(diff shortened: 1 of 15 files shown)


src/magic_folder/web/magic_folder.py
```diff
@@ -72,42 +77,41 @@ def __init__(self, get_magic_folder, get_auth_token):
...
+            data.append(status_for_item("download", item))
 
         return json.dumps(data)
+
+
+def status_for_item(kind, item):
+    d = dict(
+        path=item.relpath_u,
+        status=item.status_history()[-1][0],
+        kind='upload',
```
vu3rdd: Don't we need to assign `kind` to the passed parameter `kind` instead of `upload`?

exarkun: Oops!  Yes, thanks.  Nice catch.

PR #44, opened 2020-02-21
Select the "ci" Hypothesis profile on GitHubActions macOS CI

### LINKED ISSUE #43: Sometimes Hypothesis tests fail on CI because of the Hypothesis deadline
The Hypothesis deadline: https://hypothesis.readthedocs.io/en/latest/settings.html#hypothesis.settings.deadline

The CI execution environment is shared.  We can't make any guarantees about how long our tests will take run (more generally, none of Linux, macOS, and Windows are real-time operating systems and Python is not a real-time programming language so it's hard to argue that you can ever guarantee how long a test will take to run in any supported environment, but I digress...).

Fortunately the Hypothesis deadline can be disabled.  Do this for all CI.

exarkun:
Fixes #43

### DIFF
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -29,6 +29,8 @@ jobs:
       run: |
         pip wheel --wheel-dir ./wheelhouse git+https://github.com/tahoe-lafs/tahoe-lafs@master#egg=tahoe-lafs
     - name: Test with tox
+      env:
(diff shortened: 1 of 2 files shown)


exarkun:
I'm not convinced this is actually a reliable solution but after some effort I have not yet managed to reproduce the problems that make me skeptical of it.

At minimum, it seems like a *step* in the right direction.  Thus, I will merge this and if the problems continue to arise I will investigate further.

Issue #6, opened 2020-02-06
Re-arrange the repository to make more sense as a stand-alone


exarkun:
The repository structure continues to reflect the structure of the magic-folders implementation as it existed inside Tahoe-LAFS.  Without all of the rest of Tahoe-LAFS around it, this structure makes less sense.  Consolidate it somewhat, removing redundancies that exist in the path hierarchies and the extremely sparse directories which now contain only one or two files.

Files / modules which could use a better name (suggested "better name" too, but maybe there's even-better names):

 - [x] `magic_folder/scripts/magic_folder_cli.py` -> `magic_folder/cli.py` (https://github.com/LeastAuthority/magic-folder/pull/49)
 - [x] `magic_folder/frontends/magic_folder.py` -> `magic_folder/magic_folder.py` (https://github.com/LeastAuthority/magic-folder/pull/70)
 - [x] `magic_folder/web/magic_folder.py` -> .`magic_folder/web.py` (https://github.com/LeastAuthority/magic-folder/pull/69)
 - [ ] others?

tomprince:
Anything left to do here?

meejah:
Would be nice to have only one `test_magic_folder.py` and to flatten the `docs` stuff more, but also that's just "nice to have" and we could file different tickets later (so i'll just close this)

PR #49, opened 2020-03-10
move scripts/magic_folder_cli to magic_folder/cli

meejah:
started a bit of re-org while getting up to speed on "stand-alone magic-folder". It did occur to me that maybe re-orgs are disruptive ... but this one is at least "just one file"...? (Anyway, we could hold off on this if that seems like a good idea).

https://github.com/LeastAuthority/magic-folder/issues/6

### DIFF
--- a/src/magic_folder/cli.py
+++ b/src/magic_folder/cli.py
@@ -118,21 +118,21 @@ from allmydata.client import (
     read_config,
 )
 
-from magic_folder.frontends.magic_folder import (
+from .frontends.magic_folder import (
     MagicFolder,
     load_magic_folders,
     save_magic_folders,
     maybe_upgrade_magic_folders,
 )
-from web.magic_folder import (
+from .web.magic_folder import (
     magic_folder_web_service,
 )
 
-from status import (
+from .status import (
     status as _status,
 )
 
-from _coverage import (
+from ._coverage import (
     coverage_service,
 )
 
--- a/src/magic_folder/cli.py
+++ b/src/magic_folder/cli.py
@@ -118,21 +118,21 @@ from allmydata.client import (
     read_config,
 )
 
-from ..frontends.magic_folder import (
+from magic_folder.frontends.magic_folder import (
     MagicFolder,
     load_magic_folders,
     save_magic_folders,
     maybe_upgrade_magic_folders,
 )
-from ..web.magic_folder import (
+from web.magic_folder import (
     magic_folder_web_service,
 )
 
-from ..status import (
+from status import (
     status as _status,
 )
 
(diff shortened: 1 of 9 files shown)


meejah:
@exarkun this could be merged -- unless you think its better to do all re-org at once?

exarkun:
Incremental improvement is great.   Maybe #6 should be updated to list the files that need to move around and then we can watch the progress we make towards being all-the-way done with that.

I'm gonna leave a couple comments inline here, but apart from those I think this should get merged.

src/magic_folder/cli.py
```diff
@@ -118,21 +118,21 @@
     read_config,
 )
 
-from ..frontends.magic_folder import (
+from magic_folder.frontends.magic_folder import (
```
exarkun: Can this be `from .frontends...`?  I'm not dead set on this.  I don't think there's a clear winner between absolute and explicit relative imports in the general case.

src/magic_folder/cli.py
```diff
@@ -118,21 +118,21 @@
...
     MagicFolder,
     load_magic_folders,
     save_magic_folders,
     maybe_upgrade_magic_folders,
 )
-from ..web.magic_folder import (
+from web.magic_folder import (
```
exarkun: I think this definitely should be either `from .web...` or `from magic_folder.web...`.  *Implicit* relative imports are awful.

src/magic_folder/cli.py
```diff
@@ -118,21 +118,21 @@
...
     magic_folder_web_service,
 )
 
-from ..status import (
+from status import (
```
exarkun: Same comment here regarding implicit relative imports.

src/magic_folder/cli.py
```diff
@@ -118,21 +118,21 @@
...
     status as _status,
 )
 
-from .._coverage import (
+from _coverage import (
```
exarkun: And here

src/magic_folder/test/cli/test_magic_folder.py
```diff
@@ -50,9 +50,8 @@
 from ...frontends.magic_folder import (
     MagicFolder,
 )
-from ...scripts import (
-    magic_folder_cli,
-)
+from magic_folder import cli as magic_folder_cli
```
exarkun: explicit relative import here would be more consistent with surrounding code, not that I'm a huge fan of the dot counting that goes into understanding something like `from ...frontends.magic_folder import ...`

meejah:
okay, i changed the imports so they're all relative. (And merged master)

PR #53, opened 2020-03-27
refactor the join command

vu3rdd:
`join` is split into an outer function that interacts with the command line options and another which does the core functionality. This would eventually move the command implementations closer to the main magic-folder module.

### DIFF
--- a/src/magic_folder/scripts/magic_folder_cli.py
+++ b/src/magic_folder/scripts/magic_folder_cli.py
@@ -371,39 +371,51 @@ class JoinOptions(BasedirOptions):
         self.local_dir = None if local_dir is None else argv_to_abspath(local_dir, long_path=False)
         self.invite_code = to_str(argv_to_unicode(invite_code))
 
-def join(options):
-    fields = options.invite_code.split(INVITE_SEPARATOR)
+def _join(invite_code, node_directory, local_dir, name, poll_interval):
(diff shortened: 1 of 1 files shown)


vu3rdd:
Addresses issue #52.

src/magic_folder/scripts/magic_folder_cli.py
```diff
@@ -371,39 +371,57 @@ def parseArgs(self, invite_code, local_dir):
...
+    try:
+        invite_code = options.invite_code
+        node_directory = options["node-directory"]
+        local_directory = options.local_dir
+        name = options['name']
+        poll_interval = options["poll-interval"]
+
+        rc = _join(invite_code, node_directory, local_directory, name, poll_interval)
+    except Exception as e:
+        print("%s" % str(e), file=options.stderr)
```
exarkun: `%s` implies `str(...)` so you can just do `"%s" % e`.

The types accepted by `%` are kind of a trap.  The right-hand argument does not necessarily have to be a tuple but you should ignore that and always pass a tuple.  So `"%s" % (e,)`.

Probably, though, new code should prefer `format`.  So `"{}".format(e)`.

However, `print` will take essentially anything and apply `str(...)` if necessary.  So `print(e, ...)`.

vu3rdd:
@exarkun Yes, definitely plan to rewrite the tests to use these new API.

PR #61, opened 2020-04-01
Refactor create command

vu3rdd:
This also moved `join` into its own module.

### DIFF
--- a/src/magic_folder/create.py
+++ b/src/magic_folder/create.py
@@ -108,6 +108,29 @@ def tahoe_create_alias(node_directory, alias, treq):
 
 @inlineCallbacks
 def magic_folder_create(alias, nickname, name, node_directory, local_dir, poll_interval, treq):
+    """
+    Create a magic-folder with the specified ``name`` (or ``default``
+    if not specified) and optionally invite the client and join with
+    the specified ``nickname`` and the specified ``local_dir``.
+
+    :param unicode alias: The alias of the folder to which the invitation is
+        being generated.
+
+    :param unicode nickname: The nickname of the invitee.
+
+    :param unicode name: The name of the magic-folder.
+
+    :param unicode node_directory: The root of the Tahoe-LAFS node.
+
+    :param unicode local_dir: The directory on the filesystem that the user wants
+        to sync between different computers.
+
+    :param integer poll_interval: Periodic time interval after which the
+        client polls for updates.
+
+    :param HTTPClient treq: An ``HTTPClient`` or similar object to use to make
+        the queries.
+    """
 
     # make sure we don't already have a magic-folder with this name before we create the alias
     maybe_upgrade_magic_folders(node_directory)
--- a/src/magic_folder/create.py
+++ b/src/magic_folder/create.py
@@ -97,8 +97,6 @@ def tahoe_create_alias(node_directory, alias, treq):
     # mkdir+add_alias
 
     nodeurl = get_node_url(node_directory)
-    if not nodeurl.endswith("/"):
-        nodeurl += "/"
 
     try:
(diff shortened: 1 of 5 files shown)


src/magic_folder/create.py
```diff
@@ -0,0 +1,139 @@
...
+
+@inlineCallbacks
+def tahoe_create_alias(node_directory, alias, treq):
+    # mkdir+add_alias
+
+    node_url = get_node_url(node_directory)
+    if not node_url.endswith("/"):
+        node_url += "/"
+    url = node_url + "uri?t=mkdir"
+
```
meejah, line 98: Other places use `hyperlink` (and we depend on it already). Not sure if that ends up with *shorter* code, but might be more robust :)

src/magic_folder/create.py
```diff
@@ -0,0 +1,139 @@
...
+    if response.code != OK:
+        raise Exception(format_http_error("Error", response))
+
+    new_uri = yield readBody(response)
+
+    _add_alias(node_directory, alias, new_uri)
+    # probably check for others..
+
+    # add_line_to_aliasfile(aliasfile, alias, new_uri)
+    returnValue(0)
```
meejah, line 109: can probably remove this line

src/magic_folder/create.py
```diff
@@ -0,0 +1,139 @@
...
+
+@inlineCallbacks
+def magic_folder_create(alias, nickname, name, node_directory, local_dir, poll_interval, treq):
+
```
meejah, line 113: This could benefit from a docstring (probably goes for the other new functions in here too)

src/magic_folder/create.py
```diff
@@ -0,0 +1,139 @@
...
+    if nickname is not None:
+        # inviting itself as a client
+        from twisted.internet import reactor
+        treq = HTTPClient(Agent(reactor))
+
+        try:
+            invite_code = yield _invite(node_directory, alias, nickname, treq)
+        except Exception:
+            raise Exception("Failed to invite after create")
```
meejah, line 133: Probably not actionable for this PR, but would be nice to print out what the user should type to re-try (they're in the situation here where they've created a magic-folder and they wanted to invite themselves, but that failed for some reason).

vu3rdd:
@meejah Thanks for the review. I have tried to take care of most of the suggestions in the commits pushed a while ago. Please have a look when you get time.

src/magic_folder/create.py
```diff
@@ -0,0 +1,156 @@
...
+# Until there is a web API, we replicate add_alias.
+# https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3286
+def add_line_to_aliasfile(aliasfile, alias, cap):
+    # we use os.path.exists, rather than catching EnvironmentError, to avoid
+    # clobbering the valuable alias file in case of spurious or transient
+    # filesystem errors.
+    if os.path.exists(aliasfile):
+        with codecs.open(aliasfile, "r", "utf-8") as f:
+            aliases = f.read()
+            f.close()
```
meejah, line 58: Don't need to explicitly `close` here; the context-manager will close when the "`with`" block ends (if you exit it "normally" or via exception etc).

src/magic_folder/create.py
```diff
@@ -0,0 +1,156 @@
...
+        if not aliases.endswith("\n"):
+            aliases += "\n"
+    else:
+        aliases = ""
+    aliases += "%s: %s\n" % (alias, cap)
+    with codecs.open(aliasfile+".tmp", "w", "utf-8") as f:
+        f.write(aliases)
+        f.close()
```
meejah, line 66: `close` not needed

Issue #67, opened 2020-04-04
`magic-folder run` doesn't exit on listen error


meejah:
If the magic-folder daemon fails to listen on its web-port, an exception is printed but the program keeps running (and logs "successful startup"). Magic-folder should exit if this fails.

(Also, I'm not sure that the `--web-port` should be optional? It currently defaults to `tcp:9889`. Maybe a better default would be "whatever is in `magic-folder.url` if that file exists in the node-dir" or just not having a default at all (i.e. it's a required option).

exarkun:
> (Also, I'm not sure that the --web-port should be optional? It currently defaults to tcp:9889. Maybe a better default would be "whatever is in magic-folder.url if that file exists in the node-dir" or just not having a default at all (i.e. it's a required option).

I wouldn't object to removing the default value.  Perhaps file a separate ticket for that part though.

meejah:
Okay, filed a ticket for the second part.

meejah:
Implemented

PR #68, opened 2020-04-04
add docstring, exit if we can't listen

meejah:
Don't see any other way to get the Service to stop eating the exception .. even if the API starts with `_` ..
See https://github.com/LeastAuthority/magic-folder/issues/67

### DIFF
--- a/src/magic_folder/web/magic_folder.py
+++ b/src/magic_folder/web/magic_folder.py
@@ -18,13 +18,26 @@ from allmydata.util.hashutil import (
     timing_safe_compare,
 )
 
+
 def magic_folder_web_service(web_endpoint, get_magic_folder, get_auth_token):
+    """
+    :param get_magic_folder: a callable taking a `nickname` and
+        returning a MagicFolder
+
+    :param get_auth_token: a callable returning the current auth-token
+
+    :returns: a new IService implementing the token-based Web API
+    """
     root = Resource()
     root.putChild(b"api", MagicFolderWebApi(get_magic_folder, get_auth_token))
-    return StreamServerEndpointService(
+    srv = StreamServerEndpointService(
         web_endpoint,
         Site(root),
     )
+    srv._raiseSynchronously = True
+    return srv
+
+
 
 def error(request, code, message):
     request.setResponseCode(code, message)
--- a/newsfragments/67.bugfix
+++ b/newsfragments/67.bugfix
@@ -1,0 +1,1 @@
+Exit magic-folder entirely if we can't listen on startup


src/magic_folder/web/magic_folder.py
```diff
@@ -18,13 +18,26 @@
...
+    :returns: a new IService implementing the token-based Web API
+    """
     root = Resource()
     root.putChild(b"api", MagicFolderWebApi(get_magic_folder, get_auth_token))
-    return StreamServerEndpointService(
+    srv = StreamServerEndpointService(
         web_endpoint,
         Site(root),
     )
+    srv._raiseSynchronously = True
```
exarkun, line 37: Ow ow ow.  Oog.

If we do this then at minimum we need a ticket filed against Twisted asking for the functionality we need and a link to that ticket here.

However, some grepping in Twisted shows me that you can have this behavior if you use `twisted.application.strports.service` instead of constructing the `StreamServerEndpointService` directly... However that API takes a string and guess that `web_endpoint` is an endpoint not a string.  Maybe the caller can be changed, though ...

However, this whole feature looks dodgy to me.  It relies on endpoint.listen returning an already-fired Deferred?  I imagine that's often the case for a TCP address collision but what about on Windows?  What about with asyncioreactor?  What about with a Tor endpoint (would you ever actually want that?  I'm not sure, maybe not).  Certainly the interface makes no such promises.

Yet you're right, I see no other facilities in `StreamServerEndpointService` for accomplishing this.  So a Twisted ticket may indeed be in order... not that we can expect anyone else to resolve it for us, I expect, nor will we get any benefit from it until magic-folder and tahoe-lafs are ported to Python 3 ... but maybe that delay is not the end of the world.

As for what we might actually want here, I suppose some public API to subscribe to success/failure notification of startup?  `svc.notifyOnListening().addErrback(shut_down_magic_folder)`?

meejah: > However, this whole feature looks dodgy to me. It relies on endpoint.listen returning an already-fired Deferred?

I don't understand what you're referring to here? `_raiseSynchronously` seems to handle deferreds properly...

> Yet you're right, I see no other facilities in StreamServerEndpointService for accomplishing this.

So I guess we have the following options:

 - don't use `StreamServerEndpointService`
 - use `StreamServerEndpointService` and rely on the (yes, private) `_raiseSynchronously` API anyway
 - augment Twisted's `IService` API, port everything to Python3, then fix this issue

exarkun:
```
    def privilegedStartService(self):
        """
        Start listening on the endpoint.
        """
        service.Service.privilegedStartService(self)
        self._waitingForPort = self.endpoint.listen(self.factory)
        raisedNow = []
        def handleIt(err):
            if self._raiseSynchronously:
                raisedNow.append(err)
            elif not err.check(CancelledError):
                log.err(err)
        self._waitingForPort.addErrback(handleIt)
        if raisedNow:
            raisedNow[0].raiseException()
        self._raiseSynchronously = False
```

The effect of setting `_raiseSynchronously` to `True` is that `raisedNow` might  have something appended to it sometimes.  That value will only be appended when the Deferred returned by `listen` fails.

`raisedNow` is checked synchronously after calling `listen` and before `privilegedStartService` is called and it is never checked after that.  Also the only way this feature can have any effect is by causing `privilegedStartService` to raise an exception (which it can only do synchronously).

So if you have an endpoint that returns a Deferred that hasn't already fired, `_raiseSynchronously` will cause no behavior to change.

exarkun: > So I guess we have the following options:

Two more options:

* Use `twisted.application.strports.service` which leverages the same feature but without using a private interface
* Wrap the underlying endpoint with a type that proxies the endpoint interface but also tracks failure and exposes it to us (probably in a Deferred-friendly way) and then use that new feature (add a callback that prints the error and shuts down the process) instead of anything having to do with `_raiseSynchronously`

meejah: To use `strports.service` we'd I guess have to register a string-parser for `RecordLocation` .. but not clear how we'd give it a callback-function.

exarkun: I meant those two options to be alternatives.  I forgot about `RecordLocation`.  That's probably the *reason* the API takes an endpoint instead of an endpoint string description.... So let's not go that route. :/

Here's a sketch of what I meant by the other option

```
@attr.s
@impolementer(IStreamServerEndpoint)
class ListenObserver(object):
    _endpoint = attr.ib()
    _observers = attr.ib(default=attr.Factory(list))

    def observe(self):
        self._observers.append(Deferred())
        return self._observers[-1]

    def listen(self, protocolFactory):
        d = self._endpoint.listen(protocolFactory)
        d.addBoth(self._deliver_result)
        return d

    def _deliver_result(self, result):
        observers = self._observers
        self._observers = []
        for o in observers:
            o.callback(result)
```

meejah: Additional wrinkle: `MagicFolderService` does implicit async work in `__init__`, kind of: it leverages `__attrs_post_init__` to do some extra setup which includes calling `magic_folder_web_service` and starting the services. As a consequence of starting the service, we try to listen.

(I have a branch proposing a thing incoming shortly .. the above arguably is a shortcoming of `Service` / `IService` "in general" but we can sidestep that by using an async-friend `create()` pattern).

meejah: okay, here's .. an approach (this one takes out the one integration test, that I had based on some integration-refactoring branch :/ so there no tests at all here):
https://github.com/LeastAuthority/magic-folder/pull/85/files

meejah:
This was accomplished with an alternative implementation not relying on private Twisted APIs.

PR #69, opened 2020-04-07
move web code to web.py

meejah:
see https://github.com/LeastAuthority/magic-folder/issues/6


exarkun:
This looks good to me.  If you get CI to run and it's green, merge.

PR #70, opened 2020-04-07
move frontends/ to top level

meejah:
See also #6


exarkun:
This looks good to me.  If you get CI to run and it's green, merge.

PR #71, opened 2020-04-07
refactor conftest/fixtures, and one more integration test

meejah:
This moves a bunch of `conftest.py` code into `util.py` (so it can be used in tests). Then, that code is used in a new `leave` test. This pairs two new participants to test `magic-folder leave` -- new participants and a new magic-folder are used instead of the `magic-folder` fixture in case "pair, leave, re-join" causes weird things to happen in "the" magic-folder fixture...


meejah:
re-based against current master

PR #78, opened 2020-04-15
Use the new pypy buster image for pypy CI

vu3rdd:


.circleci/config.yml
```diff
@@ -168,10 +168,10 @@ jobs:
         user: "nobody"
 
 
-  pypy2.7-7.1:
+  pypy2.7-7.3:
     <<: *DEBIAN
     docker:
-      - image: "magicfolderci/pypy:2.7-7.1.1-jessie"
+      - image: "pypy:2.7-7.3.0-buster"
         user: "nobody"
```
meejah, line 175: So before we were using a custom image, and now it's the straight pypy image? (It looks like this is where the `/nonexistsant` error probably comes from: `fix-permissions.sh` is put into our custom images, if I understand the `.circleci` folder correctly..)

vu3rdd: ah, okay. Thanks. I have just started a CI job to build our custom images.

PR #84, opened 2020-04-16
Ticket67 approach 2

meejah:
Here's at least an integration test for #67 .. it's based off a integration-test refactoring branch because I needed some of the fixtures as helpers.


meejah:
(oh, it seems the branch i based this off has more than I thought .. the only relevant bits to this PR are the last two commits)

meejah:
The guts of this change (without the integration test or the other changes that brought in) are here: https://github.com/LeastAuthority/magic-folder/pull/85

PR #85, opened 2020-04-16
ticket 67: exit if we can't listen

meejah:
See #67


src/magic_folder/web.py
```diff
@@ -18,14 +18,25 @@
...
 )
 
+
 def magic_folder_web_service(web_endpoint, get_magic_folder, get_auth_token):
+    """
+    :param web_endpoint: a IStreamServerEndpoint where we should listen
+
+    :param get_magic_folder: a callable that returns a MagicFolder given a name
+
+    :param get_auth_token: a callabl that returns the current authentication token
```
vu3rdd, line 28: typo => "callable" a missing e?

meejah: :+1:

PR #87, opened 2020-04-17
our own eliotutil 

meejah:
...with former fields from tahoe


vu3rdd:
This goes further. However, it fails because tests invokes `tahoe magic-folder`.

vu3rdd:
It works fine with tahoe-lafs 1.14.0.

p
```diff
@@ -0,0 +1,87 @@
+diff --git a/integration/test_general_cli.py b/integration/test_general_cli.py
```
exarkun, line 1: What is this whole file?

src/magic_folder/magic_folder.py
```diff
@@ -38,6 +38,11 @@
     yamlutil,
     eliotutil,
 )
+from util.eliotutil import (
```
exarkun, line 41: Let's go with explicit relative imports ... `from .util.eliotutil import ...`

p
vu3rdd: hmm.. I have no idea what that is and how that sneaked in. @meejah do you know?

meejah: oh, weird that's from another PR of mine I think

PR #90, opened 2020-04-17
remove _get_json_for_fragment() from cli

vu3rdd:
Integration test use `_get_json_for_fragment()`, so this has been modified to use the new status API.

### DIFF
--- a/integration/test_magic_folder.py
+++ b/integration/test_magic_folder.py
@@ -19,9 +19,18 @@ from eliot import (
 from twisted.internet.defer import (
     returnValue,
 )
+
 from twisted.web.client import (
     Agent,
     readBody,
+)
+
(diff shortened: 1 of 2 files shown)


meejah:
Hmm, does this mean the new status API doesn't need a token??

vu3rdd:
> Hmm, does this mean the new status API doesn't need a token??

It is being accessed inside the `status()` function, it looks like: https://github.com/LeastAuthority/magic-folder/blob/master/src/magic_folder/status.py#L71

meejah:
Ahhh I see. Sorry, read it too fast :)

vu3rdd:
Thanks. I will keep it open for now until #87 is in and after that will do a CI run and then merge once it is all green.

PR #115, opened 2020-04-30
Use the released version of tahoe-lafs instead of the git master HEAD

vu3rdd:


.circleci/populate-wheelhouse.sh
```diff
@@ -36,15 +36,14 @@ PIP="${BOOTSTRAP_VENV}/bin/pip"
 # Tell pip where it can find any existing wheels.
 export PIP_FIND_LINKS="file://${WHEELHOUSE_PATH}"
 
-# We don't yet have a compatible release.
-# https://pip.pypa.io/en/stable/reference/pip_install/#git
-export TAHOE_LAFS="git+https://github.com/tahoe-lafs/tahoe-lafs@master#egg=tahoe-lafs"
+# tahoe-lafs >= 1.14.0 is a compatible release
+export RUNTIME_DEPS="tahoe-lafs"
```
exarkun, line 40: I might suggest `tahoe-lafs==1.14.0` here.  I don't think we want the tahoe dependency to float now since Tahoe-LAFS officially has no public Python API and we use a  bunch of Tahoe-LAFS Python APIs.

We might get lucky and 1.15.0 doesn't break anything in magic-folder but we might not and without a pinned version, anyone installing *this* version of magic-folder will get whatever the latest Tahoe-LAFS at installation time happens to be.

Pinning only half works in Python (during upgrades and for dependencies shared by multiple projects, constraints aren't always respected) but it's probably still better than nothing.

PR #122, opened 2020-05-06
Significantly rework the Nix packaging

### LINKED ISSUE #121: Nix packaging has unused and misleading nevow and eliot packages
Nevow is a transitive dependency from Tahoe-LAFS.  We don't need to include it explicitly ourselves.  We get it when we say we depend on Tahoe-LAFS.

Eliot is a transitive and direct dependency.  Tahoe-LAFS supplies its own Nix packaging for Eliot which we have to play nicely with.  It's hard to do this directly because Tahoe-LAFS puts its Eliot package in with an overlay that it extends nixpkgs with when we call it.  This is fine if we want to use it unchanged but we want to turn off its test suite because it's flaky.

exarkun:
Fixes #121


exarkun:
Thanks.  Looks like the coverage failure is probably spurious and due to "normal" changes in coverage from run to run.  So, ignoring it and merging.

Issue #109, opened 2020-04-28
Create and upload a snapshot when a file is added to a magic-folder


exarkun:
The watchdog system should trigger a file modified/created event and that would put an item into the queue that eventually create an in-memory representation of the snapshot and an on-disk representation so that even if the computer crashes or we lose network connectivity, we maintain the ability to recreate history of the changes. These local snapshot representation is later "committed" to the grid. 

So, this task comprise of these separate tickets:

- [ ] [Observe File Addition](https://github.com/LeastAuthority/magic-folder/issues/138).
- [x] [Python API to manually add files to upload queue](https://github.com/LeastAuthority/magic-folder/issues/161).
- [x] [Create in-memory representation of the Local Snapshot](https://github.com/LeastAuthority/magic-folder/issues/139).
- [x] [Create on-disk representation of the Local Snapshot](https://github.com/LeastAuthority/magic-folder/issues/140).
- [x] [Commit the Local Snapshot to the grid](https://github.com/LeastAuthority/magic-folder/issues/141).
- [x] [Add an HTTP API to create a local snapshot of a file](https://github.com/LeastAuthority/magic-folder/issues/266).
- [x] [Process local snapshots into remote snapshots](https://github.com/LeastAuthority/magic-folder/issues/265)
- [x] [Add verified-fakes hot fix to magic-folder repo](https://github.com/LeastAuthority/magic-folder/issues/142).

meejah:
I wonder if it would be "easier" (less work) to make the "Observe Filesystem Changes" into the "API" version? (That is, "provide an API that tells us to queue a particular file for creation of a new snapshot").

From recent discussion it sounds like users might want more control of this *anyway*. Another "pro" is that I think it'll be easier to test an API-based Snapshot creator rather than faking out what watchdog (or whatever) does. Then the "Observe Filesystem Changes" part can be a followup (and use the API, even).

vu3rdd:
@meejah That sounds like a good idea. 

Also, just for the purpose of writing this down more explicitly, the flow would be:

1. Observe file addition (either via API or via watchdog)
2. If watchdog triggered them, call the API to queue these additions.
3. Create local snapshots.
4. Persist these snapshots to the disk.
5. Simultaneously push these snapshots into a queue.
6. The queue is emptied to "commit" the Local Snapshots into the grid and these become "Remote Snapshots".
7. Update the sqlite to add/update the local_files entry for the relative file path with the latest snapshot URI.

PR #125, opened 2020-05-15
Create snapshot as Tahoe immutable directories

vu3rdd:
This PR is towards #109. It adds a new module called `snapshot.py` that adds a `snapshot_create` function, which takes a file path and a list of parent snapshots. The returned URI represents the Tahoe-LAFS immutable directory that corresponds to the snapshot.

There are many pieces missing from this implementation. 
- We haven't figured out about signatures, authorship etc. 
- We haven't yet hooked up the Uploading parts into the snapshots and so the `latest_snapshot` column of the db entry is fake at the moment.
- We currently use a new schema that just adds an extra column to the old schema. This was to make sure that we don't fail all the existing tests with the change of schema. 

Once the snapshots are fully implemented, we wouldn't need certain entries from the db and at that point, we would change the db schema accordingly.

### DIFF
--- a/src/magic_folder/magic_folder.py
+++ b/src/magic_folder/magic_folder.py
@@ -1483,6 +1483,7 @@ class Uploader(QueueMixin):
                         filecap,
                         last_downloaded_uri,
                         last_downloaded_timestamp,
+                        u'URI:DIR2-RO:0',
                         pathinfo,
                     )
                     self._count('files_uploaded')
@@ -1519,6 +1520,7 @@ class Uploader(QueueMixin):
                         last_uploaded_uri=dirnode.get_uri(),
                         last_downloaded_uri=None,
                         last_downloaded_timestamp=now,
+                        latest_snapshot=u'URI:DIR2-RO:0',
                         pathinfo=pathinfo,
                     )
                 upload_d.addCallback(_dir_succeeded)
@@ -1568,6 +1570,7 @@ class Uploader(QueueMixin):
                         filecap,
                         last_downloaded_uri,
                         last_downloaded_timestamp,
+                        u'URI:DIR2-RO:0',
                         pathinfo
                     )
                     self._count('files_uploaded')
@@ -2008,6 +2011,7 @@ class Downloader(QueueMixin, WriteFileMixin):
                 last_uploaded_uri,
                 last_downloaded_uri,
                 last_downloaded_timestamp,
+                u'URI:DIR2-RO:0',
                 written_pathinfo,
             )
             self._count('objects_downloaded')
--- a/src/magic_folder/magic_folder.py
+++ b/src/magic_folder/magic_folder.py
@@ -383,7 +383,7 @@ class MagicFolder(service.MultiService):
         service.MultiService.__init__(self)
 
         clock = clock or reactor
-        db = magicfolderdb.get_magicfolderdb(dbfile, create_version=(magicfolderdb.SCHEMA_v1, 1))
+        db = magicfolderdb.get_magicfolderdb(dbfile, create_version=(magicfolderdb.SCHEMA_v2, 1))
         if db is None:
             raise Exception('ERROR: Unable to load magic folder db.')
 
--- a/src/magic_folder/magic_folder.py
+++ b/src/magic_folder/magic_folder.py
@@ -383,7 +383,7 @@ class MagicFolder(service.MultiService):
         service.MultiService.__init__(self)
 
         clock = clock or reactor
-        db = magicfolderdb.get_magicfolderdb(dbfile, create_version=(magicfolderdb.SCHEMA_v2, 1))
+        db = magicfolderdb.get_magicfolderdb(dbfile, create_version=(magicfolderdb.SCHEMA_v1, 1))
         if db is None:
             raise Exception('ERROR: Unable to load magic folder db.')
 
@@ -1483,7 +1483,6 @@ class Uploader(QueueMixin):
                         filecap,
                         last_downloaded_uri,
                         last_downloaded_timestamp,
-                        u'URI:DIR2-RO:0',
                         pathinfo,
                     )
                     self._count('files_uploaded')
@@ -1520,7 +1519,6 @@ class Uploader(QueueMixin):
                         last_uploaded_uri=dirnode.get_uri(),
                         last_downloaded_uri=None,
                         last_downloaded_timestamp=now,
-                        latest_snapshot=u'URI:DIR2-RO:0',
                         pathinfo=pathinfo,
                     )
                 upload_d.addCallback(_dir_succeeded)
@@ -1570,7 +1568,6 @@ class Uploader(QueueMixin):
                         filecap,
                         last_downloaded_uri,
                         last_downloaded_timestamp,
-                        u'URI:DIR2-RO:0',
                         pathinfo
                     )
                     self._count('files_uploaded')
@@ -2011,7 +2008,6 @@ class Downloader(QueueMixin, WriteFileMixin):
                 last_uploaded_uri,
                 last_downloaded_uri,
                 last_downloaded_timestamp,
-                u'URI:DIR2-RO:0',
                 written_pathinfo,
             )
             self._count('objects_downloaded')
--- a/src/magic_folder/magicfolderdb.py
+++ b/src/magic_folder/magicfolderdb.py
@@ -10,7 +10,6 @@ from .util.eliotutil import (
     LAST_UPLOADED_URI,
     LAST_DOWNLOADED_URI,
     LAST_DOWNLOADED_TIMESTAMP,
-    LATEST_SNAPSHOT,
     PATHINFO,
     validateSetMembership,
     validateInstanceOf,
@@ -21,8 +20,7 @@ from eliot import (
 )
 
 PathEntry = namedtuple('PathEntry', 'size mtime_ns ctime_ns version last_uploaded_uri '
-                                    'last_downloaded_uri last_downloaded_timestamp '
-                                    'latest_snapshot')
+                                    'last_downloaded_uri last_downloaded_timestamp')
 
 PATHENTRY = Field(
     u"pathentry",
@@ -34,7 +32,6 @@ PATHENTRY = Field(
         "last_uploaded_uri": v.last_uploaded_uri,
         "last_downloaded_uri": v.last_downloaded_uri,
         "last_downloaded_timestamp": v.last_downloaded_timestamp,
-        "latest_snapshot": v.latest_snapshot,
     },
     u"The local database state of a file.",
     validateInstanceOf((type(None), PathEntry)),
@@ -49,7 +46,7 @@ _INSERT_OR_UPDATE = Field.for_types(
 
 UPDATE_ENTRY = ActionType(
     u"magic-folder-db:update-entry",
-    [RELPATH, VERSION, LAST_UPLOADED_URI, LAST_DOWNLOADED_URI, LAST_DOWNLOADED_TIMESTAMP, LATEST_SNAPSHOT, PATHINFO],
+    [RELPATH, VERSION, LAST_UPLOADED_URI, LAST_DOWNLOADED_URI, LAST_DOWNLOADED_TIMESTAMP, PATHINFO],
     [_INSERT_OR_UPDATE],
     u"Record some metadata about a relative path in the magic-folder.",
 )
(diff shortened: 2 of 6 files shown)


exarkun:
> This PR is towards #109 

Logistically the PR needs to actually close a ticket.  The news builder won't let you have two `109.feature` fragments.  This isn't just a tool limitation, it's part of the design of the system to have a workflow where every change to master has some self-contained description.

So, figure out which part of #109 this is, file a new ticket, and target this PR at that ticket.

vu3rdd:
> 
> So, figure out which part of #109 this is, file a new ticket, and target this PR at that ticket.

Thanks. Will file some additional tickets and retarget this PR.

src/magic_folder/magicfolderdb.py
```diff
@@ -72,9 +75,41 @@
...
+ADD_COLUMN = \
+    """
+    ALTER TABLE local_files ADD COLUMN latest_snapshot VARCHAR(256);
+    """
+
+UPDATE_v1_to_v2 = ADD_COLUMN
+
+UPDATERS = {
+    2: UPDATE_v1_to_v2,
+}
```
exarkun, line 109: It looks like the upgrade logic is incomplete.  For what it's worth, I have a very rudimentary SQLite3 schema management code in ZKAPAuthorizer.  Of course, there are also many very featureful (and complex) SQL schema management libraries out there as well.

Taking a step back, I wonder if we want to make any schema changes at this point at all though.  I see that various SQL has been adjusted to load this field from and save this field to the database.  However nothing actually uses this value.  I see this as an attempt to break the large task into digestible smaller pieces and I think that's good.  However, it's not clear where this piece goes.  I suppose a future PR could add a new `Uploader` or `Downloader`-alike which makes use of this information.  Those classes won't need a lot of the existing database fields though, as the existing uploader/downloader don't need the new fields.  So sharing the database between two different consumers seems like it adds extra complexity to both old and new.

If you have a vision for where this goes, great!  Can you lay it out somewhere?  Or if you want to talk about constructing such a vision, let's do that.

src/magic_folder/snapshot.py
```diff
@@ -0,0 +1,172 @@
...
+@inlineCallbacks
+def tahoe_create_snapshot_dir(nodeurl, content, parents, timestamp, treq):
+    """
+    :param unicode content: readcap for the content.
+    :param [unicode] parents: List of parent snapshot caps
+    :param unicode author: readcap that represents the author pubkey
+    :param integer timestamp: POSIX timestamp that represents the creation time
+    :return Deferred[unicode]: The readcap associated with the newly created
+        snapshot.
+    """
```
exarkun, line 90: It would probably be useful to have a structured representation of most/all of this information as a Python object so that there is a thing that can be pointed at and called a "snapshot" rather than needing to manage all of a snapshots fields as separate parameters/variables.

src/magic_folder/snapshot.py
```diff
@@ -0,0 +1,172 @@
...
+    if response.code != OK:
+        returnValue((yield bad_response(url, response)))
+
+    result = yield readBody(response)
+    returnValue(result)
+
+@inlineCallbacks
+def _store_file_immutable(nodeurl, filepath):
+
+    from twisted.internet import reactor
```
exarkun, line 134: treq should probably be a parameter here.  In which case `_store_file_immutable` doesn't really do anything useful above and beyond `tahoe_put_immutable`, does it?

src/magic_folder/snapshot.py
```diff
@@ -0,0 +1,172 @@
...
+
+    # - store the file content and get the immutable cap (content)
+    #     PUT /uri?format=CHK
+    content_cap = yield _store_file_immutable(nodeurl, filepath)
+
+    now = time.time()
+
+    # - HTTP POST mkdir-immutable
+    from twisted.internet import reactor
+    treq = HTTPClient(Agent(reactor))
```
exarkun, line 165: treq should be a parameter

src/magic_folder/test/test_magic_folder.py
```diff
@@ -2452,3 +2471,108 @@ def setUp(self):
         d = super(RealTestAliceBob, self).setUp()
         self.inotify = get_inotify_module()
         return d
+
+class SnapshotTest(AsyncTestCase):
```
exarkun, line 2475: I'd like to try to find a better testing strategy than "launch Tahoe-LAFS, throw some HTTP at it".  These tests are going to be slow, failure prone, and a pain in the ass to debug when they fail.  All of these things will also deny us the ability to have really in-depth testing of all of the edge-cases for the snapshot-handling tests.

meejah:
I haven't looked at the code yet (just read the description) -- I think it would be more flexible and future-proof if it took a "file-like object" instead of a file-path. Even if that's only used for testing immediately :)

I'll look at the code itself (and exarkun's review) in a couple hours.

src/magic_folder/snapshot.py
meejah: As a general point, I've usually tried to have `reactor` be an argument to functions that need it .. this at bare minimum lets you pass a `Clock()` (or whatever) for testing and avoids the imports inside a method. (Obviously here that is captured by "pass the http client itself" as @exarkun suggests). 

I guess I'm partly mentioning it here as a question: *should* this code-base prefer to have the reactor (or reactor-using thing) passed in? I think "yes".

src/magic_folder/magicfolderdb.py
```diff
@@ -20,7 +21,8 @@
 )
 
 PathEntry = namedtuple('PathEntry', 'size mtime_ns ctime_ns version last_uploaded_uri '
-                                    'last_downloaded_uri last_downloaded_timestamp')
+                                    'last_downloaded_uri last_downloaded_timestamp '
+                                    'latest_snapshot')
```
meejah, line 26: I think this echos another comment somewhat, but if we bundled "the database changes" separately I think using something besides `namedtuple` here would be good? (e.g. at `@attrs` class?). I would hope we don't need to keep many of these fields too, as they're pretty specific to the current way the up/downloader and conflict-detection works (`version`, anything with `_uri_` in it ...).

Also, some of these will be mirrored by the snapshot metadata, right? Like the timestamps?

src/magic_folder/snapshot.py
```diff
@@ -0,0 +1,172 @@
...
+from .common import (
+    bad_response,
+)
+
+
+# version of the snapshot scheme
+SNAPSHOT_VERSION = 1
+
+@inlineCallbacks
+def tahoe_put_immutable(nodeurl, filepath, treq):
```
meejah, line 50: For a public API I think it would be better to pass a file-like object (or maybe our own interface?) instead of a file-path. This falls into the general idea of "pass the thing you actually need, not information to get the thing".

Passing a file-like would make testing easier and also not tie the API to "a filesystem" (e.g. what if I want to upload a JSON document that I just serialized from a `dict` in-memory).

src/magic_folder/snapshot.py
```diff
@@ -0,0 +1,172 @@
...
+def _store_file_immutable(nodeurl, filepath):
+
+    from twisted.internet import reactor
+    treq = HTTPClient(Agent(reactor))
+    rocap = yield tahoe_put_immutable(nodeurl, filepath, treq)
+
+    returnValue(rocap)
+
+@inlineCallbacks
+def snapshot_create(node_directory, filepath, parents):
```
meejah, line 141: At risk of this just being "bikeshed" territory ;) I think it would be more-consistent to be named `create_snapshot` to mirror other tahoe and magic-folder code. Based on a definitely-scientific `grep` survey, `def create_*` is more popular than `def *_create` in both codebases. See-also the comment earlier about "a file-like object" versus file-path.

src/magic_folder/test/test_magic_folder.py
meejah: treq has some built-in testing support, like `treq.testing.RequestTraversalAgent` that could probably be leveraged to provide a fake Tahoe API -- that itself is probably a whole PR, maybe, but could be a useful direction to make the tests more immediate and less directly tied to "a running tahoe"...

src/magic_folder/magic_folder.py
```diff
@@ -1568,6 +1570,7 @@ def _add_db_entry(filenode):
                         filecap,
                         last_downloaded_uri,
                         last_downloaded_timestamp,
+                        u'URI:DIR2-RO:0',
```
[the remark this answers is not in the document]
vu3rdd: This PR is terribly incomplete. It only implemented just the part that creates a tahoe-lafs snapshot representation, so I was faking all the db operations. Also, something I didn't make it very clear in the PR description is that since the db storage is not done yet, this was co-existing with the existing mechanisms of db operation. I thought addressing db operation changes would be a big change since all the tests would break, so I thought the upload/download of snapshots could be done as a separate PR.  So, these `URI:DIR2-RO` (I thought I changed it to `CHK` URLs as that's what they are) are fake things passed to the db to store against the file path. I was unclear on how to go about it when I started, but I have a slightly better picture now.. I think I will address some of these in this PR itself in subsequent commits.

src/magic_folder/snapshot.py
```diff
@@ -0,0 +1,172 @@
...
+    url = node_url.child(
+        u"uri",
+    ).add(
+        u"format",
+        u"CHK",
+    )
+
+    # XXX: check whether we need to set any headers.
+    put_uri = url.to_uri().to_text().encode("ascii")
+    with open(str(filepath), "rb") as file:
```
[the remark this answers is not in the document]
vu3rdd: I didn't know about any of these issues. Thanks.

src/magic_folder/magicfolderdb.py
vu3rdd: Yes, most of the existing fields in the database would go away as we discussed in the design document. I was trying to make database and uploader changes as well along with this PR and then realized it would get very big and unmanageable to review. 

I think it makes sense to address this in a later PR along with `Uploader` and `Downloader` changes. I will remove these changes for now.

src/magic_folder/snapshot.py
vu3rdd: > I guess I'm partly mentioning it here as a question: should this code-base prefer to have the reactor (or reactor-using thing) passed in? 
I do not have knowledge about twisted so deeply to answer that question.

src/magic_folder/test/test_magic_folder.py
vu3rdd: I will explore the `RequestTraversalAgent` option. Yes, launching Tahoe-LAFS node definitely looks ugly and it was very hard to debug even simple type errors that I was making while doing the development. An option is the "sans-io" type of approach where we generate Tahoe requests as messages?

src/magic_folder/magicfolderdb.py
vu3rdd: Thanks. I will read up on `@attrs` class.

src/magic_folder/snapshot.py
vu3rdd: Yes, passing "file like" things instead of file paths sounds very good. .

src/magic_folder/test/test_magic_folder.py
meejah: The "ideal" situation here would be if we had a "verified fake" Tahoe API endpoint .. that is, the "root" that we can give to `RequestTraversalAgent`. Then we'd be doing in-memory requests to a fake version of a Tahoe node (for any API requests).

The "verified" part I guess would come from integration-style tests (that actually start a Tahoe and confirm that our "fake" does the same thing as the real one). I am thinking of this article when talking about it: https://pythonspeed.com/articles/verified-fakes/

However, we do not have that. This would be a whole PR (or more!) by itself -- maybe this is something I could help with. At first, we could cover *just* enough to get some tests for snapshots. Meantime, I think the best approach for this PR is to get as much coverage from unit-style tests on the helpers and so on as we can.

...it is maybe even premature to write such a test? That is, if we can't write an integration test for snapshots yet (which I think is true?) then can we write this "verified fake" style of test yet?

exarkun: maybe moot though since there isn't such a python API ... yet?

meejah: Yes, I think such an API would be great to have. I'd previously proposed this as a way to "greenfield" a namespace without `allmydata` too (i.e. at first the only public stuff in such an API would be the minimum needed to do public Tahoe stuff) .. but, yeah, there isn't such an API.

On the flip side, since `tahoe` can only be interacted with via the command-line and WebUI there is no need for it to be a Python program at all ;)

meejah:
This PR is superseded by others (right? if not, please re-open)

PR #137, opened 2020-06-15
document how invites work

vu3rdd:


meejah:
Looks good .. it's probably worth mentioning that Alice needs to be the creator of the magic-folder to invite people. Since it's "up to the user" to co-ordinate writes in Tahoe, there should only be ONE person who has the write-cap for the collective.

docs/invites.rst
```diff
@@ -0,0 +1,47 @@
+.. -*- coding: utf-8 -*-
+
+.. _invites:
```
exarkun, line 3: Cool.  You probably want to link it into the index, too.

Issue #142, opened 2020-06-19
Add allmydata.testing.web "hot fix" to magic-folder


vu3rdd:
Since we would like to make magic-folder work with a released version of tahoe-lafs, the tahoe-lafs PR from @meejah that adds verified-fakes testing for the Web APIs may be added into magic-folders for now.

https://github.com/tahoe-lafs/tahoe-lafs/pull/728

PR #143, opened 2020-06-19
Fast and easy testing of programs that use Tahoe-LAFS web API

vu3rdd:
This is pulled from @meejah's PR and address #142:

  https://github.com/tahoe-lafs/tahoe-lafs/pull/728

which makes it easy and fast to do testing against Tahoe-LAFS web API.
Without this module, testing such programs would involve spinning
up a real Tahoe-LAFS node, which is slow and tedious.

Until there is an official Tahoe-LAFS release which includes the above code, we pull in this code as a hot-fix and should be removed and the dependent modules modified to use the equivalent modules from Tahoe-LAFS.

### DIFF
--- a/newsfragments/142.other
+++ b/newsfragments/142.other
@@ -1,0 +1,1 @@
+hot-fix from Tahoe-LAFS repo to do Tahoe-LAFS web api testing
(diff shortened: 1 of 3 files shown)


meejah:
Shouldn't we wait until we have the "testing/web" etc code in Tahoe and then copy that? (i.e. wait until https://github.com/tahoe-lafs/tahoe-lafs/pull/728 is merged)?

vu3rdd:
I thought you were already done with reviews and was going to merge it. Will wait until you merge it.

PR #144, opened 2020-06-20
139.create local snapshot

### LINKED ISSUE #139: Create in-memory representation of the local snapshot
The inotify/watchdog process pushes modified files into a work queue. These items needs to be emptied from the queue and an in-memory representation of the Local Snapshot is constructed.

meejah:
Fixes #139


meejah:
The NixOS thing looks like spurious failure? But @meejahcircleci account can't re-start it

src/magic_folder/snapshot.py
```diff
@@ -0,0 +1,247 @@
...
+    if not isinstance(author, LocalAuthor):
+        raise ValueError(
+            "create_snapshot 'author' must be a LocalAuthor instance"
+        )
+
+    # when we do uploads, we will distinguish between remote and local
+    # parents, so the "parents" list may contain either kind in the
+    # future.
+    parents_local = []
+    for idx, parent in enumerate(parents_local):
```
vu3rdd, line 200: `enumerate(parents)` ? Because `parents` is a list of only LocalSnapshots at the moment?

src/magic_folder/test/test_snapshot.py
```diff
@@ -0,0 +1,228 @@
...
+        )
+
+        # now modify the same file and create a new local snapshot
+        data2 = io.BytesIO(content2)
+        d = create_snapshot(
+            name=filename,
+            author=self.alice,
+            data_producer=data2,
+            snapshot_stash_dir=self.stash_dir,
+            parents=local_snapshots[0],
```
vu3rdd, line 221: `parents=local_snapshots` since `parents` is of type "List of LocalSnapshots"?

meejah:
Should be ready to go now; my tools show 98% patch-coverage let's see if codecov agrees :/

src/magic_folder/snapshot.py
```diff
@@ -0,0 +1,247 @@
...
+            return f.read()
+
+
+@inlineCallbacks
+def create_snapshot(name, author, data_producer, snapshot_stash_dir, parents=None):
+    """
+    Creates a new LocalSnapshot instance that is in-memory only. All
+    data is stashed in `snapshot_stash_dir` before this function
+    returns.
+
```
vu3rdd, line 174: Missing docstring for the `name` parameter.

meejah: fixed, thanks

PR #146, opened 2020-06-20
get rid of --basedir option

### LINKED ISSUE #145: Remove redundant --basedir
Many sub-commands take a `--basedir` / `-C` option. This behavior is inherited from tahoe, but overlaps with the global `--node-directory` option. To avoid confusion I think we should delete the option from all the sub-commands (also, I don't believe it's actually used anyway).

meejah:
fixes #145

### DIFF
--- a/newsfragments/145.bugfix
+++ b/newsfragments/145.bugfix
@@ -1,0 +1,1 @@
+Remove --basedir option from all subcommands (use --node-directory instead)
--- a/newsfragments/145.bugfix
+++ b/newsfragments/145.bugfix
@@ -1,1 +1,1 @@
-Remove --basedir option from all subcommands (use --node-directory instead)
+Sub-commands no longer accept the --basedir option; use --node-directory instead
--- a/src/magic_folder/cli.py
+++ b/src/magic_folder/cli.py
@@ -171,7 +171,7 @@ class CreateOptions(usage.Options):
 
         if self.nickname and not self.local_dir:
             raise usage.UsageError("If NICKNAME is specified then LOCAL_DIR must also be specified.")
-        node_url_file = os.path.join(self['node-directory'], u"node.url")
+        node_url_file = os.path.join(self.parent['node-directory'], u"node.url")
         self['node-url'] = fileutil.read(node_url_file).strip()
 
 
@@ -186,7 +186,7 @@ def create(options):
         treq = HTTPClient(Agent(reactor))
 
         name = options['name']
-        nodedir = options["node-directory"]
+        nodedir = options.parent["node-directory"]
         localdir = options.local_dir
         rc = yield _create(options.alias, options.nickname, name, nodedir, localdir, options["poll-interval"], treq)
         print("Alias %s created" % (quote_output(options.alias),), file=options.stdout)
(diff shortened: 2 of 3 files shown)


newsfragments/145.bugfix
```diff
@@ -0,0 +1 @@
+Remove --basedir option from all subcommands (use --node-directory instead)
```
exarkun, line 1: I like Twisted's newsfragment style guide: https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles

Though I notice now it doesn't cover things like CLI interfaces.  It seems like there is a logical way to extend it to this, though.  Can we adopt such a convention for magic-folder?

src/magic_folder/cli.py
```diff
@@ -1032,6 +1022,19 @@ class MagicFolderCommand(BaseOptions):
         "readcaps contained in the master magic-folder directory."
     )
 
+    def parseOptions(self, *args, **kw):
+        """
+        Override usage.Options
```
exarkun, line 1027: Maybe mention that we're trying to guarantee there is a `node-directory`.

Also, fwiw, maybe there's another way to do this.  We define the `node-directory` option and we define it like:

```
    ["node-directory", "d", None, NODEDIR_HELP]
```

If that were instead

```
    ["node-directory", "d", None, NODEDIR_HELP, FilePath]
```

then I think a value would be required (and then it would be stored as a `FilePath` instead of a `str` or `bytes` or whatever it is now).

Whether you want to make that change or not, the error case here probably needs a new test.

newsfragments/145.bugfix
meejah: Adopting some kind of guidelines would be good.

For this, I didn't really know whether to call it a bugfix or a feature (or even "incompat"?) and I guess from the guidelines it should be like a "now" sentence?

"--node-directory is now the only way to specify the Tahoe node"

or

"Sub-commands no longer accept the --basedir option; use --node-directory instead"

meejah: Maybe "The --node-directory option is now required and is the only way to specify the Tahoe directory"?

src/magic_folder/cli.py
meejah: Just adding `FilePath` as the factory isn't enough, because that just results in a full-screen exception (instead of a `UsageError`). That's probably a better way to store the path, but I'll leave that to a different PR (because it would be a bunch more changes to everything that uses `node-directory`). I filed https://github.com/LeastAuthority/magic-folder/issues/147

newsfragments/145.bugfix
exarkun: The middle option there looks best to me but the others look good as well.

src/magic_folder/cli.py
exarkun: Ah, I thought it might still produce a `UsageError` because https://github.com/twisted/twisted/blob/trunk/src/twisted/python/usage.py#L50-L63 but I guess I have misunderstood how that code works or gets used.

meejah: hmm, weird .. definitely looks like it SHOULD say it's required (at least)

meejah: I made an SSCCE that also fails to fail: http://paste.debian.net/plain/1153387

meejah:
While adding a few more tests, I re-worked the options-code as well (in case you wanted to look again @exarkun )

src/magic_folder/cli.py
exarkun: Hm okay looking more carefully I guess the coercer doesn't get called on the default value, it only gets called on values given as options.

PR #149, opened 2020-06-23
no blank piece in URI

### LINKED ISSUE #148: integration tests broken with tahoe master

meejah:
fixes #148 
Also a somewhat better error message upon failure.


meejah:
fix should be in Tahoe, not here (a trailing-slash is documented as acceptable in Tahoe).

PR #155, opened 2020-06-26
Documentation updates

sajith:
This unfortunately has to be a rather big set of changes.  Sorry about that!

Rendered HTML after this set of updates is [here](https://nonzen.in/tmp/tahoe/magic-folder/).  For comparison/reference, [this](https://magic-folder.readthedocs.io/en/latest/) is what we have currently in Read the Docs.


docs/usage.rst
```diff
@@ -0,0 +1,188 @@
...
+   $ magic-folder -n $ALICE_NODE invite magic: bob >invitecode
+   $ export INVITECODE=$(cat invitecode)
+
+   $ BOB_NODE=../grid/bob
+   $ BOB_FOLDER=../local/bob
+
+   $ magic-folder -n $BOB_NODE join "$INVITECODE" $BOB_FOLDER
+
+   $ deamonize magic-folder -n $ALICE_NODE run
+   $ daemonize magic-folder -n $BOB_NODE run
```
meejah, line 172: spelling (daemonize).

docs/usage.rst
```diff
@@ -0,0 +1,188 @@
...
+
+If you want to test that things work as expected using a single
+machine, you can create two separate Tahoe-LAFS nodes, and assign
+corresponding magic folders with them, like so:
+
+.. code-block:: console
+
+   $ ALICE_NODE=../grid/alice
+   $ ALICE_FOLDER=../local/alice
+
```
meejah, line 160: I think these will need `export ALICE_NODE=...` if people are copy-pasting into a shell (instead of putting this all in one shell-script) right? At least, for bash.

meejah:
Oh, and I think we decided to just push branches to the main repo instead of via forks (because the CI only appears to run if you push to the main reop anyway). Please remember to delete these branches though :)

docs/usage.rst
```diff
@@ -0,0 +1,188 @@
...
+Inviting Participant Devices
+----------------------------
+
+A new participant device is invited to collaborate on a magic folder
+using the ``magic-folder invite`` command:
+
+
+.. code-block:: console
+
+   $ magic-folder --node-device=alice invite magic: bob
```
vu3rdd, line 90: `--node-directory` instead of `--node-device`?

docs/usage.rst
sajith: Ah, thank you!  Fixed in 2eeb8cf.

docs/usage.rst
sajith: Not sure `export` would be necessary, since no sub-shells would be involved here?  `$ALICE_NODE` should expand in the same bash process.

docs/usage.rst
sajith: Thank you!  Fixed in 29b60b4.

sajith:
> Oh, and I think we decided to just push branches to the main repo instead of via forks (because the CI only appears to run if you push to the main reop anyway). Please remember to delete these branches though :)

Ah, I did not know that.  Pushing to main repo is one less thing to prevent me from doing messy stupid things though!   Hopefully I won't be able to push to `master`?

meejah:
> Hopefully I won't be able to push to master?

Not sure, I can't see that setting :/ (but if you CAN then you can also force-push to right thing again if you accidentally pushed master ;)

Issue #140, opened 2020-06-19
Create an on-disk representation of the Local Snapshot


vu3rdd:
Write the in-memory representation into the disk so that even if the daemon dies or the computer crashes, we pick up these snapshots later for committing into the grid. This also makes offline working possible, while maintaining history.

For the initial version, we can come up with some representation and later iterate on it. We will have to provide compatibility after the first public release of the standalone magic-folder. But until then, we can play with the internal formats and representations.

vu3rdd:
Closed by #157.

PR #157, opened 2020-06-26
140.on disk localsnapshots

vu3rdd:
Address #140

### DIFF
--- a/newsfragments/140.feature
+++ b/newsfragments/140.feature
@@ -1,0 +1,1 @@
+persist LocalSnapshots on the disk to preserve history even while the computer if offline.
--- a/src/magic_folder/magicfolderdb.py
+++ b/src/magic_folder/magicfolderdb.py
@@ -70,6 +70,16 @@ CREATE TABLE local_files
  last_downloaded_uri VARCHAR(256),                -- URI:CHK:...
  last_downloaded_timestamp TIMESTAMP
 );
+
+CREATE TABLE local_snapshots
+(
+ index              INTEGER PRIMARY KEY,
+ foldername         TEXT,                         -- magic folder name that the file lives in
+ filename           TEXT,                         -- file name that the snapshot represents
+ content_path       TEXT,                         -- file path to the stashed contents
+ metadata_blob      BLOB,                         -- a JSON blob corresponding to the metadata
+ parents_local      INTEGER                       -- table index of the local parent snapshot
+);
 """
 
 
--- a/src/magic_folder/magicfolderdb.py
+++ b/src/magic_folder/magicfolderdb.py
@@ -236,3 +236,17 @@ class MagicFolderDB(object):
                                     (path, foldername, serialized_snapshot))
                 action.add_success_fields(insert_or_update=u"update")
             self.connection.commit()
+
+    def get_snapshot(self, path, foldername):
+        """
+        return a serialized blob that corresponds to the given path's latest snapshot.
+        Traversing the parents would give the entire history of snapshots.
+        """
+        self.cursor.execute("SELECT snapshot_blob FROM local_snapshots"
+                            " WHERE foldername=? AND path=?",
+                            (foldername, path))
+        row = self.cursor.fetchone()
+        if not row:
+            return None
+        else:
+            return snapshot_blob[0]
--- a/src/magic_folder/magicfolderdb.py
+++ b/src/magic_folder/magicfolderdb.py
@@ -217,42 +217,49 @@ class MagicFolderDB(object):
                 action.add_success_fields(insert_or_update=u"update")
             self.connection.commit()
 
-    def store_local_snapshot(self, serialized_snapshot, path, foldername):
+    def store_local_snapshot(self, serialized_snapshot, name, foldername):
         """
         Store or update the given serialized form of Local Snapshot for the
-        given file path and folder name.
+        given magicpath of the file (mangled file path) and folder name.
 
         :param str serialized_snashot: A JSON representation of a LocalSnapshot
 
-        :param str path: Relative path of the file whose snapshot is being stored
+        :param str name: magicname of the filepath whose snapshot is being stored
 
         :param str foldername: Magic folder name
         """
         action = STORE_OR_UPDATE_SNAPSHOTS(
-            relpath=path,
+            relpath=name,
             foldername=foldername,
         )
         with action:
             try:
                 self.cursor.execute("INSERT INTO local_snapshots VALUES (?,?,?)",
-                                    (path, foldername, serialized_snapshot))
+                                    (name, foldername, serialized_snapshot))
                 action.add_success_fields(insert_or_update=u"insert")
             except (self.sqlite_module.IntegrityError, self.sqlite_module.OperationalError):
                 self.cursor.execute("UPDATE local_snapshots"
                                     " SET foldername=?, snapshot_blob=?"
                                     " WHERE path=?",
-                                    (foldername, serialized_snapshot, path))
+                                    (foldername, serialized_snapshot, name))
                 action.add_success_fields(insert_or_update=u"update")
             self.connection.commit()
 
-    def get_snapshot(self, path, foldername):
+    def get_snapshot(self, name, foldername):
         """
-        return a serialized blob that corresponds to the given path's latest snapshot.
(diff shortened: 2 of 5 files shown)


src/magic_folder/magicfolderdb.py
```diff
@@ -202,3 +216,39 @@ def did_upload_version(self, relpath_u, version, last_uploaded_uri, last_downloa
                                      relpath_u))
                 action.add_success_fields(insert_or_update=u"update")
             self.connection.commit()
+
+    def store_local_snapshot(self, serialized_snapshot, path, foldername):
+        """
```
meejah, line 221: the args should be documented .. and is "path" a mangled-path or a full absolute path or ..? If it's a filesystem path it sholud maybe be a `FilePath` instance. If it's the mangled path, maybe use `name` (or `snapshot_name`) so we're more-consistent with `LocalSnapshot.name`?

src/magic_folder/snapshot.py
```diff
@@ -164,6 +164,58 @@ def _get_synchronous_content(self):
         with open(self.content_path, "rb") as f:
             return f.read()
 
+    def serialize(self):
+        """
```
meejah, line 168: Since we used `to_json()` elsewhere for "serialize to JSON" maybe "to_json()" and "from_json()" for these methods? Kind of a "bikeshed" comment :/

src/magic_folder/snapshot.py
```diff
@@ -164,6 +164,58 @@ def _get_synchronous_content(self):
...
+        local_snapshot_dict = json.loads(serialized)
+
+        def deserialize_dict(snapshot_dict, author):
+            name = snapshot_dict["name"]
+
+            # deserialize metadata
+            metadata = {}
+            metadata["ctime"] = snapshot_dict["metadata"]["ctime"]
+            metadata["mtime"] = snapshot_dict["metadata"]["mtime"]
+
```
meejah, line 209: Above, we serialized everything that's in metadata; why are we only looking at these two now? In Tahoe it seems that the metadata is pretty open-ended so I think we should probably just say `metadata = snapshot_dict["metadata"]` to take all of it.

src/magic_folder/test/test_snapshot.py
```diff
@@ -45,8 +47,13 @@
     create_local_author_from_config,
     write_local_author,
     create_snapshot,
+    LocalSnapshot,
 )
+from allmydata.util.fileutil import abspath_expanduser_unicode
```
meejah, line 53: IMO we should avoid any new `allmydata.*` imports; can we use `FilePath` instead to get whatever functionality we wanted from this?

src/magic_folder/test/test_snapshot.py
```diff
@@ -99,10 +106,22 @@ def setUp(self):
         self.alice = create_local_author("alice")
         self.stash_dir = mktemp()
         os.mkdir(self.stash_dir)
+
+        # create a magicfolder db
+        self.tempdb=abspath_expanduser_unicode(unicode(mktemp()))
+        os.mkdir(self.tempdb)
```
meejah, line 112: We should avoid using the `unicode` keyword which is python2-only. I think this can just be "`self.tempdb = mktemp()`" though like `stash_dir` above?

src/magic_folder/test/test_snapshot.py
```diff
@@ -99,10 +106,22 @@ def setUp(self):
...
+        dbfile= abspath_expanduser_unicode(u"test_snapshot.sqlite", base=self.tempdb)
+        self.db = magicfolderdb.get_magicfolderdb(dbfile, create_version=(magicfolderdb.SCHEMA_v1, 1))
+
+        self.failUnless(self.db, "unable to create magicfolderdb from %r" % (dbfile,))
+        self.failUnlessEqual(self.db.VERSION, 1)
```
meejah, line 117: I think in new code we should use `.format()`. That said, I'm not sure if we've gotten consensus around any of this kind of string handling stuff :/

src/magic_folder/test/test_snapshot.py
```diff
@@ -255,3 +274,119 @@ def test_snapshots_with_parents(self, content1, content2, filename):
...
+
+        # now modify the same file and create a new local snapshot
+        data2 = io.BytesIO(content2)
+        d = create_snapshot(
+            name=filename,
+            author=self.alice,
+            data_producer=data2,
+            snapshot_stash_dir=self.stash_dir,
+            parents=snapshots,
+        )
```
meejah, line 312: Might be better / more-explicit as `parents=[snapshots[0]]` to make a new list ... but also *maybe*  `create_snapshot` should take care to deep-copy the incoming parents list?

src/magic_folder/test/test_snapshot.py
```diff
@@ -255,3 +274,119 @@ def test_snapshots_with_parents(self, content1, content2, filename):
...
+
+        serialized = snapshots[1].serialize()
+
+        reconstructed_local_snapshot = LocalSnapshot.deserialize(serialized, self.alice)
+
+        self.assertThat(
+            reconstructed_local_snapshot,
+            MatchesStructure(
+                name=Equals(filename),
+            )
```
meejah, line 323: Should assert some things about the rest of the content too (e.g. it should have a single parent).

vu3rdd:
Thanks. I pushed a few commits that address some of the comments. Not very satisfied with the handling of db pathname (basename + "/" + dbfile. May be I should use a real `FilePath` there..

src/magic_folder/magicfolderdb.py
vu3rdd: That seem like a good idea. It was simple enough that, I went ahead and implemented that in this PR.

src/magic_folder/test/test_snapshot.py
vu3rdd: I missed this out in the earlier commits. Fixed now.

src/magic_folder/magicfolderdb.py
vu3rdd: On second thought, I thought I would merge in the changes and submit those changes as a separate PR.

newsfragments/140.feature
```diff
@@ -0,0 +1 @@
+persist LocalSnapshots on the disk to preserve history even while the computer if offline.
```
exarkun, line 1: I like Twisted's style guidelines for news fragments: https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles

More generally, I guess we should have *some* style guide because the fragments are all assembled at release time and if they are consistent the result looks vastly better than if they are not.

(just wrote that on https://github.com/LeastAuthority/magic-folder/issues/164)

vu3rdd: Thanks. I wasn't aware how the news file was being used. I will submit a new PR that fixes the news file.

src/magic_folder/magicfolderdb.py
```diff
@@ -202,3 +214,47 @@ def did_upload_version(self, relpath_u, version, last_uploaded_uri, last_downloa
...
+                self.cursor.execute("INSERT INTO local_snapshots VALUES (?,?)",
+                                    (name, serialized_snapshot))
+                action.add_success_fields(insert_or_update=u"insert")
+            except (self.sqlite_module.IntegrityError, self.sqlite_module.OperationalError):
+                self.cursor.execute("UPDATE local_snapshots"
+                                    " SET snapshot_blob=?"
+                                    " WHERE path=?",
+                                    (serialized_snapshot, name))
+                action.add_success_fields(insert_or_update=u"update")
+            self.connection.commit()
```
exarkun, line 241: I'm afraid that an exception somewhere above will leave an open transaction on this connection.

These two links demonstrate what I think might be a safer pattern:

https://github.com/PrivateStorageio/ZKAPAuthorizer/blob/ace36b198920dbeb745c7c82a9e8b4764d7af91a/src/_zkapauthorizer/model.py#L188-L201
https://github.com/PrivateStorageio/ZKAPAuthorizer/blob/ace36b198920dbeb745c7c82a9e8b4764d7af91a/src/_zkapauthorizer/model.py#L262-L283

vu3rdd: Thanks @exarkun. I will make another PR to fix the problem.

Issue #158, opened 2020-06-27
Better handing of --web-port <number>


sajith:
Currently running `magic-folder --node-directory nodes/alice run --web-port 6000` will print this message:

```
Traceback (most recent call last):
  File "magic-folder/venv/lib/python2.7/site-packages/twisted/application/service.py", line 288, in startService
    service.startService()
  File "magic-folder/src/magic_folder/cli.py", line 1104, in startService
    d = maybeDeferred(do_magic_folder, self.options)
  File "magic-folder/venv/lib/python2.7/site-packages/twisted/internet/defer.py", line 151, in maybeDeferred
    result = f(*args, **kw)
  File "magic-folder/src/magic_folder/cli.py", line 1092, in do_magic_folder
    return maybeDeferred(f, so)
--- <exception caught here> ---
  File "magic-folder/venv/lib/python2.7/site-packages/twisted/internet/defer.py", line 151, in maybeDeferred
    result = f(*args, **kw)
  File "magic-folder/src/magic_folder/cli.py", line 621, in main
    options["web-port"],
  File "magic-folder/src/magic_folder/cli.py", line 705, in from_node_directory
    return cls(reactor, config, webport, magic_folders)
  File "<attrs generated init magic_folder.cli.MagicFolderService>", line 10, in __init__
    self.__attrs_post_init__()
  File "magic-folder/src/magic_folder/cli.py", line 671, in __attrs_post_init__
    serverFromString(self.reactor, self.webport),
  File "magic-folder/venv/lib/python2.7/site-packages/twisted/internet/endpoints.py", line 1731, in serverFromString
    nameOrPlugin, args, kw = _parseServer(description, None)
  File "magic-folder/venv/lib/python2.7/site-packages/twisted/internet/endpoints.py", line 1649, in _parseServer
    getPlugins(IStreamServerEndpointStringParser), endpointType
  File "magic-folder/venv/lib/python2.7/site-packages/twisted/internet/endpoints.py", line 1665, in _matchPluginToPrefix
    raise ValueError("Unknown endpoint type: '%s'" % (endpointType,))
exceptions.ValueError: Unknown endpoint type: '6000'
```

Perhaps adding the `tcp:` prefix to web port would be a small improvement?

exarkun:
[...]

PR #159, opened 2020-06-27
Handle `magic-folder run --web-port <number>`

sajith:
Fix for #158.  It does two things:

* Adds a hint to `magic-folder run --help`.
* Adds `tcp:` prefix to `--web-port` parameter if it is not present.


newsfragments/158.minor
```diff
@@ -0,0 +1 @@
+Prefix `tcp:` to `--web-port` argument when necessary.
```
meejah, line 1: I *think* the text in `.minor` newsfragments isn't ever rendered into the news file? (so might want to change this to `158.bugfix` to see it). `tox -e draftnews` should show was the news will look like .. but, looks like it is currently broken :/

src/magic_folder/cli.py
```diff
@@ -667,6 +668,10 @@ def __attrs_post_init__(self):
             ),
             Agent(self.reactor),
         )
+
+        if self.webport.isdigit():
+            self.webport = "tcp:{}".format(self.webport)
```
exarkun, line 673: Apart from the discussion on the linked ticket, the implementation would be better if instead constructed a `TCP4ServerEndpoint` here instead of doing string mashing.

exarkun:
There is still a UI concern related to this but it's moved now, right?  There is (or soon will be) no more `magic-folder run --web-port ...`.  Instead there is `magic-folder create/migrate --listen-endpoint ...`, I think.

So the issue isn't resolved but this PR is probably obsolete because it makes changes to an option that no longer exists?

sajith:
Even with the expected changes, I'm unsure if `run` sub-command will have an option to override a previously configured endpoint. But what contains in this PR probably is not the right solution and likely obsolete.  Closing for now.

PR #163, opened 2020-06-30
magicfolderdb: store_local_snapshot and get_local_snapshot work with LocalSnapshot instances

### LINKED ISSUE #162: store_local_snapshot and get_snapshot should work with LocalSnapshot instances
[store_local_snapshot](https://github.com/LeastAuthority/magic-folder/blob/master/src/magic_folder/magicfolderdb.py#L218) and [get_snapshot](https://github.com/LeastAuthority/magic-folder/blob/master/src/magic_folder/magicfolderdb.py#L243) currently work with serialized strings. Their input/outputs needs to be converted to/from `LocalSnapshot`. Instead these functions can be made to work directly with `LocalSnapshot` instances.

vu3rdd:
Fixes #162.

### DIFF
--- a/src/magic_folder/magicfolderdb.py
+++ b/src/magic_folder/magicfolderdb.py
@@ -220,29 +220,27 @@ class MagicFolderDB(object):
                 action.add_success_fields(insert_or_update=u"update")
             self.connection.commit()
 
-    def store_local_snapshot(self, snapshot, name):
+    def store_local_snapshot(self, snapshot):
         """
         Store or update the given Local Snapshot for the
         given the magicpath of the file (mangled file path).
 
         :param str snapshot: A LocalSnapshot instance
-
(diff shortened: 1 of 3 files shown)


src/magic_folder/magicfolderdb.py
```diff
@@ -215,19 +220,20 @@ def did_upload_version(self, relpath_u, version, last_uploaded_uri, last_downloa
                 action.add_success_fields(insert_or_update=u"update")
             self.connection.commit()
 
-    def store_local_snapshot(self, serialized_snapshot, name):
+    def store_local_snapshot(self, snapshot, name):
         """
```
meejah, line 224: Can we just use `snapshot.name` in this method instead of passing name here (or, what should the caller pass for "name" if not `snapshot.name`?)

vu3rdd: pushed a commit.

PR #168, opened 2020-07-01
167.db exception safety

### LINKED ISSUE #167: make db operations exception safe
The problem is described in this review discussion: https://github.com/LeastAuthority/magic-folder/pull/157#pullrequestreview-440013621

vu3rdd:
Fixes #167

### DIFF
--- a/src/magic_folder/magicfolderdb.py
+++ b/src/magic_folder/magicfolderdb.py
@@ -24,6 +24,10 @@ from eliot import (
     ActionType,
 )
 
+from functools import (
+    wraps
+)
+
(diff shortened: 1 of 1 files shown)


src/magic_folder/magicfolderdb.py
```diff
@@ -24,6 +24,10 @@
     ActionType,
 )
 
+from functools import (
+    wraps
```
exarkun, line 28: Personally, I like trailing commas here since it minimizes future diffs.  Right now this is purely a style preference so change it only if you want.  (I might try to formalize formatting as some kind of [black configuration](https://pypi.org/project/black/) and make it a CI check someday though.)

PR #169, opened 2020-07-01
Start some dev docs, including info about news fragments

### LINKED ISSUE #164: Add a style guide for news fragments
I like Twisted's style guidelines for news fragments: https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles

More generally, I guess we should have *some* style guide because the fragments are all assembled at release time and if they are consistent the result looks vastly better than if they are not.

exarkun:
Fixes #164

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -1,0 +1,59 @@
+.. _Magic-Folder Development:
+
+Magic-Folder Development
+========================
+
+Process Overview
+----------------
+
+The high-level view of the Magic-Folder development process looks like this:
+
+#. A shortcoming, defect, or new area of functionality is identified relating to the software.
+   For example,
+   perhaps files with names matching a certain pattern are never uploaded to the grid
+   or a developer decides to add a history browsing user experience to the software.
+#. `An issue is filed`_ describing the shortcoming, defect, or new area of functionality.
+   The issue describes how the software will be different than it is now and why this is desirable.
+#. Development work proceeds in a branch (typically created from a recent master branch revision).
+   Prior to this development or in parallel with it,
+   communication within the team helps identify and clear roadblocks early.
+#. When development has progressed sufficiently,
+   the branch is `proposed for review`_.
+#. One or more other developers `review the changes`_.
+#. The previous two steps may repeat a number of times until the changes pass review.
+#. Upon passing review,
+   the changes are `merged into the master branch`_.
+
+
+.. _An issue is filed:
+
+Filing Issues
+-------------
+
+The project uses `GitHub issues <https://github.com/leastauthority/magic-folder/issues>`_ for issue tracking.
+
+.. _proposed for review:
+
+Proposing Changes
+-----------------
+
+To propose a change,
+create a `GitHub Pull Request <https://github.com/leastauthority/magic-folder/pulls>`_ against master.
+
+.. _review the changes:
+
+Reviewing Changes
+-----------------
+
+A number of things should be true before a change is merged into the master branch.
+These things are checked by automation to the greatest extent possible.
+Part of a reviewers job is to verify that the automation has signed off on the changes ("CI is green").
+The first part of the reviewer's job, however, is to ensure the intent of the ticket is satisfied by the changes.
+
+
+  *
+
+.. _merged into the master branch:
+
+Merging Changes
+---------------
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -55,7 +55,7 @@ News Fragments
 One of the mechanical checks performed by continuous integration against proposed changes is the existence of a "news fragment".
 News fragments are assembled at release time by `towncrier <https://pypi.org/project/towncrier/>`_ to contribute to the release announcement.
 News fragments are meant to be user-facing and should have a consistent style.
-News fragments can use a inline formatting directives from reStructuredText.
+News fragments can use any inline formatting directives from reStructuredText.
 
 magic-folder's news fragment style is adapted from `the style guidelines from the Twisted project>`_.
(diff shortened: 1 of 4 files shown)


docs/development.rst
```diff
@@ -0,0 +1,198 @@
...
+  * be in a branch named ``<ticket number>.<some short descriptive text>``.
+  * have appropriate developer-facing and user-facing documentation.
+  * have full line and branch coverage provided by the test suite.
+  * be implemented in a way that is not unnecessarily difficult to understand and maintain.
+  * satisfy all of the mechnical checks performed by the continuous integration system.
+
+News Fragments
+~~~~~~~~~~~~~~
+
+One of the mechnical checks performed by continuous integration against proposed changes is the existence of a "news fragment".
```
sajith, line 55: "mechanical" is mis-spelled here.

docs/development.rst
```diff
@@ -0,0 +1,198 @@
...
+News fragments are assembled at release time by `towncrier <https://pypi.org/project/towncrier/>`_ to contribute to the release announcement.
+News fragments are meant to be user-facing and should have a consistent style.
+News fragments can use a inline formatting directives from reStructuredText.
+
+magic-folder's news fragment style is adapted from `the style guidelines from the Twisted project>`_.
```
sajith, line 60: That trailing `>` would be a problem when rendering this document as HTML, as in:

https://github.com/LeastAuthority/magic-folder/blob/164.dev-docs-etc/docs/development.rst#news-fragments

docs/development.rst
```diff
@@ -0,0 +1,198 @@
...
+  * The entry SHOULD contain a high-level description of the change suitable for end users.
+  * When the changes touch Python code,
+    the grammatical subject of the sentence SHOULD be a Python class/method/function/interface/variable/etc,
+    and the verb SHOULD be something that the object does.
+    The verb MAY be prefixed with "now".
+  * When the changes touch user interface,
+    the grammatical subject of the sentence SHOULD identify the part of the user interface affected
+    and the verb SHOULD be something that the user interface does.
+  * For bugfix,
+    it MAY contain a reference to the version in which the bug was introduced.
```
sajith, line 74: Leading spaces before `*` will make this list render as a blockquote:

https://github.com/LeastAuthority/magic-folder/blob/164.dev-docs-etc/docs/development.rst#id2

docs/development.rst
```diff
@@ -0,0 +1,198 @@
...
+First and foremost,
+the reviewer's job is to ensure the objective of the corresponding ticket has been satisfied.
+
+Some specific areas to which a reviewer can pay attention:
+
+  * Is the implementation unnecessarily difficult for a human reader to understand
+    (and maintain)?
+  * Does the test suite make correct assertions about the behavior of the code under test?
+  * Does the documentation (developer- and user-facing) accurately describe the new behavior?
+
```
sajith, line 185: This list too will be rendered within a blockquote because of leading spaces, as in:

https://github.com/LeastAuthority/magic-folder/blob/164.dev-docs-etc/docs/development.rst#reviewing-changes

docs/development.rst
```diff
@@ -0,0 +1,198 @@
...
+create a `GitHub Pull Request <https://github.com/leastauthority/magic-folder/pulls>`_ against master.
+
+Changes should:
+
+  * have a corresponding ticket.
+  * be in a branch named ``<ticket number>.<some short descriptive text>``.
+  * have appropriate developer-facing and user-facing documentation.
+  * have full line and branch coverage provided by the test suite.
+  * be implemented in a way that is not unnecessarily difficult to understand and maintain.
+  * satisfy all of the mechnical checks performed by the continuous integration system.
```
vu3rdd, line 50: typo - mechnical -> mechanical.

docs/development.rst
```diff
@@ -0,0 +1,198 @@
...
+
+News Fragments
+~~~~~~~~~~~~~~
+
+One of the mechnical checks performed by continuous integration against proposed changes is the existence of a "news fragment".
+News fragments are assembled at release time by `towncrier <https://pypi.org/project/towncrier/>`_ to contribute to the release announcement.
+News fragments are meant to be user-facing and should have a consistent style.
+News fragments can use a inline formatting directives from reStructuredText.
```
vu3rdd, line 58: "an" inline?

docs/development.rst
```diff
@@ -0,0 +1,198 @@
...
+
+magic-folder's news fragment style is adapted from `the style guidelines from the Twisted project>`_.
+The fragment types accepted are canonically defined by the towncrier configuration file in the project root.
+
+Here are a few guidelines which should help you write good news fragments:
```
vu3rdd, line 63: I am wondering if a suggestion to look at the rendered output of `towncrier --draft` can be added here as a final check on how the written comment would be look in the final release note.

(Trying that myself, revealed that we should name the towncrier config file as `pyproject.toml` and not `towncrier.pyproject.toml` -- or I might have done something wrong).

exarkun: Good idea.  `tox -e draftnews` is the easiest way to preview results.  `pyproject.toml` is a mess.

exarkun: (and I added a suggestion to use that tox command to the docs)

docs/development.rst
exarkun: On GitHub, apparently, but not when I do local renders with Sphinx.  But I guess it also works with Sphinx without the leading spaces, so I'll remove them.  I *hope* readthedocs renders like local Sphinx builds and not like GitHub.  If that's true, we should probably treat readthedocs/Sphinx as the authority.  GitHub doesn't really want us to use rst at all and would greatly prefer "github flavored markdown" for maximum vendor lock-in. :)

PR #183, opened 2020-07-06
Caching for NixOS on CircleCI

### LINKED ISSUE #182: Cache most of the build on the NixOS CI job
The NixOS job doesn't use binary wheels so it builds several dependencies every time.  It would be faster if it cached the results of these, a feature which CircleCI offers.

exarkun:
Fixes #182

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -304,6 +304,28 @@ jobs:
 
     steps:
       - "checkout"
+
+      - restore_cache:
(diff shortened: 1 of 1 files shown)


exarkun:
Since this is a CI-configuration change, and that only for how dependencies are cached, only I feel okay self-reviewing and merging.  Feedback welcome nevertheless, if anyone is interested.

Issue #189, opened 2020-07-07
Design configuration storage


meejah:
Currently, magic-folder uses the Tahoe-LAFS "node directory" to store configuration information in. This made a lot of sense when magic-folder was a tahoe subcommand but now that it is a stand-alone project, that makes less sense.

See also #135 

Given that the workflow for someone who has a tahoe-based magic folder is to run `magic-folder migrate` we could decide to have magic-folder's configuration in its own place. There are several pieces of configuration we need right now (and probably more in the future):

- the endpoint we should listen on (currently `--web-port` required command-line option)
- a "stash directory" for files that are to be uploaded (not configured yet)
- the location of our Tahoe client ("node directory" or simply web-api endpoint, currently `--node-directory` required command-line option)
- SQLite database for each magic-folder (currently `private/magicfolder_<name>.sqlite`)
- author keypairs for each magic-folder (currently `private/magicfolder_<name>.privkey`)
- configuration of each magic-folder (currently `private/magic_folders.yaml`)
  - name
  - collective-dircap
  - poll-interval
  - upload-dircap
  - umask
  - (other per-magic-folder config?)

meejah:
fixed by https://github.com/LeastAuthority/magic-folder/pull/200

PR #194, opened 2020-07-08
sketch of config management

meejah:
See #189

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -103,12 +103,11 @@ UI to Change the Config
 
 Writing configuration files is hard, error-prone and not a great
 interface. It also makes it hard (or impossible) to change where or
-how configuration is stored. All configuration shall be changed by
-command-line. Since most command-line tools that interact with
-magic-folder should be thin wrappers over an underlying HTTP API there
-should also be "an HTTP API" to do that. GUI front-ends could use the
-command-line or the HTTP API. Nothing except the HTTP API should be
-touching the actual stored configuration.
+how configuration is stored. All configuration shall be changed by an
+HTTP API. There will also be CLI commands that are thin wrappers of
+the HTTP API. GUI front-ends could use the command-line or the HTTP
+API. Nothing except the HTTP API should be touching the actual stored
+configuration.
 
 (Of course we can't stop users from editing the config files
 themselves, but this should be discouraged).
@@ -123,31 +122,48 @@ command-line option. (Should we have a default? Seems sensible, most
 OSes have a reasonably well-defined place for programs to store
 configuration data).
 
-Data shall be stored in flat files. Data shall be serialized as JSON,
-unless otherwise specified. JSON shall be pretty-printed with 4-space
-indents and newlines.
+All global configuration and state shall be stored in an SQLite3
+database. This database will have tables for global configuration and
+state.
 
 The "configuration directory" will look like this:
 
 - `confdir` (e.g. `~/.config/magic-folder`)
   - `global.sqlite`
-    - a table for configuration (just key/value?)
-      - "endpoint": a Twisted server-endpoint string
-      - "tahoe_node_directory": a path to our Tahoe client's node-dir (or just "tahoe_endpoint")
-      - ... can be expanded for any global config
   - `api_token`: 32 bytes of binary data
-  - `magicfolder_<name>.sqlite`
-    - a table for this folder's configuration (just key/value?)
-        - "author_name": arbitrary name
-        - "author_private_key": 32 bytes signing key
-        - "poll_interval": int, how often to check for updates
-        - "stash_directory": path to our local-snapshot stash directory
-        - ... can be expanded for any per-magic-folder config
-    - a table for this folder's local snapshots
-    - a table for this folders remote snapshot data cache
-    - `stashdir_<name>/`, the default location for stashed file-data for this magic-folder
+  - `magic_folder.pid`: contains the PID of the currently-running process
 
-We could roll all the per-magic-folder state and config into tables in
-the `global.sqlite` database instead. This might make it harder to
-purge data relating to a single magic-folder and make cross-folder
-information leaks easier.
+The `global.sqlite` database will have the following tables:
+
+- "version" with a single row containing "1"
+- "magic_folders"
+  - columns: `name`, `location`
+  - `location` is a local path for that folder's configuration
+- "config"
+  - columns: "api_endpoint", "tahoe_client_url"
+  - `api_endpoint` is a Twisted server string description
+  - `tahoe_client_url` is a URL to our Tahoe-LAFS client's Web-UI
+
+Each "magic folder" location is a directory containing the
+stash-directory and state for that magic-folder. It will look like
+this:
+
+- `<magic folder location>/`
+  - `README`: a file containing information about magic-folder
+  - `state.sqlite`: the state database
+  - `stash/`: the stash-directory containing LocalSnapshot content
+
+The `state.sqlite` for each magic-folder shall contain the following
+tables:
+
+- "version" (will always contain 1 row)
+  - column: "version"
+  - "version" is an int, currently `1`
+- "config" (will always contain 1 row)
+  - columns: "author_name", "author_private_key", "poll_interval"
+  - "author_name" is a string of unicode
+  - "author_private_key" is a 32-byte blob (a NaCl Signing key)
+  - "poll_interval" says how often (in seconds) to poll for updates
+- "local_snapshots"
+  - columns: "ID", "name", "metadata", "content_path", "parents_local"
+  - actually, whatever https://github.com/LeastAuthority/magic-folder/issues/197 says
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -69,8 +69,15 @@ WANT to store is:
   - some state (see below)
 - the node-directory of our Tahoe-LAFS client
   - we may need the `private/api_auth_token`
-  - we will need the `node.url`
-  - (it MAY be sufficient to only store the API endpoint, not node-directory)
+  - we will need the `node.url` (at least)
+  - (it MAY be sufficient to only store the API endpoint (the contents
+    of `node.url`), not node-directory. If it's feasible to only store
+    the API endpoint, we should do that instead of caring about the
+    node-directory. It would probably still be convenient to refer to
+    a Tahoe client by node-directory (but we can just pull out the
+    Web-API endpoint). If we only store the endpoint and the Tahoe
+    config changes, the user would also have to change their
+    magic-folder config.
 - our own API endpoint (a Twisted server endpoint-string,
   e.g. `tcp:1234` or `unix:/tmp/foo`)
 
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -81,13 +81,13 @@ WANT to store is:
 - our own API endpoint (a Twisted server endpoint-string,
   e.g. `tcp:1234` or `unix:/tmp/foo`)
 
-State that we need to store (per magic-folder):
+State that we need to store:
 
 - a secret token that protects the API (only users who can read the
   file can access the endpoint)
-- list of other participants (see note)
-- local cache of remote snapshots
-- local cache of snapshots we haven't yet uploaded
+- list of other participants (per magic-folder) (see note)
+- local cache of remote snapshots (probably per-magic folder)
+- local copy of snapshots we haven't yet uploaded (probably per-magic folder)
   - metadata details of the snapshot
   - "stash directory" for file-data of the snapshot
 
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -1,0 +1,134 @@
+.. -*- coding: utf-8 -*-
+
+.. _config:
+
+Configuration Storage
+=====================
+
+Context and History
+-------------------
+
+`magic-folder` originally was extracted from Tahoe-LAFs where it was a
+sub-command / sub-system. Thus, it inherited Tahoe's method of
+configuration. This is a menagerie of various files in a "node
+directory" including a `tahoe.cfg` INI-style file. There is a
+`private/` sub-directory which often has "extra-private" information
+(like secret keys).
+
+
+Although a magic-folder is necessarily tied to a Tahoe-LAFS client we
+don't *need* to store our configuration there.
+
+We have decided that a user upgrading from a `tahoe magic-folder`
+style Magic Folder to a stand-alone Magic Folder will run
+`magic-folder migrate`. This allows us the freedom to store our
+configuration in any way we like.
+
+
+Legacy Configuration and State
+------------------------------
+
+In versions of Tahoe-LAFS that had magic-folder enabled, the following
+configuration could be used:
+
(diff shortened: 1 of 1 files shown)


docs/config.rst
```diff
@@ -0,0 +1,134 @@
...
+    - `arbitrary_name_of_folder/`
+      - `config.json`
+        - "author":
+          - "name": arbitrary name
+          - "private_key": base64-encoded signing key
+        - "poll_interval": int, how often to check for updates
+        - "stash_directory": path to our local-snapshot stash directory
+        - ... can be expanded for any per-magic-folder config
+      - `local_snapshots.sqlite`: snapshots we haven't yet uploaded
+      - `remote_snapshots.sqlite`: cache of remote snapshot data (ours and other users)
```
exarkun, line 134: Here's a counter-proposal:

There is one SQLite3 database that contains all of these things.

Cons:
* It's marginally harder to inspect and twiddle manually (dev inconvenience)

Pros:
* It's marginally harder to inspect and twiddle manually (molly guard against user tampering)
* It's easier to backup
* You always have a consistent view
* You don't have to worry about ridiculous OS-specific filesystem failure modes (how are folder names encoded, what if you try to delete something that the antivirus software is still scanning, etc)
* You don't have to base64 binary data
* You have a formal declaration of the structure of the configuration (the SQL schema)
* It's faster to query (eg "list of remote snapshots created by Alice sorted by created date")

The only data that seems like it might benefit from exclusion from the SQLite3 database is the API key.  It's a little different from the other data:

* It's meant to be directly exposed to other software
* You can't get it from the HTTP API since you need it to *use* the HTTP API
* It doesn't matter if it gets lost.  Just generate a new one.  Maybe magic-folder should do this *anyway* once in a while.

docs/config.rst
```diff
@@ -0,0 +1,134 @@
...
+  - we may need the `private/api_auth_token`
+  - we will need the `node.url`
+  - (it MAY be sufficient to only store the API endpoint, not node-directory)
+- our own API endpoint (a Twisted server endpoint-string,
+  e.g. `tcp:1234` or `unix:/tmp/foo`)
+
+State that we need to store (per magic-folder):
+
+- a secret token that protects the API (only users who can read the
+  file can access the endpoint)
```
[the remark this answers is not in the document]
meejah: I will re-word (my intention wasn't to change the token to be per-magic-folder, it just looks like that because of the "(per magic-folder)" parenthetical.

docs/config.rst
meejah: Tahoe creates a new API-key on every startup, IIRC. So, yes, changing it is good. I don't know how frequently that should happen, but warner and I agreed "at every startup" sounded like a good start.

It's not clear above if you mean to include the stashed file-data in the database too, but I'm not sure if that is a good idea -- and might also run into weird limits (what if I store a 4GB file in the database)? Another reason to keep the stash-dir separate is that users might want it to be on a different file-system or at least "somewhere else entirely" (I guess I didn't capture "make it configurable" above though).

Another "con" for "everything is in the database" is that upgrading / updating things (adding / deleting config?) might become a little harder?

So here's a hybrid counter-counter proposal :)

 - there is one "global" database for all config and any global state (not including the API key)
 - there is a per-magic-folder database (for config and state of that magic-folder)
 - there is a per-magic-folder stash-directory (and the location is configurable)

It seems "better" to keep the different magic-folders isolated, but I also haven't thought too hard about this (either way). So I guess I could be convinced that "one database" for all global config, any global state with separate tables for each magic-folder's state and config is fine too. That would still leave the stash-dirs separate (with their default location being along with the other config).

meejah: okay, i wrote down the latter in the doc and pushed it.

meejah: > It's faster to query (eg "list of remote snapshots created by Alice sorted by created date")

I think we'd want to keep such queries limited to ONE magic-folder (however, that can be accomplished via separate database OR just separate tables in one database).

I think I still slightly prefer to keep each folder's data more-strictly separated. One compelling reason is that probably makes it easier to purge all data relating to it ("delete files named X") and be confident it's actually gone.

exarkun: > It's not clear above if you mean to include the stashed file-data in the database too

Woops, sorry about that.  I did *not* mean for file contents to be in the database.  It might be interesting to explore SQLite3 as storage for those eventually but I think "regular files" is a better first choice.

> It seems "better" to keep the different magic-folders isolated, but I also haven't thought too hard about this (either way). So I guess I could be convinced that "one database" for all global config, any global state with separate tables for each magic-folder's state and config is fine too.

I think we will have a "list the magic folders that exist" API so, thinking about how that would be implemented ... If all the magic-folders have their config and non-file-content state in one database, it's just one SELECT statement to get the answer.  If there's a database per magic-folder then I guess it would be something like an os.listdir.  Either way probably works but maybe os.listdir seems more error prone (filesystem encoding and permissions and stuff).

If we want to do anything more than just have a list of folder names then the single-database solution starts to look a lot better.  Maybe we want "list of magic-folders and last modified timestamp" or "list of magic-folders sorted by creation time" or "sorted by stored size".  For all of those, it would be nice to have one database to open and issue queries against rather than having to open N different databases and issue the queries against all of them.

The per-magic-folder stash directory does complicate the backup story a little bit.  If we have those and a single SQLite3 database you actually just back up the database to back up your magic-folder state... So that does nullify one of the single-database "Pro" items.

On the details of the SQLite3 schema, I wouldn't have a table per folder.  Instead, there'd still be something like the "local_snapshots" table and presumably some new "remote_snapshots" table and snapshots from all magic-folders would go into these tables.  They'd be differentiated by adding a column referencing the magic-folder they belong to and there'd be a new table holding the other per-magic-folder configuration (stash directory, author name, etc).

meejah: Would one of the above solutions be more obvious if we had an answer to the question, "how separate do different magic-folders need to be?"

My intuition is that they *should* be separate, but that isn't necessarily a compelling reason. And it's not obvious that "there are N databases" necessarily means less risk of the code screwing up and showing you a snapshot / etc from the wrong magic-folder. Some actual user-facing things that might be reasons:

 - backups (already mentioned) might want to keep things separate (e.g. "I want to back up the 'work-stuff' magic-folder but DO NOT want to back up the 'activist-stuff' magic-folder")
 - reliably purging things (e.g. "I want to make sure that everything to do with the 'activist-stuff' magic-folder is deleted")

Maybe a (slightly different) hybrid proposal would be that the global database keeps a table about "magic-folders we know and love" which points to "a directory, somewhere". Inside this directory is the stash-dir and a "state.sqlite" file which contains all the config + state of that one magic-folder.

This means you can back up everything to do with that magic-folder by backing up that one directory. It means you can purge everything [*] to do with that magic-folder by deleting that one directory.

[*] there'd still be an entry in the global state to delete if you did the purging/deleting yourself but ...

exarkun: > I think I still slightly prefer to keep each folder's data more-strictly separated. One compelling reason is that probably makes it easier to purge all data relating to it ("delete files named X") and be confident it's actually gone.

Let's make this scenario more explicit.  Is this what you're thinking?  Alice puts something in her magic-folder.  As a result, a cap for the content ends up her local magic-folder instance's state (on her local disk somehow, presumably).  Later she deletes the file from her magic-folder.  Then her machine falls into the hands of an attacker.  Can the attacker recover the deleted file now?  I guess they can because we're adding history so the cap still exists even though the file was deleted.  Maybe a slightly different scenario, then.  Alice deletes the whole magic-folder containing the content she wants to repudiate.  If a user does that then it's perhaps reasonable for them to expect that the contents of the magic-folder are no longer recoverable from the local disk state?

Am I on the right track?  If so, we might have to do some kind of filesystem-specific scrubbing on deletion to really achieve the goal - and I bet database schema only makes a marginal difference (ie, we could accomplish this scrubbing whether we have a single database or a database per-magic-folder).

meejah: I guess another reason might be: I would expect we'll have something like a `MagicFolder` instance per magic-folder and then nothing has to share databases (i.e. each `MagicFolder` has a database that is just for it). Maybe this doesn't matter though (especially because everything is in one process anyway).

meejah: Yes I was thinking of deleting and purging an entire magic-folder (not just one file) and thinking that "delete these files" is more likely to achieve that with standard tools (rather than "run `magic-folder purge X` and trust us that it deletes every row in the one database that relates to that"). So I guess the target audience is "someone technical at an org who may feel better about using their own 'scrubbing tool' to remove a folder".

meejah: Here is an attempt to articulate my intuition: we have separate magic-folders for some reason. I think it's reasonable to expect that the different stuff doesn't ever get mixed. (Or, to flip this around, why not just have "participants" and "files" and you can share anything with any participants? Why bother with named magic-folders?)

So thinking about managing different magic-folders, it seems like it's more likely to meet user-expectations if they are actually-separate as much as feasible. So I hear the point about "more flexible queries" (e.g. yes, maybe you _do_ want to list folders by "last update time" or similar) but also this could easily lead to mixing things together in API calls etc and essentially making the separation between different magic-folders less clear. I think I'm just extending this idea of "different magic-folders are totally separate" right down to the storage-layer.

exarkun: > Here is an attempt to articulate my intuition

Thanks!  I'm going to try to follow along.

> we have separate magic-folders for some reason

Sounds good.

> I think it's reasonable to expect that the different stuff doesn't ever get mixed.

I agree.

> Or, to flip this around, why not just have "participants" and "files" and you can share anything with any participants? Why bother with named magic-folders?

This part I'm not so sure about.  Why is the primitive sharing unit "a directory and all its transitive children" instead of "a file"?  I think I might not quite understand how this is a flipped version of the previous statement.

> So thinking about managing different magic-folders, it seems like it's more likely to meet user-expectations if they are actually-separate as much as feasible.

In some ways, this makes sense.  The sharing configuration for the folders is different, so it makes sense to have a user interface which reflects that difference.  Files put in folder A go to group A', files put in folder B go to group B', etc.

I mentioned some user-facing functionality above ("list all of Alice's snapshots across all folders I have access to") but the decision about whether to use one database or many doesn't directly impact the availability of that user-facing functionality.  We could either implement it or not implement it no matter what we decide here.

To what extent does the user have expectations about the underlying storage magic-folder uses?  In most cases I'd bet there is no expectation at all.  But for sufficiently technically adept users, perhaps there is.  I think it's hard to guess what expectations are going to be common (and perhaps it's also not the case that the most common expectation is the one to cater to, anyway).

"magic-folder has *a* database where it keeps its state" seems like a perfectly unsurprising expectation for someone to have, though.  This may just reveal my particular software experiences, though.

> So I hear the point about "more flexible queries" (e.g. yes, maybe you do want to list folders by "last update time" or similar) but also this could easily lead to mixing things together in API calls etc and essentially making the separation between different magic-folders less clear.

I think maybe this conclusion reaches a bit too far.  True, we could have API calls that blur the lines between folders but we don't have to.  Right now, all of the actual APIs we've proposed implementing are limited to a single magic-folder and the HTTP resource hierarchy reflects this (with resources accessible at `/v1/<just one folder name>/<more stuff>`).

I think I can see how having to open more than one database in order to interact with more than one magic-folder would make it harder to unintentionally mix information from different magic-folders together - by making it harder and more awkward to work with such information from more than one magic-folder.

Perhaps this is the core disagreement?  I don't want to go around handing out footguns but I also don't want to make useful things harder than they need to be.  Single-database seems clearly easier than multi-database.  To allay concerns about unintentional mixing of information from different folders, I would make sure the internal Python APIs for accessing the data have appropriate guardrails on them (there should *be* Python APIs for accessing the data, for one thing - not random application code constructing its own SQL statement).  Since the only cross-folder requirement we even have right now is "list the magic folders that exist" we probably shouldn't even build any Python APIs that work on more than one magic-folder yet.  So application code using those APIs should be completely safe.  Of course, there could be a bug in SQLite3 that gives back incorrect rows - but there could just as easily be a bug in the filesystem that gives back the contents of the wrong file, so I don't think that fact helps one way or another.

> By extension, this could/should/might extend into the API, too, encouraging users of that API to treat different magic-folders completely separately.
 
(re-ordering your comments because responding to this next makes more sense, I think.)

It seems like this is in agreement with the response I wrote just above.

> I think I'm just extending this idea of "different magic-folders are totally separate" right down to the storage-layer. 

This part I think is mistaken.  SQLite3 is a semi-arbitrary level in the storage layer.  In some ways, the Python APIs I mentioned above (for accessing magic folder state, and which have appropriate guardrails) are also the storage layer and extending this idea to that point seems reasonable.  Why pick SQLite3 instead of that layer?  Or why pick SQLite3 instead of the filesystem layer?  All of the SQLite3 databases are going to live on the same filesystem unless we take further steps.  And below that there's the disk - someone *really* serious may want to scrub the physical device to eliminate one folder while keeping other folders.

meejah: Okay, I think we agree on the vast majority of the above. (The API layout, the Python APIs, ..)

The only disagreement seems to be about "should magic-folder separation be expressed on the filesystem", approximately?

> All of the SQLite3 databases are going to live on the same filesystem unless we take further steps.

Not necessarily; if we follow the last of my suggestions (where the global state says "magic-folder `foo` has data at `/path/to/data`") then a user could choose to have `foo` on one file-system (e.g. `/mnt/encrypted_drive`) and another magic-folder `bar` on a different file-system (e.g. `/mnt/work_stuff`). The only record of `bar` in the "main" filesystem (wherever the global state is) would be the table listing the magic-folders. All the "important" data/state would be in `/mnt/work_stuff/state.sqlite` and as local stashed files in that same directory.

exarkun: > Okay, I think we agree on the vast majority of the above. (The API layout, the Python APIs, ..)
> 
> The only disagreement seems to be about "should magic-folder separation be expressed on the filesystem", approximately?
> 
> > All of the SQLite3 databases are going to live on the same filesystem unless we take further steps.
> 
> Not necessarily; if we follow the last of my suggestions (where the global state says "magic-folder `foo` has data at `/path/to/data`") then a user could choose to have `foo` on one file-system (e.g. `/mnt/encrypted_drive`) and another magic-folder `bar` on a different file-system (e.g. `/mnt/work_stuff`). The only record of `bar` in the "main" filesystem (wherever the global state is) would be the table listing the magic-folders. All the "important" data/state would be in `/mnt/work_stuff/state.sqlite` and as local stashed files in that same directory.

Okay.  That seems like it provides real user-facing security/privacy improvements compared to the single database design.  I'm game to try the multi-database approach.

meejah: Ideally we could reach out to some contacts that "want to use something like this" and ask them if that would make any difference at all to them. In any case, I'll adjust the writeup to reflect that and we can go from there if/when we have better feedback.

Purely mechanically, should we aim to merge just this documentation or should it stay in this branch until this branch also grows an implementation so that the documentation matches reality?

exarkun: I think merging the docs and implementation as one PR makes sense.  We might want to put in more (or less) detail about the implementation but not know that until the implementation is underway or complete.

meejah:
This is superseded by this PR which includes implementation: https://github.com/LeastAuthority/magic-folder/pull/200

Issue #161, opened 2020-06-30
A new python API to "manually" trigger an upload


vu3rdd:
In order to simplify testing and overall integration of various pieces of snapshot implementation, it would be nice to have a python function to manually add files to the  upload pipeline instead of using the watchdog/inotify. Internally, watchdog can also rely on the same API to add files to the upload queue and trigger the pipeline.

exarkun:
I think this API would essentially be:

```
POST /v1/magic-folder/<folder name>/files/<relative path>/changed

200 OK
```

The result would be that, for the folder indicated by `<folder name>`, a file that already exists on the filesystem that can be found at `<relative path>` (relative to the folder's local filesystem root) will have a new snapshot created from the filesystem state at that time (or slightly later) and uploaded.

This would be the equivalent to the action we would hope to take when an INotify-based system delivers a CLOSE_WRITE event, presumably.

PR #195, opened 2020-07-09
Create LocalSnapshot instances

### LINKED ISSUE #192: Create LocalSnapshots due to file-changes
Part of #141 

Given input from an API (or watchdog?) that a new version of a file in our magic-folder should be created we:

- create a LocalSnapshot instance
- serialize it to the database
- pass the LocalSnapshot instance to the upload queue (

meejah:
Fixes #192 (nearly).

This creates and serializes LocalSnapshot instances (in a queue) when told to.

Still needs:

- [ ] ~some way to call `add_files` (API call, ideally)~ (https://github.com/LeastAuthority/magic-folder/issues/161)
- [x] test coverage for the error-cases
- [x] test coverage for "add a directory" (should add all children)

Also maybe the API should just be `add_file` (i.e. a single file-or-directory) to avoid one level of looping. (inotify or the API could just call it multiple times).


meejah:
codecov seems wrong again; I get 93% from local tool.

vu3rdd:
@meejah how are you measuring the coverage locally?

meejah:
`git diff master.. | cuv diff -`
So, "personal local tool that may very well also have bugs ;)": https://github.com/LeastAuthority/magic-folder/pull/195#issuecomment-658145188

meejah:
(codecov commented on another PR I pushed before any of the CI runs finished, so .. I don't know how it did that .. and/or it's not counting integration-test coverage -- but i'm not counting that usually either)

vu3rdd:
At this point, `git diff master | cuv diff -` says 91% coverage. Some of the error states are not covered yet. Will work on that now.

vu3rdd:
OK, 96% coverage now.

vu3rdd:
`git diff master | cuv diff - | less -R` says 97% now.

src/magic_folder/test/test_local_snapshot.py
```diff
@@ -0,0 +1,261 @@
...
+    def test_add_a_file_twice(self, filename, content1, content2):
+        foo = self.magic_path.child(filename)
+        with foo.open("w") as f:
+            f.write(content1)
+
+        self.snapshot_creator.add_files(foo)
+        self.snapshot_creator.startService()
+
+        foo_magicname = path2magic(foo.asTextMode('utf-8').path)
+        stored_snapshot1 = self.db.get_local_snapshot(foo_magicname, self.author)
```
exarkun, line 243: The implementation of this test feels a little fishy to me, starting around here.  The test essentially requires that `startService` synchronously create the first `foo` snapshot.  It does now but that doesn't seem like an essential part of the interface.  If snapshot creation changes to be asynchronous (as was the initial intent) then this test will break by failing to find a snapshot at this point.  It would be nice if that didn't happen.

The reason it is awkward to test this here is that the public interface of `snapshot_creator` doesn't support learning when snapshots have been created nor does it have an interface to force snapshots to be created right away.

Perhaps that means `snapshot_creator` should have a different interface.  One idea that comes to mind is that the thing which provides `IService` and does the asynchronous job processing (`_process_queue`) should be one unit which can be tested in isolation and the snapshot creating piece should be a different unit which can also be tested in isolation.  The snapshot creating piece then can have (and needs to have) an interface for making work happen so that the `_process_queue` piece can drive it).  Then *all* these tests for snapshot creation logic can skip over all of the service stuff and async queue stuff and just test the snapshot creation.

src/magic_folder/magic_folder.py
```diff
@@ -2134,3 +2182,149 @@ def trap_conflicts(f):
...
+        """
+        self._service_d.cancel()
+        d = self._service_d
+        self._service_d = None
+        return d
+
+    def add_files(self, *paths):
+        """
+        Add each item in `paths` to our queue. If an item is itself a
+        directory, we add all its children. If the path does not exist
```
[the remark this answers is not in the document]
meejah: Because we only process "files" in this queue. Perhaps that is the wrong design and we should process "files or directories" and ... do the recursion/walking when we "process" a directory?

src/magic_folder/magic_folder.py
```diff
@@ -2134,3 +2182,149 @@ def trap_conflicts(f):
...
+                raise ValueError(
+                    "every argument to upload_files must be a FilePath"
+                )
+
+            def add_if_file(p):
+                if p.isfile():
+                    try:
+                        p.segmentsFrom(self.magic_path)
+                    except ValueError:
+                        raise ValueError(
```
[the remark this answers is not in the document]
meejah: At some point I speculated (at least to myself, not sure if I wrote that down anywhere) that maybe the right API is just `add_file` and something else figures out the rest. If the `add_file` had a basically straight-through to the HTTP API version, then "outside stuff" (GUI, CLI) could do whatever is appropriate and we don't have to depend / mentally-model whatever magic-folder is going to do.

In this thinking, "add_file()" with a directory is simply an error. If you want all the children added, you do that. (This still doesn't prevent us from creating an `add_directory()` method with matching HTTP API in the future which does ... something).

Another issue with "magically recurse directories" is figuring out **when** the walking occurs .. do we walk it once (upon adding) or are we expected to always watch it for new files. The "more automatic" stuff is essentially what the "legacy" magic-folder implementation tries to accomplish, but I feel this has caused a lot of weird code ... and maybe also paint for e.g. GridSync to figure out what it's going to do.

I guess personally I'm trying to work outwards from a "very simple thing that works" so my inclination here is to just not handle directories at all (it's an error) and deal with that elsewhere (e.g. something outside this queue will be handling inotify/watchdog stuff which probably should be the thing that deals with directories).

So I guess **all that hand-waving boils down to a concrete suggestion** here: get rid of the walking, get rid of the multiple paths. Change this API to `add_file(FilePath)` and it must be a child of our magic-folder (or it's an error). If it's a directory, that's an error too. We leave any inotify/walking/watchdog stuff to some other piece. We could hook up an HTTP API (or not) but definitely not as part of this PR.

vu3rdd:
OK, I think I addressed most of the comments (except a couple - one of them, a very important LocalSnapshotCreator interface change). But hopefully the rest is all in and if anyone has time for another pass, they can have a look.

meejah:
I have looked over the commits from @vu3rdd since I last looked: :+1: 

I did a little cleanup (and some more `FilePath` instead of `os.*` things). I took at stab at adding some documentation of "snapshots" too. From my perspective, this can be merged (when CI goes green).

docs/snapshots.rst
```diff
@@ -0,0 +1,75 @@
...
+into the Tahoe-LAFS Grid. A Local Snapshot exists only on one device's
+local disk (so other participants cannot see it yet).
+
+A Snapshot has one or more parents, forming a Directed Asyclic Graph
+(DAG). Snapshots are signed "as" they are uploaded into the Grid:
+first the actual content is uploaded as an immutable, yielding a
+read-only capability-string; then this string and other metadata is
+signed. The signature, metadata and pointer to the content constitutes
+the Remote Snapshot and itself is uploaded into the Tahoe Grid,
+represented as an immutable directory.
```
exarkun, line 24: This section might benefit from explaining some more of the "why" to go along with the "what".

docs/snapshots.rst
```diff
@@ -0,0 +1,75 @@
+.. -*- coding: utf-8 -*-
+
+.. _snapshots:
+
+
```
exarkun, line 5: Perhaps we could introduce the purpose of this document here, including who the intended audience is (developers, I guess).

src/magic_folder/magic_folder.py
```diff
@@ -364,14 +397,17 @@ def from_config(cls, client_node, name, config):
                 local_dir_config,
                 base=client_node.config.get_config_path(),
             ),
-            dbfile=abspath_expanduser_unicode(db_filename),
+            db=database,
             umask=config["umask"],
             name=name,
             downloader_delay=poll_interval,
+            clock=reactor,
+            snapshot_service=snapshot_service,
```
exarkun, line 405: If the only thing `MagicFolder` does with `snapshot_service` is add it as a child of itself, maybe this caller should just add it as a child and `MagicFolder` can be unaware of it?

src/magic_folder/magic_folder.py
```diff
@@ -2134,3 +2190,137 @@ def trap_conflicts(f):
...
+                parents = [parent_snapshot]
+
+            # need to handle remote-parents when we have remote
+            # snapshots
+
+            # when we handle conflicts we will have to handle multiple
+            # parents here (or, somewhere)
+
+            relpath_u = path.asTextMode(encoding="utf-8").path
+            SNAPSHOT_CREATOR_PROCESS_ITEM.log(relpath=relpath_u)
```
exarkun, line 2231: This would most likely make more sense as an action than as a message.  That way it could contain the `create_snapshot` call and anything it logs (or its failure).

src/magic_folder/test/cli/test_magic_folder.py
```diff
@@ -264,17 +276,38 @@ def _clean(ign):
 
     def init_magicfolder(self, client_num, upload_dircap, collective_dircap, local_magic_dir, clock):
         dbfile = abspath_expanduser_unicode(u"magicfolder_default.sqlite", base=self.get_clientdir(i=client_num))
+        try:
+            db=get_magicfolderdb(dbfile, create_version=(SCHEMA_v1, 1))
```
exarkun, line 280: Elsewhere the return value of `get_magicfolderdb` is checked against `None`, as if errors are signaled that way.  I wonder which it is.

src/magic_folder/test/test_local_snapshot.py
```diff
@@ -0,0 +1,336 @@
...
+    """
+    A way to test LocalSnapshotService with an in-memory database.
+    """
+    processed = attr.ib(default=attr.Factory(dict))
+    def process_item(self, path):
+        Message.log(
+            message_type=u"memory-snapshot-creator:process_item",
+            path=path.asTextMode("utf-8").path,
+        )
+        if path not in self.processed.keys():
```
exarkun, line 68: `x in dict` and `x in dict.keys()` are the same except `x in dict` is faster and uses less memory.

src/magic_folder/test/test_local_snapshot.py
```diff
@@ -0,0 +1,336 @@
...
+            # record number of parents
+            self.processed[path] = 0
+        else:
+            num_parents = self.processed[path]
+            self.processed[path] = num_parents + 1
```
exarkun, line 73: Also this is a typical idiom:

```
    self.processed[path] = self.processed.get(path, 0) + 1
```

src/magic_folder/test/test_local_snapshot.py
```diff
@@ -0,0 +1,336 @@
...
+            succeeded(Always()),
+        )
+
+        self.assertThat(
+            self.snapshot_service.stopService(),
+            succeeded(Always())
+        )
+
+        self.assertThat(self._snapshot_creator.processed.get(foo), Not(Equals(None)))
+        self.assertThat(self._snapshot_creator.processed.get(foo), Equals(0))
```
exarkun, line 127: If it equals 0 it must not equal None so I think the first assertion here is redundant.

src/magic_folder/test/test_local_snapshot.py
```diff
@@ -0,0 +1,336 @@
...
+
+    @given(lists(path_segments().map(lambda p: p.encode("utf-8")), unique=True),
+           lists(binary(), unique=True))
```
exarkun, line 130: Probably doesn't matter if the contents are unique?

What does matter is the the lists are the same length - which they won't currently always be since the separate `lists` strategies are free to pick their own lengths.  The solution here is probably to use the `data` strategy to draw examples imperatively inside the body of the test.

```
@given(lists(...), data()):
def test_add_multiple_files(self, filenames, data):
    for filename in filenames:
        content = data.draw(binary())
        ...
```

`data.draw(binary())` will return an example like the one that would have been passed to the function if you'd passed `binary()` to the `given` decorator.

src/magic_folder/test/test_local_snapshot.py
```diff
@@ -0,0 +1,336 @@
...
+
+        self.assertThat(
+            len(self._snapshot_creator.processed.keys()),
+            Equals(len(files))
+        )
+
+        self.assertThat(
+            sorted(self._snapshot_creator.processed.keys()),
+            Equals(sorted(files))
+        )
```
exarkun, line 170: If the processed dict's keys equal the sorted files list then their lengths must be equal, also, so the length assertion seems redundant.

src/magic_folder/test/test_local_snapshot.py
```diff
@@ -0,0 +1,336 @@
...
+            f.write(content1)
+
+        self.snapshot_service.startService()
+        self.snapshot_service.add_file(foo)
+
+        with foo.open("wb") as f:
+            f.write(content2)
+
+        # it should use the previous localsnapshot as its parent.
+        self.snapshot_service.add_file(foo)
```
exarkun, line 232: I think there should be an assertion that the previous add_file call succeeded somewhere before this assertion.  If the first add_file hasn't happened yet then there will be no parent for the second.

src/magic_folder/test/test_local_snapshot.py
```diff
@@ -0,0 +1,336 @@
...
+        )
+
+    def setup_example(self):
+        """
+        Hypothesis-invoked hook to create per-example state.
+        Reset the database before running each test.
+        """
+        # XXX: Not sure if this function should exist at all and even if
+        # so, it shouldn't exist in magicfolderdb module perhaps?
+        self.db._clear_snapshot_table()
```
exarkun, line 278: It's probably fine for the short while longer that magicfolderdb is going to exist.

src/magic_folder/test/test_local_snapshot.py
```diff
@@ -0,0 +1,336 @@
...
+        if path not in self.processed.keys():
+            # record number of parents
+            self.processed[path] = 0
+        else:
+            num_parents = self.processed[path]
+            self.processed[path] = num_parents + 1
+        return defer.succeed(None)
+
+
+class LocalSnapshotServiceTests(SyncTestCase):
```
exarkun, line 77: A big advantage of splitting the service from the snapshot creator is that it makes testing easier.  A lot of the tests in this class that have to do with creating snapshots should move to the `LocalSnapshotCreator` test case (or else they are already redundant with tests there).  The tests for `LocalSnapshotService`  Can be much simpler.  The service doesn't know anything about snapshots or parents and has zero logic to dealing with either.  It can be completely tested without a database, without an author, etc.

`MemorySnapshotCreator` can really literally just append to a list to prove the service is behaving correctly.  The list should always be equal to exactly the valid paths that have been passed to `add_file`, in the order they were passed.

src/magic_folder/test/test_local_snapshot.py
```diff
@@ -0,0 +1,336 @@
...
+    def test_create_snapshot_twice(self, filename, content1, content2):
+        """
+        If a snapshot already exists for a file, adding a new snapshot to it
+        should refer to the existing snapshot as a parent.
+        """
+        foo = self.magic_path.child(filename)
+        with foo.open("wb") as f:
+            f.write(content1)
+
+        self.snapshot_creator.process_item(foo)
```
exarkun, line 319: We should check that the returned Deferred has fired before continuing.

src/magic_folder/test/test_local_snapshot.py
vu3rdd: Thanks. I didn't know about data() strategy.

src/magic_folder/magic_folder.py
meejah: It probably WILL want to do other things (e.g. "status API" will want access to it -- or at least to the creator-of-local-things queue). I'm kind of unclear what the right pattern for "a bunch of `IService` things really is, though; maybe we can gain access by asking for the service by name? Or put the arg back as above if/when we actually-do need access?

I don't have any strong feelings on where the "setServiceParent" call goes ..

exarkun: I think removing it and putting it back if we need it is the right path.  We don't know exactly what we'll need.  Better to leave the field wide open for when someone actually sits down to figure it out.

vu3rdd:
I think I have addressed all the comments now.

PR #199, opened 2020-07-10
Authorized /v1 resource hierarchy

### LINKED ISSUE #198: Authorize v1 HTTP API calls using a "bearer token" and the Authorization header
There is currently queryarg/form post-based token authorization for the old API.  We don't like this.  It forces the token into the request URL or requires the endpoint accept a POST.  The former is a security concern, the latter is too limiting and awkward to use.

Create the root of the `/v1` resource hierarchy and require a bearer token (same value as before) in order to access it and all its children.

exarkun:
Fixes #198


src/magic_folder/test/test_web.py
```diff
@@ -570,3 +586,126 @@ def get_status(self):
...
+            one element then it creates a bad request with multiple
+            authorization header values.
+
+        :param [unicode] child_segments: Additional path segments to add to the
+            request path beneath **v1**.
+        """
+        # We're trying to test the *unauthorized* case.  Don't randomly hit
+        # the authorized case by mistake.
+        assume([good_token] != bad_tokens)
+
```
meejah, line 620: Should this be something more like `assume(good_token not in bad_tokens)`..? Or is `bad_tokens` always a 1-element list?

src/magic_folder/web.py
```diff
@@ -19,6 +21,115 @@
...
+        MagicFolderWebApi(get_magic_folder, get_auth_token),
+    )
+    root.putChild(
+        b"v1",
+        BearerTokenAuthorization(
+            _v1_resource,
+            get_auth_token,
+        ),
+    )
+    return root
```
meejah, line 52: This looks like the root will have two resources side-by-side (`/v1` and `/api`) .. I would assume the intent is to have `/v1/api` with `api` a child of `BearerTokenAuthorization` instead of root?

src/magic_folder/web.py
```diff
@@ -19,6 +21,115 @@
...
+
+        If it does not, return an ``Unauthorized`` resource.
+        """
+        if _is_authorized(request, self._get_auth_token):
+            # Authorization checks out, let the protected resource do what it
+            # will.
+            return self._resource.getChildWithDefault(path, request)
+        # Don't let anything through that isn't authorized.
+        return Unauthorized()
+
```
meejah, line 98: Aside: is `getChild` just not-really-used? `Resource`'s implementation seems to simply always return a no-child object. (I guess the ask here is "do we need to implement `getChild` too, or not?)

src/magic_folder/web.py
```diff
@@ -41,6 +151,36 @@ def error(request, code, message):
...
+    :param (IO bytes) get_auth_token: A callable which returns the
+        authorization token value which is required for requests to be
+        authorized.
+
+    :return bool: True if and only if the given request contains the required
+        authorization materials.
+    """
+    authorization = request.requestHeaders.getRawHeaders(u"authorization")
+    if authorization is None or len(authorization) == 0:
+        return False
```
meejah, line 173: Could maybe just be `if not authorization`?

src/magic_folder/web.py
```diff
@@ -41,6 +151,36 @@ def error(request, code, message):
...
+    if len(authorization) > 1:
+        return False
+    auth_token = get_auth_token()
+    expected = u"Bearer {}".format(auth_token).encode("ascii")
+    return timing_safe_compare(
```
meejah, line 178: We're depending on this being "not just binary data" on disk then I guess? (It looks like tahoe-lafs -- whose token we're currently re-using -- is currently using urlsafe-base64 encoding on-disk so that should be fine)

src/magic_folder/test/test_web.py
exarkun: `bad_tokens` is any length.  The reason I did the (non-)equality test is to allow `bad_tokens` to equal things like `[good_token, bad_token]` and include those in the rejected case.  If somebody shows up with two `Authorization` headers, something weird is happening and it seems like fail-closed is the right move.  I'm open to being convinced otherwise though.

src/magic_folder/web.py
exarkun: It will.  `/api` is all of the old API stuff though.  I just tried to leave it in place.  I expect we'll delete it once enough of the new API is done.  Actually, "enough" doesn't need to be very much since the only thing the old API exposes is that buggy "status" endpoint I think.  Anyway, the intent here was "leave `/api` as it is, add a new `/v1` which requires `Authorization`-header-based authorization for itself and all children" and later on we'll hang more and more stuff off of `/v1`.

src/magic_folder/web.py
exarkun: I forgot this and had to go look it up.  `IResource` only includes `getChildWithDefault`.  `getChild` is a convenience helper provided by `Resource` alongside `putChild`.  Real resource traversal calls `getChildWithDefault` which checks to see if there was a matching `putChild` and only calls `getChild` if there wasn't.

So an `IResource` wrapper needs to ignore `getChild` and just deal with `getChildWithDefault`.  `self._resource` might override `getChild` and that's fine as long as `self._resource` is a `Resource` subclass: its `getChildWithDefault` will call its `getChild` if appropriate.

src/magic_folder/web.py
exarkun: Yea ... I started thinking about encoding and then thought there was no reason ... then Hypothesis told me my headers were being populated with incompatible garbage.  So, indeed, we do need the token to be "HTTP header safe".

The approach taken by the code now is to ignore this fact and just require the token read from disk to be okay.  I'd be okay with adding some safeguards here.  What would make sense?  Automatically encode the token read from disk?  Then it doesn't matter what's it disk, we can make it safe.  Clients have to agree with that encoding, of course.

src/magic_folder/test/test_web.py
meejah: Ahh, I see. Okay yeah that makes sense.

src/magic_folder/web.py
meejah: Ah, okay.

src/magic_folder/web.py
meejah: I think the *easiest* thing is to just say the token is "literally all the bytes in the file on disc" and we also ensure that's "safe" -- I did just adjust the config-PR to reflect this (so "our" on-disk token will also use urlsafe-base64 like tahoe does). This way clients inter-operating don't have to figure out how to correctly encode "whatever they found on disk" (e.g. 32 bytes) and instead just load it all and stuff it in a header.


No action for this PR, but: we should probably double-check with our cryptographers about this whole scheme. Tor for example uses the on-disk token as a key to an HMAC (of a fixed string plus a server nonce). But, that's a two-step handshake. I guess we'd have to decide if "a thing that can sniff local traffic but NOT read local files" is a threat we care about...

exarkun: Filed #202

PR #200, opened 2020-07-13
189.configuration with implementation

meejah:


meejah:
(how is it possible that codecov is the first reported status? local tool says 91%)

meejah:
(Not sure what's up with the debian9 failures..)

meejah:
CodeCov is clearly just wrong: it is counting changes to `cli.py` around `_item_status` against this patch/project coverage but I haven't changed any of that code. Is it associating the wrong builds with the wrong PRs or something??

meejah:
(weird, again codecov is the first to complete..)

meejah:
come on, codecov: how did the "patch" coverage only hit 77% but somehow that also raised the project coverage to 87%??

meejah:
I marked this is as ready-for-review because we *could* merge it now -- or we could put arbitrarily more implementation in here (API, `magic-folder migrate`, tie into all the other commands, etc)

meejah:
I can't even tell what the "details" link in `codecov/patch` checker is trying to show me; it's got a whole bunch of lines that aren't touched by this PR in the visualization .. but maybe that's what it is supposed to do?

meejah:
See also some previous discussion here: https://github.com/LeastAuthority/magic-folder/pull/194

docs/config.rst
```diff
@@ -0,0 +1,180 @@
...
+.. _current:
+
+Where and How To Store the Configuration
+----------------------------------------
+
+Somewhat mirroring Tahoe-LAFS's design, configuration shall all be
+found in a single directory. This directory can be specified via
+command-line option. (Should we have a default? Seems sensible, most
+OSes have a reasonably well-defined place for programs to store
+configuration data).
```
exarkun, line 130: is there an API we can call, yet, that just returns the right path to use for all platforms? :/

docs/config.rst
```diff
@@ -0,0 +1,180 @@
...
+The `global.sqlite` database will have the following tables:
+
+- "version" with a single row containing "1"
+- "magic_folders"
+  - columns: `name`, `location`
+  - `location` is a local path for that folder's configuration
+- "config"
+  - columns: "api_endpoint", "tahoe_client_url"
+  - `api_endpoint` is a Twisted server string description
+  - `tahoe_client_url` is a URL to our Tahoe-LAFS client's Web-UI
```
exarkun, line 152: I guess the reason we might want the node directory instead of the http api root url is that some nodes are configured to listen on a random port  every startup?  And, specifically, gridsync configures its tahoe nodes to do this.  It might be annoying to have to update database state every time the data node restarts.

Or, we could mitigate this by *allowing* `tahoe_client_url` as a cli argument to `magic-folder run` and having it override the database value?

Or we could have a node directory in the db and read the node.url ourselves...

Or some combination of the above, I guess.

exarkun: Pointing at the node directory means the magic-folder process could re-read a changed node.url itself and follow tahoe-lafs restarts/reconfigurations easily.

docs/config.rst
```diff
@@ -0,0 +1,180 @@
...
+  - `stash/`: the stash-directory containing LocalSnapshot content
+
+The `state.sqlite` for each magic-folder shall contain the following
+tables:
+
+- "version" (will always contain 1 row)
+  - column: "version"
+  - "version" is an int, currently `1`
+- "config" (will always contain 1 row)
+  - columns: "author_name", "author_private_key", "poll_interval"
```
exarkun, line 170: This elemt is redundant I guess, since all the columns are listed below (and there are more than these three).

docs/config.rst
```diff
@@ -0,0 +1,180 @@
...
+  - "author_name" is a string of unicode
+  - "author_private_key" is a 32-byte blob (a NaCl Signing key)
+  - "stash_path" is a local path where we stash data awaiting upload
```
exarkun, line 173: We could make this a path relative to the directory containing this database, making the whole config directory somewhat easily relocateable (or at least not imposing an additional barrier to relocation).

exarkun: otoh since it involves rewriting configuration we probably want some kind of `magic-folder move-folder-state` command rather than just letting people `mv` stuff, so the path could always be rewritten by that tool.

integration/test_general_cli.py
```diff
@@ -1,5 +1,5 @@
 from os import mkdir
-from os.path import join
+from os.path import join, exists
```
exarkun, line 2: Maybe we could gradually drift towards `twisted.python.filepath` and away from `os.path`.

integration/test_general_cli.py
```diff
@@ -44,3 +44,26 @@ def test_web_port_required(request, reactor, temp_dir, introducer_furl):
...
+            "init",
+            "--config", node_dir,
+            "--listen-endpoint", "tcp:1234",
+            "--tahoe-url", "http://localhost:3456",
+        ],
+    )
+    yield proto.done
+
+    assert exists(join(node_dir, "global.sqlite"))
+    assert exists(join(node_dir, "api_token"))
```
exarkun, line 69: If this PR also adds some `magic-folder show-config` command that dumps the config (json?) then the test can assert that the output config matches the input config and remain a bit more oblivious to the exact internal representation.

src/magic_folder/cli.py
```diff
@@ -136,6 +144,57 @@
...
+
+        # required args
+        if self['listen-endpoint'] is None:
+            raise usage.UsageError("--listen-endpoint / -l is required")
+        if self['tahoe-url'] is None:
+            raise usage.UsageError("--tahoe-url / -t is required")
+
+        # validate
+        if exists(self['config']):
+            raise usage.UsageError("Directory '{}' already exists".format(self['config']))
```
exarkun, line 178: Dang I wish anyone at all was working on twisted.python.usage.  These are such basic, common checks; it would be great if you could just declare them.

src/magic_folder/cli.py
```diff
@@ -993,6 +1052,7 @@ class MagicFolderCommand(BaseOptions):
     stderr = sys.stderr
 
     subCommands = [
+        ["init", None, InitializeOptions, "Initialize a Magic Folder daemon."],
         ["create", None, CreateOptions, "Create a Magic Folder."],
```
exarkun, line 1056: I wonder if "init" vs "create" will be memorable for folks (whoever the folks that are directly interacting with this CLI are...).  I don't have any suggestions for different words to use, though.

src/magic_folder/initialize.py
```diff
@@ -0,0 +1,38 @@
+# Copyright 2020 Least Authority TFA GmbH
+# See COPYING for details.
+
+"""
+Implements the magic-folder init command.
+"""
+
+from twisted.python.filepath import (
+    FilePath,
```
exarkun, line 9: :sunglasses:

src/magic_folder/initialize.py
```diff
@@ -0,0 +1,38 @@
...
+        an appropriate exception is raised.
+    """
+
+    create_global_configuration(
+        FilePath(config_dir),
+        listen_endpoint,
+        tahoe_url,
+    )
+
+    return 0
```
exarkun, line 38: `succeed(0)` perhaps?

src/magic_folder/test/test_cli.py
```diff
@@ -55,3 +67,20 @@ def test_good(self):
...
+    Confirm operation of 'magic-folder initialize' command
+    """
+
+    def setUp(self):
+        super(TestInitialize, self).setUp()
+        self.temp = FilePath(self.mktemp())
+
+    def tearDown(self):
+        super(TestInitialize, self).tearDown()
+        rmtree(self.temp.path, ignore_errors=True)
```
exarkun, line 83: `FilePath.remove` is also a thing.  Additionally, `mktemp` creates the temporary directory beneath `_trial_temp` and trial deletes old `_trial_temp` directories when it starts up a new run.  This behavior makes me feel okay about tests leaving behind all of their temporary files.  They won't grow without bound and it's nice to automatically have them available for inspection instead of having to go out of your way to get them.

src/magic_folder/test/test_config.py
```diff
@@ -0,0 +1,166 @@
...
+from ..snapshot import (
+    create_local_author,
+)
+
+
+class TestGlobalConfig(SyncTestCase):
+
+    def setUp(self):
+        super(TestGlobalConfig, self).setUp()
+        self.temp = FilePath(self.mktemp())
```
exarkun, line 36: A common thing to be broken is handling of non-ASCII paths.  It would be nice to exercise `create_global_configuration` with one of those.

src/magic_folder/test/test_config.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def tearDown(self):
+        super(TestGlobalConfig, self).tearDown()
+        rmtree(self.temp.path)
+
+    def test_create(self):
+        create_global_configuration(self.temp, "tcp:1234", "tcp:localhost:5678")
```
exarkun, line 43: It's supposed to return a Deferred that succeeds with 0, maybe that's an okay assertion to make here?  Although it's a bit on the weak side.  Maybe this doesn't add much on top of the other happy-path tests - `test_load_db`, `test_rotate_api_key`, etc?

src/magic_folder/config.py
```diff
@@ -0,0 +1,419 @@
...
+CREATE TABLE version
+(
+    version INTEGER  -- contains one row, set to 1
+);
+
+CREATE TABLE magic_folders
+(
+    name          TEXT PRIMARY KEY,  -- UTF8 name of this folder
+    location      TEXT               -- UTF8 path to this folder's configuration/state
+);
```
exarkun, line 56: I *think* UTF8 bytes is the right choice here?  I'm not sure if there are some nasty surprises waiting on macOS or Windows though.  I think it wouldn't hurt to be explicit about these being bytes and not unicode (though "UTF8" pretty strongly implies bytes).  SQLite3 "type affinity" is a bit odd.  This is a Python 3 transcript but the basic idea holds for Python 2.  Notice the types of the 3 returned rows.

```
>>> import sqlite3
>>> conn = sqlite3.connect(":memory:")
>>> curs = conn.cursor()
>>> curs.execute("CREATE TABLE foo ( a TEXT )")
<sqlite3.Cursor object at 0x7f0b9f771d50>
>>> curs.execute("INSERT INTO foo VALUES (?)", (u"\N{SNOWMAN}",))
<sqlite3.Cursor object at 0x7f0b9f771d50>
>>> curs.execute("INSERT INTO foo VALUES (?)", (u"snowman",))
<sqlite3.Cursor object at 0x7f0b9f771d50>
>>> curs.execute("INSERT INTO foo VALUES (?)", (b"snowman",))
<sqlite3.Cursor object at 0x7f0b9f771d50>
>>> curs.execute("SELECT * FROM foo")
<sqlite3.Cursor object at 0x7f0b9f771d50>
>>> curs.fetchall()
[('☃',), ('snowman',), (b'snowman',)]
>>> 
```

src/magic_folder/config.py
```diff
@@ -0,0 +1,419 @@
...
+_magicfolder_config_schema = """
+CREATE TABLE version
+(
+    version INTEGER  -- contains one row, set to 1
+);
+
+CREATE TABLE stash
+(
+    path TEXT    -- the path to our stash-directory
+);
```
exarkun, line 76: Is this a left-over?  It seels like `config.stash_path` represents the same thing.

src/magic_folder/config.py
```diff
@@ -0,0 +1,419 @@
...
+    stash_path           TEXT,              -- local path for stash-data
+    collective_dircap    TEXT,              -- read-capability-string
+    upload_dircap        TEXT,              -- write-capability-string
+    magic_directory      TEXT,              -- local path of sync'd directory
+    poll_interval        INTEGER            -- seconds
+);
+
+CREATE TABLE local_snapshots
+(
+    id VARCHAR(256) PRIMARY KEY, -- identifier (hash of .. stuff)
```
exarkun, line 91: `VARCHAR(256)` is identical to `TEXT` in SQLite3.

src/magic_folder/config.py
```diff
@@ -0,0 +1,419 @@
...
+    :param unicode tahoe_client_url: the Twisted client endpoint
+        string where we will contact our Tahoe LAFS client WebUI.
+
+    :returns: a GlobalConfigDatabase instance
+    """
+    if basedir.exists():
+        raise ValueError(
+            "'{}' already exists".format(basedir.path)
+        )
+    mkdir(basedir.path)
```
exarkun, line 119: Race condition here.  `mkdir` and handle the EEXIST instead?  Also there's `basedir.createDirectory` and `basedir.makedirs`.

src/magic_folder/config.py
```diff
@@ -0,0 +1,419 @@
...
+
+    # explain what is in this directory
+    with basedir.child("README").open("w") as f:
```
exarkun, line 122: maybe mode `"wt"` and unicode strings for the text

src/magic_folder/config.py
```diff
@@ -0,0 +1,419 @@
...
+            return cursor.fetchone()[0]
+
+    @api_endpoint.setter
+    def api_endpoint(self, ep_string):
+        # confirm we have a valid endpoint-string
+        from twisted.internet import reactor  # uhm...
+        # XXX so, having the reactor here sucks. But if we pass in an
+        # IStreamServerEndpoint instead, how can we turn that back
+        # into an endpoint-string?
+        serverFromString(reactor, ep_string)
```
exarkun, line 283: Yea, all that stuff you wrote.  There is no going from an IStreamServerEndpoint to a string, though.  The interface is not expressive enough.

And the endpoint parser is private so you can't really validate the endpoint in any way except with `serverFromString`.  That is pretty lame.  Maybe most of them don't use the reactor for parsing/initialization and we could get away with passing `None`? :/  Still pretty lame.

src/magic_folder/config.py
```diff
@@ -0,0 +1,419 @@
...
+                "'{}' does not exist".format(magic_path.path)
+            )
+        if state_path.exists():
+            raise ValueError(
+                "'{}' already exists".format(state_path.path)
+            )
+
+        mkdir(state_path.path)
+        stash_path = state_path.child("stash")
+        mkdir(stash_path.path)
```
exarkun, line 377: It would be nice to (eventually) make this function atomic to simplify error recovery.  If we manage to make the state path and/or the stash path but then fail to populate the database schema, the magic-folder has not been created but you cannot try again with `create_magic_folder` without doing some manual cleanup of what should be internal state.

src/magic_folder/config.py
```diff
@@ -0,0 +1,419 @@
...
+            database=connection,
+        )
+
+        # add to the global config
+        with self.database:
+            cursor = self.database.cursor()
+            cursor.execute(
+                "INSERT INTO magic_folders VALUES (?, ?)",
+                (name, state_path.path)
+            )
```
exarkun, line 418: This could fail if someone is doing something wacky - trying to create two magic-folders with the same name but different paths.  They'll race to this point and then one will succeed and the other will fail and leave intermediate state lying around.  Making the whole function atomic would probably address this.

docs/config.rst
meejah: `#python` says https://pypi.org/project/appdirs/

docs/config.rst
meejah: okay, sounds good. also storing node-directory is definitely *safer* too (e.g. for anything else we forgot about that we need from tahoe still).

src/magic_folder/cli.py
meejah: There is always `Click` :)

src/magic_folder/cli.py
meejah: My secret plan is to re-think all the CLI commands .. but not necessarily all at once.

In some ways, maybe in makes sense for `init` to be `magic-folder create` and then there's a  `magic-folder add` or similar to create a new magic-folder inside the ... "magic folder daemon" (I don't have a better term yet) ... or we could just re-name `create` to `add` (now .. or later) and leave `init`. In any case, we probably need a common term for "magic-folder the whole daemon process" versus "magic-folder, the thing you can add many of to the daemon process".

src/magic_folder/config.py
meejah: Mostly I'm looking at other SQLite3 code for inspiration; if there's something besides `TEXT` that we should use here (or some other comment / `.encode()` calls) I'm all ears :)

src/magic_folder/config.py
meejah: this was the source of hypothesis errors; i went with "wb" and a `.encode()` call ... but i'm not very confident that's correct.

src/magic_folder/test/test_config.py
meejah: This now tests different filenames (and `create_global_configuration` is synchronous).

meejah:
I believe I've addressed all the review comments. I didn't take any action on the migration code yet (that could be a follow-up? It seems the strategy there is to make all the table-creation code "migration code" so initial creation is just like "migrate from version 0 to version 1"?)

meejah:
There is now a `magic-folder migrate` (to exercise more of the code "for real") and a `magic-folder show-config` (so the integration test can have better asserts).

integration/test_general_cli.py
exarkun: Okay, good points.  `twist` has been a struggle all the way and if it's not even giving us what we want (eg ability to spit out clean json) maybe we should stop fighting with it and use something else.  Not right now, though, I suppose.  So perhaps we can just have a ticket to capture this and deal with it later.

meejah:
(oops, missed adding a file)

meejah:
Local tool shows 91% coverage.

PR #204, opened 2020-07-15
Exercise unicode path segments more

### LINKED ISSUE #203: LocalSnapshot.content_path is bytes but it also gets serialized into JSON
This raises an exception if the content is anything but ASCII-compatible bytes.

exarkun:
Fixes #203 

The implementation was fine already but the case wasn't tested so this primarily changes tests.

### DIFF
--- a/src/magic_folder/test/test_web.py
+++ b/src/magic_folder/test/test_web.py
@@ -682,7 +682,7 @@ class AuthorizationTests(SyncTestCase):
         while segments_remaining:
             name = segments_remaining.pop()
             resource = Resource()
-            resource.putChild(name, branch)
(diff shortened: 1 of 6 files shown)


src/magic_folder/test/test_web.py
```diff
@@ -118,18 +121,16 @@ class StatusTests(AsyncTestCase):
     """
     Tests for ``magic_folder.status.status``.
     """
-    @given(folder_names(), absolute_paths().map(FilePath))
+    @given(folder_names(), absolute_paths_utf8().map(FilePath))
```
vu3rdd, line 124: ooh, that's nice. Didn't know we can call other functions inside `@given`. Nice!

PR #206, opened 2020-07-17
Add `GET /v1/magic-folder` endpoint for listing Magic Folders

### LINKED ISSUE #205: Add an HTTP API to list the magic folders that are managed by a Magic-Folders instance
Accept a *GET* request to */v1/magic-folder* and return a JSON-encoded list of magic folders that the instance manages.

eg:

```
GET /v1/magic-folder HTTP/1.1

200 OK
Content-Type: application/json

{“folders”: [
    {“name”: “folder name”, “local-path”: “absolute local path”}, 
]}
```

Afterwards, the "list" command-line interface should be updated to use this instead of implementing the logic itself.

exarkun:
Fixes #205

### DIFF
--- a/src/magic_folder/test/test_web.py
+++ b/src/magic_folder/test/test_web.py
@@ -30,6 +30,7 @@ from hypothesis.strategies import (
     lists,
     text,
     binary,
+    dictionaries,
     sampled_from,
 )
 
@@ -91,6 +92,7 @@ from .matchers import (
 from .strategies import (
     path_segments,
     folder_names,
+    absolute_paths,
     absolute_paths_utf8,
     tahoe_lafs_dir_capabilities as dircaps,
     tahoe_lafs_chk_capabilities as chkcaps,
@@ -773,12 +775,32 @@ def treq_for_folder_names(auth_token, names):
 
     :return: An object like the ``treq`` module.
     """
+    return treq_for_folders(auth_token, dict.fromkeys(names, {u"directory": None}))
+
+
+def treq_for_folders(auth_token, folders):
+    """
+    Construct a ``treq``-module-alike which is hooked up to a Magic Folder
+    service with Magic Folders like the ones given.
+
+    :param unicode auth_token: The authorization token accepted by the
+        service.
+
+    :param folders: A mapping from Magic Folder names to their configurations.
+        These are the folders which will appear to exist.
+
+    :return: An object like the ``treq`` module.
+    """
     state = MagicFolderServiceState()
-    for name in names:
-        state.add_magic_folder(name, {}, object())
+    for name, config in folders.items():
+        state.add_magic_folder(name, config, object())
 
     root = magic_folder_resource(state, lambda: auth_token)
     return StubTreq(root)
+
+
+def magic_folder_config_for_local_directory(local_directory):
+    return {u"directory": local_directory}
 
 
 class ListMagicFolderTests(SyncTestCase):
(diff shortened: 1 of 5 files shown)


src/magic_folder/test/test_web.py
```diff
@@ -710,3 +738,149 @@ def get_magic_folder(name):
...
+        JSON-encoded list of Magic Folders.
+
+        :param dict[unicode, unicode] folders: A mapping from folder names to
+            local filesystem paths where we shall pretend the local filesystem
+            state for those folders resides.
+        """
+        treq = treq_for_folders(
+            auth_token, {
+                name: magic_folder_config_for_local_directory(path)
+                for (name, path)
```
meejah, line 858: nit: shouldn't need the brackets here I think?

src/magic_folder/web.py
```diff
@@ -98,17 +98,45 @@ def getChildWithDefault(self, path, request):
...
+    """
+    Implement the ``/v1/magic-folder`` HTTP API hierarchy.
+
+    :ivar MagicFolderServiceState _magic_folder_state: The Magic Folder state
+        to serve.
+    """
+    _magic_folder_state = attr.ib()
+
+    def __attrs_post_init__(self):
+        Resource.__init__(self)
```
meejah, line 126: Should this be a `super()` incantation instead since we have `object` too?

src/magic_folder/web.py
```diff
@@ -98,17 +98,45 @@ def getChildWithDefault(self, path, request):
...
+        Render a list of Magic Folders and some of their details, encoded as JSON.
+        """
+        request.responseHeaders.setRawHeaders(u"content-type", [u"application/json"])
+        return json.dumps({
+            u"folders": list(
+                {u"name": name, u"local-path": config[u"directory"]}
+                for (name, config)
+                in sorted(self._magic_folder_state.iter_magic_folder_configs())
+            ),
+        })
```
meejah, line 139: "In general" I like adding `indent=4` or similar to get slightly-more-human-readable output; costs a few more bytes but might make debugging easier (for integrations too, not just us). I guess browsers do formatting too, though, and there's `jq` or others for cli users....

src/magic_folder/test/test_web.py
exarkun: True.  I find they help readability but I know that in other parts of the syntax, I find extraneous parens harm readability... It's pretty subjective I guess.  I think I'd like to leave them but I'm also up for adopting black (which seems to be the de facto formatting winner for Python) and making everyone do what it says (with CI/tool assistance).

src/magic_folder/web.py
exarkun: Hm.  I don't know what would happen with `super()` here - specifically because it's not `MagicFolderAPIv1.__init__` that's upcalling `Resource.__init__`.  Maybe it's fine.  I also just realized I was afraid of calling the attrs-generated `__init__` twice but on further thought that seems impossible because it's generated on *this* class not on a super class...?  I'll try changing it and see if something blows up.

src/magic_folder/web.py
exarkun: Maybe some `magic_folder._json.dumps` could encode this policy for all JSON use throughout the project?

src/magic_folder/web.py
exarkun: It seems to work but I'm having trouble convincing myself it's really the right thing.  This is specifically trying to call `Resource.__init__` because the attr.s-generated `__init__` didn't call it.  `Resource.__init__` is pretty clearly calling the right thing.  `super(...).__init__` is a lot less clear to me.  This is basically just a different way to say what I said in my previous comment, I guess.

It seems like the ideal would be for the attr.s-generated `__init__` to make this super() call itself.  It wouldn't surprise me if there are boogeymen hiding behind that idea though...

I'm going to leave this for now but happy to revisit it later.

PR #209, opened 2020-07-21
Address the tahoe-lafs/cryptography RSA regression

### LINKED ISSUE #208: Tahoe-LAFS 1.14.0 and Cryptography 3.0 have a fight, magic-folder loses
Cryptography 3.0 rejects RSA public exponents other than 3 and 65537.  Tahoe-LAFS 1.14.0 uses RSA public exponent 17 for keys it generates.  This causes lots of magic-folder tests to fail.

https://github.com/tahoe-lafs/tahoe-lafs/pull/748 fixes this in Tahoe-LAFS (for some version newer than 1.14.0).  This fix won't immediately be available in a release and because of time constraints we may not want to bump magic-folder's Tahoe-LAFS dependency up to a new version anyway (magic-folder uses a lot of private Tahoe-LAFS Python APIs, even figuring out if magic-folder works with a newer version of Tahoe-LAFS is a job).

We could pin an older cryptography and maybe we should *also* do that but eventually we will upgrade Tahoe-LAFS and cryptography and we know that the result is changing the RSA public exponent.  So, for now, having magic-folder monkey-patch the Tahoe-LAFS fix into Tahoe-LAFS 1.14.0 is probably a good and expedient solution.

exarkun:
Fixes #208


exarkun:
> looks good .. but why are all the circle-ci jobs commented-out?

The integration tests failed the first run :(  I forgot about them.  I don't see a non-insane way to monkey-patch the stand-alone Tahoe-LAFS processes they launch... so they remain broken.

Plan B.  Require `cryptography<3`.  This involves rebuilding all of the CI images. :/  So jobs commented out while I get that working out to avoid burning CI credits.

PR #215, opened 2020-07-27
Delete the status endpoint and supporting code

### LINKED ISSUE #214: Get rid of the old status system and API
We're planning to remove the old status API anyway since it's buggy and doesn't give us all the information we need even when it is working right.  We were planning to keep it until we had a replacement ready but the old implementation is increasingly in the way of new development.  In particular, its deep reliance on specific irrelevant details of the old configuration system and its extremely fragile and complicated test suite mean that keeping it working any longer is going to be a disproportionately large effort.

Beneath it, there's a whole status tracking system that doesn't provide accurate or complete information and basically serves no purpose without the old API sitting on top of it.

Let's delete all of this code now before we waste any more time maintaining it.

exarkun:
Fixes #214


meejah:
Once again I'm convinced codecov is "just plain wrong": how can deleting a bunch of code cause coverage to go down?

src/magic_folder/test/strategies.py
```diff
@@ -243,17 +243,14 @@ def queued_items():
     """
     Build ``QueuedItem`` instances.
     """
-    def an_item(path, progress, size, when):
-        item = QueuedItem(path, progress, size)
-        item.set_status('queued', when)
-        return item
+    def an_item(path, progress, size):
+        return (path, progress, size)
```
vu3rdd, line 247: we can delete `an_item()` function, can't we?

PR #219, opened 2020-07-29
downloader 

### LINKED ISSUE #128: download of files based on snapshots
With snapshots, each client would periodically scan each other's DMD and download the snapshots (until it finds its own current snapshot by walking up through the DAG).

meejah:
Fixes #128

### DIFF
--- a/docs/downloader.rst
+++ b/docs/downloader.rst
@@ -1,14 +1,14 @@
 
 .. -*- coding: utf-8 -*-
 
-.. _config:
+.. _downloader:
 
-Downloader Design
-=================
+Downloader Operation
+====================
 
-This describes code to accomplish remote to local synchronization in
-"the Leif Design" (see :ref:`Leif's Proposal: Magic-Folder
-"single-file" snapshot design`).
+This describes the general operation of the remote to local
+synchronization as outlined in "the Leif Design" (see :ref:`Leif's
+Proposal: Magic-Folder "single-file" snapshot design`).
 
 We do not describe the synchronization of the list of participants in
 a magic-folder. This list will consist of at least:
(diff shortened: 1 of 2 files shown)


meejah:
Do we have Heisencoverage here, too? How can the patch-coverage be fine (100%) but project coverage not-fine, according to codecov?

docs/downloader.rst
```diff
@@ -0,0 +1,157 @@
+
+.. -*- coding: utf-8 -*-
+
+.. _config:
+
+Downloader Design
```
exarkun, line 6: To be really useful, I imagine a design PR/review needs to consider some more concrete items.  Here's a list of ideas I have for what such items could be - they're just ideas, we might want to have a broader discussion about the kind of boxes a design PR like this one should try to check.  The more general idea here might be that I think we have at least a halfway decent shared understanding of what a code review should accomplish.  I'm not sure we know yet what a design review should accomplish.

* What are the public interfaces involved (adding, removed, changed)?  In this case, we might want to consider the HTTP API (probably the "status" API that doesn't exist yet) since that's probably the only *really* public interface that's related to downloader functionality.  Will the downloader contribute some kind of information to that API?  What kind? 
  Since it doesn't exist yet I don't think the PR that resolves #128 has to _implement_ anything for the status API but it would be good to consider it.
* What concrete data is operated on by the proposed design?  Does the downloader need to read or write some of our existing databases?  Does it need schema changes?  Which parts of the system already have the data it needs?  If the data doesn't exist, what parts of the system could be expanded to contain that data?
* If there are new long-lived objects required by the design, where do they fit into the implementation?  Are new services added to the service hierarchy?  Are existing objects being given new responsibilities?
* What does the data flow for the new functionality look like?  eg "the download polling interval starts in the per-folder configuration database, is loaded into a `MagicFolderConfig`, the config is used to initialize a `MagicFolder` service which reads the polling interval and creates an instance of a new `Downloader` service which it attaches to itself.  The `Downloader` runs a `LoopingCall` using the polling interval.  On the polling interval, the downloader calls a new function "download_new_stuff", which ..."

Lacking some concrete stuff like this, I think a "design review" phase risks becoming an unproductive time-sync.  It adds more round-trips between developers to the development process and more overall work so it needs to add some real value to make up for this.  Details like those suggested by the list above bring the design closer to an implementation in a way that makes it possible to provide more constructive feedback.  Without them, it's hard to have much of an idea about what the design really is - at least for me.  Also, a lot of times my fairly high-level, abstract ideas - similar to "The capability strings of snapshots to download will be given to a second service that is responsible for actually downloading them. These will be written into a local database" end up running into cold realities about circular dependencies or missing data which render them unworkable.  A positive review on an unworkable design is also not of much use to anyone.

docs/downloader.rst
```diff
@@ -0,0 +1,157 @@
...
+- an arbitrary ``name`` for each participant
+- a directory read-capability for each participant ("personal DMD")
+  - in this directory is a flattened representation of every file in
+    the Magic Folder (similar to the Tahoe 1.14.0 design) with each
+    entry pointing to a Snapshot. That is, there are no
+    sub-directories to recurse into, all files are listed by
+    downloading just the "personal DMD" capability of that participant.
+
+"A Snapshot" is a single version of a single file. It is represented
+by an immutable directory and contains:
```
exarkun, line 25: This data description is good.  I feel like it might belong somewhere else.  It's shared with the uploader and any other processes that want to interact with the data since they must all agree on it.  Perhaps a separate "model" or "data structures" document that holds this stuff would be good, then uploader, downloader, and anything else could reference it.

docs/downloader.rst
```diff
@@ -0,0 +1,157 @@
...
+A service awaits capability-strings of Snapshots to download. For each
+one, the function ``create_snapshot_from_capability`` is used to
+download the capability and return a ``RemoteSnapshot`` instance.
+
+This ``RemoteSnapshot`` should be serialized to a local cache in the
+magic-folder's state database.
+
+We also arrange to make local filesystem changes. This might require
+waiting to download more ``RemoteSnapshots`` if it has any parents
+that aren't cached.
```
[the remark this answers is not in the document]
meejah: Okay, that's fine .. my thinking here is that we merge this design-doc and then do other PRs to make the system look like it. Are you saying you want to split up this document so it only has "PR-sized" bits change at a time, or...?

meejah:
I've re-based this and taken out all the signature stuff. I also added some class sketches and pseudo-code style descriptions of what they do.

docs/downloader.rst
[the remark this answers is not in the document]
meejah: (No reply in nearly a year, resolving).

docs/downloader.rst
meejah: For me the point of trying to do a "design review" is more-so to avoid a situation where during say the "real" review the reviewer decides "these interfaces are all wrong" or "we should do it <completely different way>" etc. because then way less time has already been sunk.

If you don't think there's enough information here to say "sure, this seems likely to be workable" versus "nah, this is bad" then ... I guess it is a waste of time after all?

meejah:
superseded by https://github.com/LeastAuthority/magic-folder/pull/381

PR #223, opened 2020-07-31
Helper to convert api_endpoint into DecodedURL

### LINKED ISSUE #222: Expose the Magic-Folder daemon HTTP API root on GlobalConfigDatabase
The CLI and any other consumers of the HTTP API will want to know where to find it.  Make this easily found using the `GlobalConfigDatabase`.

exarkun:
Fixes #222 

There didn't seem to be any particular reason to tie this tightly to GlobalConfigDatabase.api_endpoint so I didn't.  It's just a helper you can use with that value.

### DIFF
--- a/src/magic_folder/_endpoint_parser.py
+++ b/src/magic_folder/_endpoint_parser.py
@@ -1,0 +1,68 @@
+# Copyright 2020 Least Authority TFA GmbH
+# See COPYING for details.
+
+"""
+Turn Twisted endpoint description strings into HTTP(S) URLs.
+"""
+
+from hyperlink import (
+    URL,
+)
+
+def endpoint_description_to_http_api_root(endpoint_description):
+    """
+    Parse a Twisted endpoint description string and return a ``DecodedURL`` an
+    HTTP client could use to talk to an HTTP server listening on that
+    endpoint.
+
+    This currently supports only **tcp** and **ssl** endpoints.
+
+    :param unicode endpoint_description: The endpoint description string.
+
+    :return DecodedURL: A URL for reaching the given endpoint.
+    """
+    parts = endpoint_description.split(u":")
+    return _ENDPOINT_CONVERTERS[parts[0].lower()](parts[1:])
+
+
+def _tcp_to_http_api_root(parts):
+    """
+    Construct an HTTP URL.
+    """
+    port, host = _get_tcpish_parts(parts)
+    return URL(
+        scheme=u"http",
+        host=host,
+        port=port,
+    ).get_decoded_url()
+
+
+def _ssl_to_https_api_root(parts):
+    """
+    Construct an HTTPS URL.
+    """
+    port, host = _get_tcpish_parts(parts)
+    return URL(
+        scheme=u"https",
+        host=host,
+        port=port,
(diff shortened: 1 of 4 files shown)


src/magic_folder/_endpoint_parser.py
```diff
@@ -0,0 +1,68 @@
...
+    interface = kwargs.get(u"interface", u"127.0.0.1")
+    if interface == u"0.0.0.0":
+        interface = u"127.0.0.1"
+    return port_number, interface
+
+
+_ENDPOINT_CONVERTERS = {
+    u"tcp": _tcp_to_http_api_root,
+    u"ssl": _ssl_to_https_api_root,
+}
```
sajith, line 68: Perhaps error handling could be improved here?   As in, what happens when `endpoint_description` is not really an endpoint?

I just learned that `magic-folder migrate --listen-endpoint 6001 --node-directory ../path/to/node` accepts that non-endpoint ar argument without complaining.  Thus magic folder configuration will have just `6001` for `api_endpoint`, and passing that on will result in an error here.

Of course `migrate` too probably could some error checking...

src/magic_folder/config.py
```diff
@@ -40,6 +50,11 @@
     atomic_makedirs,
 )
 
+# Export this here since GlobalConfigDatabase is what it's for.
+from ._endpoint_parser import (
+    endpoint_description_to_http_api_root,
+)
+
```
sajith, line 58: Would it make sense to add an additional helper in `GlobalConfigDatabase`, like this:

```python
    def http_api_root(self):
        """
        Base URL for HTTP API, derived from endpoint describtion.
        """
        return endpoint_description_to_http_api_root(self.api_endpoint)
```
Perhaps consumers of the code in `config.py` can avoid some duplication that way?

src/magic_folder/_endpoint_parser.py
exarkun: Yea, it would be good to reject that early.  `create_global_configuration` should probably reject it and then the migrate command could handle that exception and report something meaningful to the cli user.

By the time it makes it into the database, it should really be valid and no one else in the system should have to worry about validation.

src/magic_folder/config.py
exarkun: It might make sense, I'm not sure.  It might also come down to taste.  I almost made it a method but decided against it based on the idea that GlobalConfigDatabase's job is to allow *state* to be stored and retrieved, not to provide all of the different convenient interfaces to that state.  Adding one accessor is hardly a terrible burden but maybe it's a step in the direction we don't want to go, so I'm inclined to leave it off for now and then add it later if it turns out it would be a significant benefit (we could also consider adding the above except as a free function instead of a method which might be 99% of the convenience without expanding `GlobalConfigDatabase`'s interface.

src/magic_folder/_endpoint_parser.py
exarkun: Filed https://github.com/LeastAuthority/magic-folder/issues/226

src/magic_folder/config.py
sajith: That makes sense.  Thank you.

PR #225, opened 2020-08-03
A service that pulls LocalSnapshots from db and commits them to a Tahoe-LAFS grid

### LINKED ISSUE #193: Upload LocalSnapshot instances to Tahoe
Part of #141 

A queue of LocalSnapshots instances shall be uploaded to Tahoe. See #192 for the part that creates the LocalSnapshot instances. See #191 for the part that provides functions to do the uploads.

This will be an `IService` that awaits new work from the "create LocalSnapshots" queue. This queue should be filled from the on-disk representations at startup. This should:

- upload the LocalSnapshot to Tahoe (yielding a RemoteSnapshot)
- delete the LocalSnapshot from disk
- await more work

(What do we do with the RemoteSnapshot? I think probably add it to our cache?)

vu3rdd:
Fixed #193

### DIFF
--- a/src/magic_folder/magic_folder.py
+++ b/src/magic_folder/magic_folder.py
@@ -2223,12 +2223,6 @@ class UploaderService(service.Service):
     A service that periodically polls the database for local snapshots
     and commit them into the grid.
     """
-
-    # - expose a function that takes a file and gives back a remote snapshot URI.
-    # - The service periodically polls the database for LocalSnapshots and commits
-    #   them into the grid.
-    # - at startup, always checks the database for local snapshots and commits them.
-
     _snapshot_store = attr.ib()
     local_author = attr.ib()
     tahoe_client = attr.ib()
--- a/src/magic_folder/magic_folder.py
+++ b/src/magic_folder/magic_folder.py
@@ -2239,6 +2239,7 @@ class UploaderService(service.Service):
 
         service.Service.startService(self)
 
+        # do a looping call that polls the db for LocalSnapshots.
         self._processing_loop = task.LoopingCall(
             self._upload_local_snapshots,
             self.tahoe_client,
@@ -2246,18 +2247,11 @@ class UploaderService(service.Service):
         self._processing_loop.clock = self._clock
         self._processing = self._processing_loop.start(1, now=True)
 
-        # do a looping call that polls the db for LocalSnapshots.
-        # self._service_d = self._upload_local_snapshots(self.tahoe_client)
-
     def stopService(self):
         """
         Stop the uploader service.
         """
-        # d = self._service_d
-        #self._service_d.cancel()
         service.Service.stopService(self)
-        #self._service_d = None
-        #return d
         d = self._processing
         self._processing_loop.stop()
         self._processing = None
--- a/src/magic_folder/magic_folder.py
+++ b/src/magic_folder/magic_folder.py
@@ -2222,9 +2222,9 @@ class LocalSnapshotService(service.Service):
 
 @attr.s
 @implementer(service.IService)
-class UploadService(service.Service):
+class UploaderService(service.Service):
     """
-    A service that periodically polls the disc database for local snapshots
+    A service that periodically polls the database for local snapshots
     and commit them into the grid.
     """
 
@@ -2234,43 +2234,12 @@ class UploadService(service.Service):
     # - at startup, always checks the database for local snapshots and commits them.
 
     magic_path = attr.ib()
+    _snapshot_creator = attr.ib()
     db = attr.ib()
-    stash_dir = attr.ib(validator=attr.validators.instance_of(FilePath))
-    folder_name = attr.ib(validator=attr.validators.instance_of(unicode))
-    config = attr.ib() # Tahoe Config instance (created via `allmydata.client.read_config`)
-    _queue = attr.ib(default=attr.Factory(DeferredQueue))
 
     def startService(self):
 
-        # starts LocalSnapshotService
-        # make itself the parent of LocalSnapshotService.
-
-        # XXX: This would potentially change once the config PR is merged in.
-        try:
-            self.local_author = create_local_author_from_config(self.config, self.folder_name)
-        except RuntimeError:
-            # XXX
-            pass
-        except EnvironmentError:
-            # XXX: file does not exist
-            pass
-        except MissingConfigEntry:
-            # configuration is missing
-            pass
-
-        snapshot_creator = LocalSnapshotCreator(
-            db=self.db,
-            author=self.local_author,
-            stash_dir=self.stash_dir,
-        )
-
-        snapshot_service = LocalSnapshotService(
-            magic_path=self._magic_path,
-            snapshot_creator=snapshot_creator,
-        )
-
-        snapshot_service.setServiceParent(self)
-
+        service.Service.startService(self)
         # do a looping call that polls the db for LocalSnapshots.
         self._service_d = self._upload_local_snapshots()
 
@@ -2295,44 +2264,38 @@ class UploadService(service.Service):
 
         while True:
             # get the mangled paths for the LocalSnapshot objects in the db
-            localsnapshot_relpaths = self.db.get_all_localsnapshot_paths()
+            localsnapshot_relpaths = self._snapshot_creator.get_all_item_paths()
 
             # XXX: processing this table should be atomic. i.e. While the upload is
             # in progress, a new snapshot can be created on a file we already uploaded
             # but not removed from the db and if it gets removed from the table later,
-            # the new snapshot gets lost. i.e.
-            # in the db context:
-            #  - get_local_snapshot(name, author) for each name in the table
-            #  - write_snapshot_to_tahoe(snapshot, author_key, tahoe_client)
-            #  - if that succeeds, delete the local snapshot from db.
-            #  - store the resulting remote snapshot (capability) in a remote snapshots table.
-            #    This table will always point to the latest remote snapshot cap.
+            # the new snapshot gets lost. Perhaps this can be solved by storing each
+            # LocalSnapshot in its own row than storing everything in a blob?
+            # https://github.com/LeastAuthority/magic-folder/issues/197
             for relpath in localsnapshot_relpaths:
-                # deserialize into LocalSnapshot and copy them into the queue
-                localsnapshot = self.db.get_local_snapshot(relpath, self.local_author)
-                self._queue.put(localsnapshot)
+                # deserialize into LocalSnapshot
+                snapshot = self._snapshot_creator.get_local_snapshot(relpath)
 
-            # now upload each item in the queue
-            snapshot = yield self._queue.get(localsnapshot)
-            try:
-                remote_snapshot = yield write_snapshot_to_tahoe(
-                    snapshot,
-                    self.local_author,
-                    tahoe_client,
-                )
-            except NoServersError:
-                # Unable to reach Tahoe storage nodes because of
-                # network errors or because the tahoe storage nodes
-                # are offline.
-                self._queue.put(localsnapshot)
-                continue
-            except Exception:
-                # all other exceptions, pass on upstream
-                raise
+                # now upload each item in the queue
+                try:
+                    remote_snapshot = yield write_snapshot_to_tahoe(
+                        snapshot,
+                        self.local_author,
+                        tahoe_client,
+                    )
+                except NoServersError:
+                    # Unable to reach Tahoe storage nodes because of
+                    # network errors or because the tahoe storage nodes
+                    # are offline. Retry?
+                    # XXX: Perhaps implement exponential backoff for retry?
+                    continue
+                except Exception:
+                    # all other exceptions, pass on upstream
+                    raise
 
-            # At this point, remote snapshot creation successful for
-            # the given relpath. Remove the LocalSnapshot from the db.
-            self.db.delete_local_snapshot(relpath)
+                # At this point, remote snapshot creation successful for
+                # the given relpath. Remove the LocalSnapshot from the db.
+                self._snapshot_creator.remove_localsnapshot(relpath)
 
-            # store the remote snapshot capability in the db.
-            self.db.store_remote_snapshot(relpath, remote_snapshot)
+                # store the remote snapshot capability in the db.
+                self._snapshot_creator.store_remote_snapshot(relpath, remote_snapshot)
--- a/src/magic_folder/magic_folder.py
+++ b/src/magic_folder/magic_folder.py
@@ -2229,7 +2229,7 @@ class UploaderService(service.Service):
     #   them into the grid.
     # - at startup, always checks the database for local snapshots and commits them.
 
-    _snapshot_creator = attr.ib()
+    _snapshot_store = attr.ib()
     local_author = attr.ib()
     tahoe_client = attr.ib()
 
@@ -2260,7 +2260,7 @@ class UploaderService(service.Service):
 
         while True:
             # get the mangled paths for the LocalSnapshot objects in the db
-            localsnapshot_relpaths = self._snapshot_creator.get_all_item_paths()
+            localsnapshot_relpaths = self._snapshot_store.get_all_item_paths()
 
             # XXX: processing this table should be atomic. i.e. While the upload is
             # in progress, a new snapshot can be created on a file we already uploaded
@@ -2270,7 +2270,7 @@ class UploaderService(service.Service):
             # https://github.com/LeastAuthority/magic-folder/issues/197
             for relpath in localsnapshot_relpaths:
                 # deserialize into LocalSnapshot
-                snapshot = self._snapshot_creator.get_local_snapshot(relpath, self.local_author)
+                snapshot = self._snapshot_store.get_local_snapshot(relpath, self.local_author)
 
                 # now upload each item in the queue
                 try:
@@ -2291,7 +2291,7 @@ class UploaderService(service.Service):
 
                 # At this point, remote snapshot creation successful for
                 # the given relpath. Remove the LocalSnapshot from the db.
-                self._snapshot_creator.remove_localsnapshot(relpath)
+                self._snapshot_store.remove_localsnapshot(relpath)
 
                 # store the remote snapshot capability in the db.
-                self._snapshot_creator.store_remote_snapshot(relpath, remote_snapshot)
+                self._snapshot_store.store_remote_snapshot(relpath, remote_snapshot)
--- a/src/magic_folder/magic_folder.py
+++ b/src/magic_folder/magic_folder.py
@@ -2226,22 +2226,37 @@ class UploaderService(service.Service):
     _snapshot_store = attr.ib()
     local_author = attr.ib()
     tahoe_client = attr.ib()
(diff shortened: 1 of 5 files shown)


src/magic_folder/magic_folder.py
```diff
@@ -918,3 +962,92 @@ def add_file(self, path):
...
+
+        # get the mangled paths for the LocalSnapshot objects in the db
+        localsnapshot_names = self._snapshot_store.get_all_item_paths()
+
+        # XXX: processing this table should be atomic. i.e. While the upload is
+        # in progress, a new snapshot can be created on a file we already uploaded
+        # but not removed from the db and if it gets removed from the table later,
+        # the new snapshot gets lost. Perhaps this can be solved by storing each
+        # LocalSnapshot in its own row than storing everything in a blob?
+        # https://github.com/LeastAuthority/magic-folder/issues/197
```
exarkun, line 1020: Yea, I think this is right.  Instead of loading a whole tree of snapshots from each row in the db, it should load a single snapshot from a single row.  The schema will probably change a bit (path no longer primary key) but those details can be hidden inside the database layer.  This code can ask for an iterator of snapshots or it can ask for the newest snapshot with no other local parents or something similar, send one off for processing, repeat.

src/magic_folder/magic_folder.py
```diff
@@ -918,3 +962,92 @@ def add_file(self, path):
...
+                    # Unable to reach Tahoe storage nodes because of
+                    # network errors or because the tahoe storage nodes
+                    # are offline. Retry?
+                    # XXX: Perhaps implement exponential backoff for retry?
+                    NO_NETWORK.log()
+                    continue
+                except Exception:
+                    # all other exceptions, pass on upstream
+                    proc = Failure()
+                    write_failure(proc)
```
exarkun, line 1052: If you let these exceptions propagate through the `with action` block then Eliot will automatically mark the action a failure and record some details about the failure.  That might be sufficient for our purposes.  You still have to catch the exception outside the `with action` block to stop it propagating too far, but you may not have to _do_ anything with it there (certainly you don't have to log it again).

src/magic_folder/test/test_upload.py
```diff
@@ -0,0 +1,164 @@
...
+        # this should be picked up by the Uploader Service and should
+        # result in a snapshot cap.
+        d.addCallback(lambda _unused:
+                      self.snapshot_store.get_remote_snapshot_cap(name))
+
+        # test whether we got a capability
+        self.assertThat(
+            d,
+            succeeded(
+                AfterPreprocessing(is_uri, Always()),
```
exarkun, line 160: This should probably be something like `MatchesPredicate(is_uri, "%s is not a uri")`.  `Always` matches everything, not just True (and there is nothing like a `True` matcher that I know of that would do what we want here).

src/magic_folder/test/test_upload.py
```diff
@@ -0,0 +1,164 @@
...
+            ),
+        )
+
+        self.uploader_service.stopService()
```
exarkun, line 164: If the assertion fails this code never gets run.  In this case, it's nice if the stopService cleanups get a chance to happen anyway so a nice pattern is:

```
self.uploader_service.startService()
self.addCleanup(self.uploader_service.stopService)

...
```

"Cleanups" added this way always run even if the test fails.

src/magic_folder/test/test_upload.py
```diff
@@ -0,0 +1,164 @@
...
+    create_fake_tahoe_root,
+    create_tahoe_treq_client,
+)
+from magic_folder.tahoe_client import (
+    create_tahoe_client,
+)
+from allmydata.uri import is_uri
+
+@attr.s
+class MemorySnapshotStore(object):
```
exarkun, line 51: It would be convenient if we didn't need this.  I usually consider a ":memory:"-connected SQLite3 database to be Good Enough™ for use in tests.  It doesn't do real I/O so it's not slow like really using the database would be.  It's pretty opaque which can be annoying but with a simple CRUD-ish interface like this one I don't think that's going to be much of an issue.  Test doubles are great but avoiding them is even better. :)

src/magic_folder/magic_folder.py
vu3rdd: Would doing the following be equivalent then? It would log the exception and the failed action, which seem to be what we want here. 
```
...
except Exception:
    write_traceback()
```

src/magic_folder/test/test_upload.py
vu3rdd: I keep forgetting this..

vu3rdd: Looks like, with the new config changes, it is going to be hard to do this (without modifying the current config code..) because the filename is hard coded into the code and is not parameterized anymore. (Overall, that seem like a good decision though)

vu3rdd:
I pushed some changes. I have these two main concerns with the current state of the code:

1. The `UploaderService` is now split into a service class and another `RemoteSnapshotCreator` class. However, the Service class has almost nothing other than start/stop the service. Tests for the `RemoteSnapshotCreator` felt almost like a repeat of `UploaderService` tests, so I left them out. May be there is some better abstraction that can be made..

2. There is a race condition in the tests. Since we are polling for local snapshots in the db, we would need to wait for at least the polling interval seconds before looking for remote snapshot caps in the db. The test currently works only because we are starting it after creating the local snapshots in the db, so they immediately get picked up by the service on startup. I wonder how to take care of that.

src/magic_folder/magic_folder.py
```diff
@@ -857,3 +874,92 @@ def add_file(self, path):
...
+                    # store the remote snapshot capability in the db.
+                    yield self._state_db.store_remotesnapshot(name, remote_snapshot)
+
+                except NoServersError:
+                    # Unable to reach Tahoe storage nodes because of
+                    # network errors or because the tahoe storage nodes
+                    # are offline. Retry?
+                    # XXX: Perhaps implement exponential backoff for retry?
+                    NO_NETWORK.log()
+                    continue
```
vu3rdd, line 928: Is there a way to write a test to exercise this path?

meejah:
I haven't looked at the code closely (yet) but the usual way to deal with "I want time to advance quickly" is to pass a `twisted.internet.task.Clock` in place of the reactor (and then call `.advance()` on the clock when you want time to move ahead). So that means you could add something to the database, call `the_clock.advance(3)` and be assured that the polling timeout has fire (and so the upload should have happened).

vu3rdd:
> I haven't looked at the code closely (yet) but the usual way to deal with "I want time to advance quickly" is to pass a `twisted.internet.task.Clock` in place of the reactor (and then call `.advance()` on the clock when you want time to move ahead). So that means you could add something to the database, call `the_clock.advance(3)` and be assured that the polling timeout has fire (and so the upload should have happened).

Thanks so much! Updated the code to use `task.Clock()`.

src/magic_folder/magic_folder.py
```diff
@@ -35,6 +36,9 @@
     RELPATH,
     validateSetMembership,
 )
+from allmydata.interfaces import (
+    NoServersError,
```
meejah, line 40: We shouldn't introduce new `allmydata` imports (it's not clear to me this will ever get thrown; see below).

src/magic_folder/magic_folder.py
```diff
@@ -857,3 +874,97 @@ def add_file(self, path):
...
+                    yield self._state_db.store_remotesnapshot(name, remote_snapshot)
+
+                    # Remove the local snapshot content from the stash area.
+                    snapshot.content_path.remove()
+
+                    # Remove the LocalSnapshot from the db.
+                    yield self._state_db.delete_localsnapshot(name)
+
+
+                except NoServersError:
```
meejah, line 927: How would this ever get thrown? We will use `tahoe_client.create_immutable_directory()` under the hood which can only throw HTTP errors, AFAICT.

src/magic_folder/magic_folder.py
meejah: We would need to augment the test code to arrange to throw this error.

So in `test_snapshot` it uses `create_tahoe_client` and `create_tahoe_treq_client` and `create_fake_tahoe_root` to set up the "Tahoe" API we're using in those tests ... so the "lowest level" thing to do would be to augment the fake Tahoe root so that it e.g. always throws an error. We could for example have different kinds of "fake tahoe roots" that do different things or a (more complex) single thing that can be told how to behave. It's probably better to keep things simple with multiple kinds of roots (one kind of root can always wrap another)

To test e.g. an exponential backoff strategy here, the fake root could throw an error on the first request, but not on subsequent requests. So this could be a "wrapping" root that has a `_FakeTahoeRoot` but only calls through to it on the *second* PUT or whatever.

src/magic_folder/magic_folder.py
```diff
@@ -857,3 +874,97 @@ def add_file(self, path):
...
+@attr.s
+@implementer(service.IService)
+class UploaderService(service.Service):
+    """
+    A service that periodically polls the database for local snapshots
+    and commit them into the grid.
+    """
+    _tahoe_client = attr.ib()
+    _clock = attr.ib()
+    _polling_interval = attr.ib() # in GlobalConfigdatabase
```
meejah, line 946: We might want to discuss how to handle configuration of these things "in general". That is, either we pass them the individual pieces of config they need upon creation (like here) or we pass them a config instance .. I don't see any code that actually creates one of these but one strategy would be a method like `UploaderService.from_config()` that takes a config but pulls out what the `UploaderService` actually needs.

I think I like that way, but it has follow-on implications for how we handle "the config has changed" -- with such a strategy, each service object would have to be re-created. Whereas if each thing was given its config, it would have to be told "your config (maybe) changed". I feel like the former strategy is straightforward and easy but I haven't thought too hard about it (and there might be edge-cases that aren't great).

src/magic_folder/test/test_local_snapshot.py
```diff
@@ -59,14 +59,14 @@ class MemorySnapshotCreator(object):
     """
     A way to test LocalSnapshotService with an in-memory database.
 
-    :ivar [FilePath] processed: All of the paths passed to ``process_item``,
+    :ivar [FilePath] processed: All of the paths passed to ``store_local_snapshot``,
```
meejah, line 62: :+1:

src/magic_folder/test/test_upload.py
```diff
@@ -0,0 +1,147 @@
...
+        # test whether we got a capability
+        self.assertThat(
+            d,
+            succeeded(
+                MatchesPredicate(is_uri,
+                                 "%s is not a Tahoe-LAFS URI"),
+            ),
+        )
+
+        self.addCleanup(self.uploader_service.stopService)
```
meejah, line 147: In general, better to add cleanup lines immediately after the thing being cleaned-up (e.g. right after the `startService` call) to ensure it'll run even if some other stuff in the test goes wrong (before here).

src/magic_folder/test/test_upload.py
```diff
@@ -0,0 +1,147 @@
...
+        # advance the clock manually, which should result in the
+        # polling of the db for uncommitted LocalSnapshots in the db
+        # and then check for remote snapshots
+        self.uploader_service._clock.advance(self.polling_interval)
+
+        d.addCallback(lambda _unused:
+                      self.state_db.get_remotesnapshot(name))
+
+        # test whether we got a capability
+        self.assertThat(
```
meejah, line 139: This is certainly a good assert.

I think there are other things we should assert: that the local-snapshot database no longer has an entry for this thing (and maybe that the remote-snapshot database DOES have an entry for it now). We could even download the capability and ensure the contents match etc (but I feel like that behavior is already adequately covered by the snapshot tests).

src/magic_folder/magic_folder.py
vu3rdd: Oh, right. I was thinking that we would get a bubbled up "connection failed" socket error. However that is not the case since that error would happen in the tahoe node that is talking to the storage node and the client talking to the tahoe client node running on the local host would probably reply with a `504` (Gateway timed out) perhaps?

src/magic_folder/test/test_upload.py
vu3rdd: Thanks. Yes, a few of those are already tested by snapshot tests. The above assert is checking for an entry in the remote snapshot table and checking whether it is a valid url. I also added a test to check whether the local snapshot entry is non existent.

src/magic_folder/magic_folder.py
```diff
@@ -838,3 +852,141 @@ def add_file(self, path):
...
+            except:
+                # Unable to reach Tahoe storage nodes because of network
+                # errors or because the tahoe storage nodes are
+                # offline. Retry?
+                pass
+
+    @eliotutil.inline_callbacks
+    def _upload_one_snapshot(self, name):
+        """
+        Upload all of the snapshots for a particular path.
```
meejah, line 909: This docstring contradicts the method name; is it all, or just one? (Looks like the docstring should change)

src/magic_folder/magic_folder.py
```diff
@@ -838,3 +852,141 @@ def add_file(self, path):
...
+        # update the entry in the DMD
+        yield self._tahoe_client.add_entry_to_mutable_directory(
+            self._upload_dircap,
+            magicpath.path2magic(name),
+            remote_snapshot.capability.encode('utf-8'),
+            replace=True,
+        )
+
+        # Remove the local snapshot content from the stash area.
+        snapshot.content_path.remove()
```
meejah, line 934: I'm trying to think of failure cases, and I think we might want to mostly-ignore errors here yet still remove the database entry? (Thinking: if there's some error removing the content, could it be partially-removed at that point? If so, then the database is kind of inconsistent: there's a database entry, but corrupted or missing "content"). Caveat: **is** it possible that the `remove` fails such that the content is gone/missing but still produces an error?

src/magic_folder/test/strategies.py
```diff
@@ -317,3 +344,35 @@ def unique_value_dictionaries(keys, values, min_size=None, max_size=None):
...
+    """
+    Build ``RemoteSnapshot`` instances.
+    """
+    return builds(
+        RemoteSnapshot,
+        name=names,
+        author=authors,
+        metadata=dictionaries(text(), text()),
+        capability=tahoe_lafs_dir_capabilities(),
+        parents_raw=lists(tahoe_lafs_dir_capabilities()),
```
meejah, line 359: Not sure if this is that big a deal, but `tahoe_lafs_dir_capabilities()` will look like mutable directories -- but all snapshots will be immutable directories so the parents list should all contain immutable-caps.

The `capability` should similarly be an immutable-directory.

src/magic_folder/test/strategies.py
```diff
@@ -317,3 +344,35 @@ def unique_value_dictionaries(keys, values, min_size=None, max_size=None):
...
+    Currently this builds snapshots with no local parents.
+    """
+    return builds(
+        LocalSnapshot,
+        name=relative_paths(),
+        author=local_authors(),
+        metadata=dictionaries(text(), text()),
+        content_path=absolute_paths().map(FilePath),
+        parents_local=just([]),
+        parents_remote=lists(tahoe_lafs_dir_capabilities()),
```
meejah, line 377: Same as above comment (immutable-caps).

src/magic_folder/testing/web.py
```diff
@@ -266,16 +286,23 @@ def _add_entry_to_dir(self, request, dircap, segments):
...
             "ro_uri": content_cap,
-            "verify_cap": content.get_verify_cap().to_string(),
+            "verify_uri": content.get_verify_cap().to_string(),
+            "format": _get_node_format(content),
         }
         if content_cap != content.get_readonly().to_string():
             metadata["rw_uri"] = content_cap
 
+        if kind == "filenode" and content.get_size() is not None:
+            metadata["size"] = content.get_size()
```
meejah, line 304: Not sure if it helps or not, but the size of immutable things will be encoded in the capability itself.

src/magic_folder/magic_folder.py
```diff
@@ -723,8 +738,8 @@ def process_item(self, path):
             # If so, we use that as the parent.
             mangled_name = magicpath.mangle_path(path)
             try:
-                parent_snapshot = self.db.get_local_snapshot(mangled_name, self.author)
-            except SnapshotNotFound:
+                parent_snapshot = self._db.get_local_snapshot(mangled_name)
+            except KeyError:
                 parents = []
```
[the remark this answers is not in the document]
exarkun: I'm not really sure.  Maybe?  Possibly if you have no local snapshot then the parent would be whatever remote HEAD snapshot you can find.  Maybe the downloader needs to record which snapshot it downloaded and if you can't find a local snapshot then that snapshot is the parent.

src/magic_folder/magic_folder.py
```diff
@@ -838,3 +852,141 @@ def add_file(self, path):
...
+        # but not removed from the db and if it gets removed from the table later,
+        # the new snapshot gets lost. Perhaps this can be solved by storing each
+        # LocalSnapshot in its own row than storing everything in a blob?
+        # https://github.com/LeastAuthority/magic-folder/issues/197
+        for name in localsnapshot_names:
+            action = UPLOADER_SERVICE_UPLOAD_LOCAL_SNAPSHOTS(relpath=name)
+            try:
+                with action:
+                    yield self._upload_one_snapshot(name)
+            except:
```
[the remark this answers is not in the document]
exarkun: Yes on `except Exception:` I guess.  I'm never sure what the practical difference is.  Is it _good_ to let `KeyboardInterrupt` pass through this layer?  Dunno ...

There is logging already, at least.  The exception has to propagate through the `action` context manager and it will log the failure nicely (marking the action as failed and including a traceback).  So yea, extra magic. :)

src/magic_folder/magic_folder.py
exarkun: It does contradict it, yea ... But it's the docstring that's right, not the method name. :/  The way this works now is pretty bad.  Pulling a whole snapshot history out of the database and trying to operate on it all at once is problematic.  This will get fixed by the "schema normalization" issue eventually.  I'll rename the method for now, I guess.

src/magic_folder/magic_folder.py
exarkun: > I'm trying to think of failure cases, and I think we might want to mostly-ignore errors here yet still remove the database entry?

Something like this, probably.  There are a bunch of atomicity issues with this function as written now but they're not easily fixable with the current database schema.  I'm tempted to leave this for now and make another pass later on.  It will work in the happy-case, at least ...

> Caveat: is it possible that the remove fails such that the content is gone/missing but still produces an error?

I think the rule of thumb for answering "Is it possible for a filesystem to have confounding behavior X?" is "yes".

src/magic_folder/test/strategies.py
exarkun: Nice catch.  Yea, it would be better for these values to be realistic.

src/magic_folder/testing/web.py
exarkun: Yea, I noticed that too.  This change is only meant to make the response look more like the Tahoe-LAFS response.  It looks like Tahoe-LAFS doesn't include the size for immutable directories, only immutable files.

PR #228, opened 2020-08-03
Delete Uploader and Downloader

### LINKED ISSUE #227: Uploader and Downloader are no longer useful
These classes implement the old logic for uploading and downloading snapshots.  The on-grid schema for snapshot storage has changed such that these classes no longer work as-is.  Additionally, the complexity of their implementation and test suite and the degree to which they are intertwined with each other and other code means that adapting them to the new schema will be more work than deleting them and starting fresh.  There isn't even a benefit in keeping them until their replacements are ready as we are about to make magic-folder non-functional anyway by removing support for running against old-style configurations that live inside Tahoe-LAFS.

So, delete this code and clear the path for the new stuff.

exarkun:
Fixes #227


meejah:
since the patch coverage is 100% i'm assuming the other codecov thing is just wrong yet again

PR #230, opened 2020-08-03
229.delete magicfolderdb

### LINKED ISSUE #229: magicfolderdb is no longer useful
The magicfolderdb module implements the database for old-style magicfolders where there is no history, where the on-grid schema for snapshots doesn't completely model the relationship between snapshots, and for which support is being dropped entirely in favor of the newer version-control-like snapshot schema.

Old folders will be converted to new folders using the `magic-folder migrate` command and this is a required step before the next release of Magic-Folder will be able to interact with the folder.

exarkun:
Fixes #229 

Stacked on top of #227 / #228 so review that first.


src/magic_folder/config.py
```diff
@@ -204,40 +225,102 @@ class MagicFolderConfig(object):
...
+        :param author: an instance of LocalAuthor
+
+        :returns: An instance of LocalSnapshot for the given magicpath.
+        """
+        cursor.execute("SELECT snapshot_blob FROM local_snapshots"
+                       " WHERE path=?",
+                       (name,))
+        row = cursor.fetchone()
+        if not row:
+            return None
```
meejah, line 278: I guess more of a style/philosophy thing but maybe an exception here is "better"? (e.g. `KeyError` or ..?)

src/magic_folder/config.py
```diff
@@ -204,40 +225,102 @@ class MagicFolderConfig(object):
...
+                [snapshot_blob] = excluded.[snapshot_blob]
+
+            """,
+            (snapshot.name, snapshot.to_json()),
+        )
+
+    @with_cursor
+    def _get_all_relpaths_from(self, cursor, statement):
+        """
+        :param unicode tablename: table name to fetch the relpaths from.
```
meejah, line 308: The docs and params don't seem to match the function name or what it does .. seems more like `_set_of_results_from_sql_statement` or something..? Since it seems to only be used in the method below I would just combine them (and could switch to a helper when there's more than one user).

src/magic_folder/config.py
exarkun: Yea, I think an exception is better.  It's harder to accidentally ignore.

Issue #231, opened 2020-08-03
There are two ways to get base URL


sajith:
After #223 has been merged, Magic Folder has two ways to get the base URL for HTTP API:

 * By reading the file `/path/to/taho-node/magic-folder.url`
 * By calling `endpoint_description_to_http_api_root(api_endpoint)`

One of them is possibly redundant, and therefore a source of confusion.

(This is similar to #224, maybe?)

meejah:
Does this confusion still exist now that #216 is merged? Anything at all from `tahoe-node` directory is only for Tahoe to use (and so if it has a `magic-folder.url` that's a "legacy" magic folder of `tahoe magic-folder *` type).

I will split the part of #207 that adds an explicit "client endpoint" out so then there will only be one way: ask the config for the client-endpoint.

meejah:
The client API endpoint stuff was merged in https://github.com/LeastAuthority/magic-folder/pull/252 so I think this confusion should no longer exist. (If it does, please re-open).

Issue #224, opened 2020-08-03
Magic folder HTTP API key confusion


sajith:
I did a `magic-folder migrate`, and then restarted things:

```
$ magic-folder migrate -n /path/to/nodes/alice/ --listen-endpoint tcp:6001 --author-name alice
$ tahoe run /path/to/nodes/alice/
$ magic-folder -n /path/to/nodes/alice/ run --web-port tcp:6001
```

When making API calls, the token stored in  `~/.config/magic-folder/api_token` does not work (using [httpie](https://httpie.org)):

```
$ http :6001/v1/magic-folder authorization:"Bearer $(cat ~/.config/magic-folder/api_token)"
HTTP/1.1 401 Unauthorized
Content-Length: 0
Date: Mon, 03 Aug 2020 02:52:26 GMT
Server: TwistedWeb/20.3.0
```

But the one stored in `/path/to/nodes/alice/private/api_auth_token` works:

```
$ http :6001/v1/magic-folder authorization:"Bearer $(cat /path/to/nodes/alice/private/api_auth_token)"
HTTP/1.1 200 OK
Content-Length: 156
Content-Type: application/json
Date: Mon, 03 Aug 2020 03:00:13 GMT
Server: TwistedWeb/20.3.0

{
    "folders": [
        {
            "local-path": "/home/sajith/Documents/Shared",
            "name": "shared"
        },
        {
            "local-path": "/home/sajith/Documents/test",
            "name": "test"
        }
    ]
}
```
Looks like API key is read in `MagicFolderService` here:

https://github.com/LeastAuthority/magic-folder/blob/3bf6cbc70c6dac77952469ff32408d3856c947fd/src/magic_folder/cli.py#L691-L695

Which does this:

https://github.com/LeastAuthority/magic-folder/blob/3bf6cbc70c6dac77952469ff32408d3856c947fd/src/magic_folder/cli.py#L711-L712

But that is the wrong key to use after migration, right?  Or maybe there shouldn't be two different keys?

(#207 depends on this.)

meejah:
There is an `api_client_endpoint` and `api_token` file available for clients to read now.

Issue #207, opened 2020-07-17
`magic-folder list` should use the HTTP API instead of reaching into internal Magic-Folder state


exarkun:
#205 introduces an HTTP API which will give back a list of all known Magic Folders and their filesystem locations.  This should be all that `magic-folder list` needs to do its job.  It should use this HTTP API instead of what it does now.

exarkun:
I guess this is somewhat blocked on #200 because `magic-folder list` will need a way to *find* the HTTP API.  I could add an explicit command-line argument but we probably just want it to find the magic-folder configuration and look there, right?

If there's anything accepted on the command line, it would be an alternate location for magic-folder configuration to load, I guess, not options to override each individual piece of configuration.

PR #233, opened 2020-08-04
`magic-folder list` command handler using http api

sajith:
This PR implements #207.  This is the "I don't really know what I am doing" version. :-)

The basic thing sort of works, as long as there is an API token stored in Tahoe node (in `/path/to/nodes/alice/private/api_auth_token`).  Because of #224, the token stored in `~/.config/magic-folder/api_token` does not work.  Because of #231, I implemented two ways of getting the base URL for the API.  One of them will go. 

These two tests fail, presumably because there's not enough setup done to generate API keys.

 * `magic_folder.test.cli.test_magic_folder.ListMagicFolder.test_list_none_json`
* `magic_folder.test.cli.test_magic_folder.ListMagicFolder.test_list_some_json`


meejah:
see https://github.com/LeastAuthority/magic-folder/pull/278

Issue #238, opened 2020-08-07
Use pip cache in GitHub Actions


sajith:
[Cache](https://github.com/actions/cache/tree/master) will likely speed things up.

exarkun:
Closing ticket with no motivation.

PR #239, opened 2020-08-07
Use pip cache in GitHub Actions

sajith:
This is related to #238.

- [Example CI run without cache](https://github.com/LeastAuthority/magic-folder/runs/955252273?check_suite_focus=true): 21s to install dependencies.
- [Example CI run with cache](https://github.com/LeastAuthority/magic-folder/runs/958673132?check_suite_focus=true): 9s to install dependencies.

### DIFF
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -27,7 +27,9 @@ jobs:
     - name: Use pip cache
       uses: actions/cache@v2
       with:
-        path: ~/Library/Caches/pip
+        path:
+          - ~/Library/Caches/pip
+          - ./wheelhouse
         key: ${{ runner.os }}-pip-${{ hashFiles('setup.py') }}
         restore-keys: |
             ${{ runner.os }}-pip-
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -23,6 +23,7 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
+    # See https://github.com/actions/cache/blob/master/examples.md#python---pip
     - name: Use pip cache
       uses: actions/cache@v2
       with:
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -41,10 +41,10 @@ jobs:
       run: pip -v download . --dest ./wheelhouse
 
     - name: Test with tox
-      env:
+      env: |
         MAGIC_FOLDER_HYPOTHESIS_PROFILE: "magic-folder-ci"
-      run: |
-        PIP_FIND_LINKS=./wheelhouse tox -e py27-coverage
+        PIP_FIND_LINKS: ./wheelhouse
+      run: tox -e py27-coverage
 
     - uses: codecov/codecov-action@v1
       with:
@@ -87,8 +87,8 @@ jobs:
       run: pip -v download . --dest ./wheelhouse
 
     - name: Test with tox
-      run: |
-        PIP_FIND_LINKS=./wheelhouse tox -vvv -e integration
+      env: PIP_FIND_LINKS:./wheelhouse
+      run: tox -vvv -e integration
 
     - uses: codecov/codecov-action@v1
       with:
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -37,6 +37,9 @@ jobs:
         python -m pip install --upgrade pip
         pip install wheel tox
 
+    - name: Populate wheelhouse
+      run: pip -v download . --dest ./wheelhouse
+
     - name: Test with tox
       env:
         MAGIC_FOLDER_HYPOTHESIS_PROFILE: "magic-folder-ci"
@@ -79,6 +82,10 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install wheel tox
+
+    - name: Populate wheelhouse
+      run: pip -v download . --dest ./wheelhouse
+
     - name: Test with tox
       run: |
         PIP_FIND_LINKS=./wheelhouse tox -vvv -e integration
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -34,8 +34,10 @@ jobs:
         restore-keys: |
             ${{ runner.os }}-pip-
 
+    # Packages with pinned versions are from tox.ini; we will have to
+    # update them if/when tox.ini changes.
     - name: Populate wheelhouse
-      run: pip -v download . --dest ./wheelhouse
+      run: pip wheel --wheel-dir ./wheelhouse $(pwd)[test] pip==19.1.1 setuptools==41.0.1 wheel==0.33.4  subunitreporter==19.3.2 tox codecov python-subunit junitxml
 
     - name: Install dependencies
       run: |
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -37,7 +37,7 @@ jobs:
     # Packages with pinned versions are from tox.ini; we will have to
     # update them if/when tox.ini changes.
     - name: Populate wheelhouse
-      run: pip wheel --wheel-dir ./wheelhouse $(pwd)[test] pip==19.1.1 setuptools==41.0.1 wheel==0.33.4  subunitreporter==19.3.2 tox codecov python-subunit junitxml
+      run: pip download --dest ./wheelhouse $(pwd)[test] pip==19.1.1 setuptools==41.0.1 wheel==0.33.4  subunitreporter==19.3.2 tox codecov python-subunit junitxml
 
     - name: Install dependencies
       run: |
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -44,6 +44,7 @@ jobs:
       env:
         MAGIC_FOLDER_HYPOTHESIS_PROFILE: "magic-folder-ci"
         PIP_FIND_LINKS: ./wheelhouse
+        PIP_NO_INDEX: true
       run: tox -e py27-coverage
 
     - uses: codecov/codecov-action@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/macos.yaml
```diff
@@ -17,14 +17,26 @@ jobs:
...
     - run: "git fetch origin +refs/tags/*:refs/tags/*"
+
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v1
       with:
         python-version: ${{ matrix.python-version }}
+
+    - name: Use pip cache
+      uses: actions/cache@v2
+      if: startsWith(runner.os, 'macOS')
```
exarkun, line 28: Since this workflow is macOS only, is this redundant?

.github/workflows/macos.yaml
```diff
@@ -17,14 +17,26 @@ jobs:
...
     # Get tags not fetched by the checkout action, needed for auto-versioning.
     - run: "git fetch origin +refs/tags/*:refs/tags/*"
+
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v1
       with:
         python-version: ${{ matrix.python-version }}
+
+    - name: Use pip cache
+      uses: actions/cache@v2
```
exarkun, line 27: A link to any docs for this action might be a nice addition here.

.github/workflows/macos.yaml
sajith: Good idea, thank you!  Updated in b4ae2d7.

exarkun:
One more comment.

In the "Install dependencies" step only the tox dependencies are installed, I think:

> wheel, filelock, py, six, scandir, pathlib2, contextlib2, zipp, typing, singledispatch, importlib-resources, distlib, appdirs, configparser, importlib-metadata, virtualenv, pluggy, toml, pyparsing, packaging, tox

Our dependencies would include things like tahoe-lafs and cryptography.  It's not clear that there's any issue here but I think a lot of installation gets hidden inside the "tox ..." invocation.  I would hope tox would use this cache but it's hard to be sure since the tox output doesn't say anything about where it installs dependencies from (at least not at the verbosity level we're using).  Also the GitHub Actions cache file seems to be private (I tried to download it and got a 401 response) so it's hard to see what's actually in it.

The tox job in the two linked builds went from 5m50s to 5m35s which could be a nice bit of speedup but could also just be random variation...

Not sure there's anything actionable here.  Maybe crank up tox verbosity, if that's possible, so the logs demonstrate it's also using the cache?

.github/workflows/macos.yaml
sajith: Added a link to [relevant example](https://github.com/actions/cache/blob/master/examples.md#python---pip) in a38e077.

sajith:
> Not sure there's anything actionable here. Maybe crank up tox verbosity, if that's possible, so the logs demonstrate it's also using the cache?

At least when running locally with `tox -vvv -r coverage`, pip is using cached packages.  Also, from GitHub Actions' timestamps it doesn't look like pip invocation by tox isn't using much time: my reading is that [this step](https://github.com/LeastAuthority/magic-folder/pull/239/checks?check_run_id=958753593#step:7:20) completes in a second.

exarkun:
> > Not sure there's anything actionable here. Maybe crank up tox verbosity, if that's possible, so the logs demonstrate it's also using the cache?
> 
> At least when running locally with `tox -vvv -r coverage`, pip is using cached packages. Also, from GitHub Actions' timestamps it doesn't look like pip invocation by tox isn't using much time: my reading is that [this step](https://github.com/LeastAuthority/magic-folder/pull/239/checks?check_run_id=958753593#step:7:20) completes in a second.

Cool, and that's what I *hope* it would do.  Thanks for reminding me that github does actually timestamp each line and will show this if you ask, I forgot that was hidden behind the meatball shower.

sajith:
I noticed a couple of things when running `tox -vvv`.

- pip is using the cache when spawned from tox, as seen [here](https://github.com/LeastAuthority/magic-folder/pull/239/checks?check_run_id=958877488#step:7:322), so that is good.

- There are [these](https://github.com/LeastAuthority/magic-folder/pull/239/checks?check_run_id=958877488#step:7:320) messages, which suggests that `PIP_FIND_LINKS` is ignored:

  `WARNING: Url './wheelhouse' is ignored. It is either a non-existing path or lacks a specific scheme.`

- pip is hitting to pypy even when cache is enabled ([here](https://github.com/LeastAuthority/magic-folder/pull/239/checks?check_run_id=958877488#step:7:468)), when running tox.  I don't know what's up with that.  It happens when I run tox locally too.  If you search for `autobahn` or `cryptography`, for example, it seems to go through every published version of those packages.  Is it doing package resolution there?  Maybe we can save this time by pinning package versions?

exarkun:
> I noticed a couple of things when running `tox -vvv`.
> 
> * pip is using the cache when spawned from tox, as seen [here](https://github.com/LeastAuthority/magic-folder/pull/239/checks?check_run_id=958877488#step:7:322), so that is good.

Yea, "... using cached ..." looks good.

> * There are [these](https://github.com/LeastAuthority/magic-folder/pull/239/checks?check_run_id=958877488#step:7:320) messages, which suggests that `PIP_FIND_LINKS` is ignored:
>   `WARNING: Url './wheelhouse' is ignored. It is either a non-existing path or lacks a specific scheme.`


I think those mean that PIP_FIND_LINKS *is* being respected - but then the value is not something pip can deal with.

I don't see anything that would ever populate a wheelhouse ... so I guess that is the problem.  Possibly it doesn't make sense to have a wheelhouse for this job anyway.  The wheelhouse would be a way to pin specific versions and also avoid talking to pypi per-job.  But something has to populate it (it's just a directory with all of the wheels we need in it - and no *other* wheels).

> * pip is hitting to pypy even when cache is enabled ([here](https://github.com/LeastAuthority/magic-folder/pull/239/checks?check_run_id=958877488#step:7:468)), when running tox.  I don't know what's up with that.  It happens when I run tox locally too.  If you search for `autobahn` or `cryptography`, for example, it seems to go through every published version of those packages.  Is it doing package resolution there?  Maybe we can save this time by pinning package versions?

Yea, I guess that's the only way it is working right now since there is no wheelhouse.  My understanding is that even the pip cache is unusable if you don't talk to an index server (like pypi).

I can see two directions.  (1) delete the wheelhouse stuff or (2) set PIP_NO_INDEX=1 (I think that's the one) to make it stop talking to pypi and get the wheelhouse correctly populated so it can use that instead.

sajith:
I did a `pip download . --dest=./wheelhouse` to populate the wheelhouse, set `PIP_FIND_LINKS:./wheelhouse` and `PIP_NO_INDEX=1` and found out that it not work in our setup:  there are pinned versions `pip`, `wheel,` (current versions of these are installed in a prior step) `subunitreporter` (unsure if and how this is used), and `certifi` (doesn't look like magic folder uses `certifi` at all.)  We will need to get those versions as well, which will make CI setup hacky and brittle.

From my limited understanding and testing it looks like those pinned versions exist in `tox.ini` because of historical reasons...  today `tox` may work happily with just the dependencies listed in `setup.py`.

exarkun:
> I did a `pip download . --dest=./wheelhouse` to populate the wheelhouse, set `PIP_FIND_LINKS:./wheelhouse` and `PIP_NO_INDEX=1` and found out that it not work in our setup: there are pinned versions `pip`, `wheel,` (current versions of these are installed in a prior step) `subunitreporter` (unsure if and how this is used), and `certifi` (doesn't look like magic folder uses `certifi` at all.) We will need to get those versions as well, which will make CI setup hacky and brittle.

subunitreporter is used to run the test suite and get results in a parseable format (which, eg, circleci consumes to create a slightly more useful report than just raw test run text output - it would be cool if github actions could do something similar?  Ther's some kind of "annotations" feature but I don't know if this is helpful here.).

certifi isn't used directly.  Instead, getting an up-to-date version makes any subsequent `pip install` commands less likely to fail with some TLS error.

> 
> From my limited understanding and testing it looks like those pinned versions exist in `tox.ini` because of historical reasons... today `tox` may work happily with just the dependencies listed in `setup.py`.

If there are newer versions of any of these which actually _work_ we can update the pins but I don't think it's a good idea to let any of these packages float again.  They all have a proven track record of frequent regressions in new releases which then cause a lot of hassle for us to clean up while we're trying to work on other stuff.

The CircleCI configurations build Docker images that contain most of the dependencies in a wheelhouse.  It might be interesting to explore applying this to GitHub Actions as well (it will also have the effect of speeding up builds).

It would also be interesting if there were some kind of higher-level specification we could use to write our CI configurations so we don't have so much tedious boilerplate maintenance to do across the different services ... I know I've seen one attempt at such but I'm drawing a blank on its name right now.

sajith:
I tried a few things:

* Added wheelhouse directory to the list of things to be cached.  That appears to be working as intended.
* Following CircleCI setup, used `pip wheel` (in [this](https://github.com/LeastAuthority/magic-folder/pull/239/commits/6829f9485a0e02d5381144c48e465905a8cc7f29) commit) to populate wheelhouse directory, but it failed with messages like `error: invalid command 'bdist_wheel'` ([example](https://github.com/LeastAuthority/magic-folder/runs/968617738?check_suite_focus=true#step:6:379)).  It had worked for me in my local environment, and on here we've installed `pip` and `wheel` in a prior step, so I'm unsure why that failure happens there.  Went back to using `pip download` instead.
* Used `PIP_NO_INDEX=true` but things are still missing in the wheelhouse directory.  There're error messages like: `"ERROR: Could not find a version that satisfies the requirement bcrypt>=3.0.0; extra == "conch" (from Twisted[conch,tls]>=18.4.0->tahoe-lafs==1.14.0->magic-folder==0.0.0.post1037) (from versions: none)"` ([here](https://github.com/LeastAuthority/magic-folder/runs/968665408?check_suite_focus=true#step:8:101)).

Summary: wheelhouse directory is being cached, but not sufficiently populated to work without PyPI.  Running tests take more or less the same time as before (6-7m).  Now I'm unsure how to proceed here.

exarkun:
#238 has no motivation on it so I'm going to close it and this.

PR #240, opened 2020-08-07
Replace Ubuntu 16.04 with Ubuntu 20.04

### LINKED ISSUE #181: Add support for Ubuntu 20.04
Ubuntu 20.04 was released in April and is an LTS release supported until 2025. We should test against it on CI.

exarkun:
Fixes #181

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,10 +11,10 @@ workflows:
           requires:
             - "debian-9"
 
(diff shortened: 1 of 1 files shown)


exarkun:
Self-reviewing since this is CI configuration change only.  The NixOS CI job failure is unrelated and I filed a ticket (#245 ).  The rest looks good.

PR #243, opened 2020-08-10
remove some dead code from cli

### LINKED ISSUE #242: tahoe_client.py:{add_file, download_best_version} are obsolete
With the new design `tahoe_client.py:TahoeClient.add_file()` and `download_best_version()` apis seem obsolete.

vu3rdd:
Fixes #242. 

Delete `cli.py:TahoeClient:download_best_version()` and `cli.py:TahoeClient:add_file()` which has no tests and are not being used anywhere in the new design.

### DIFF
--- a/src/magic_folder/cli.py
+++ b/src/magic_folder/cli.py
@@ -825,24 +825,6 @@ class Node(object):
     def list(self):
         return self.tahoe_client.list_directory(self.uri)
 
-    def download_best_version(self, progress):
-        return self.tahoe_client.download_best_version(
-            self.uri, progress
-        )
-
-    def add_file(self, name, uploadable, metadata=None, overwrite=True, progress=None):
-        action = start_action(
-            action_type=u"magic-folder:cli:add_file",
-            name=name,
-        )
-        with action.context():
-            d = DeferredContext(
-                self.tahoe_client.add_file(
-                    self.uri, name, uploadable, metadata, overwrite, progress,
-                ),
-            )
-            return d.addActionFinish()
-
 @attr.s(frozen=True)
 class TahoeClient(object):
     node_uri = attr.ib()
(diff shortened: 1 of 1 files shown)


src/magic_folder/cli.py
```diff
@@ -898,96 +873,6 @@ def list_directory(self, uri):
...
-                b"GET",
-                uri,
-            )
-            if response.code != 200:
-                raise Exception(
-                    "Error response from download endpoint: {code} {phrase}".format(
-                        **vars(response)
-                    ))
-
-        returnValue((yield readBody(response)))
```
exarkun, line 919: `download_best_version` seems like a clear duplicate of `magic_folder.tahoe_client.TahoeClient.download_capability` so deleting this one seems good (unfortunate that we implemented this (at least) twice).

src/magic_folder/cli.py
```diff
@@ -898,96 +873,6 @@ def list_directory(self, uri):
...
-                            ],
-                        }).encode("utf-8"),
-                    ),
-                ),
-            )
-            if response.code != 200:
-                raise Exception("Error response from metadata endpoint: {code} {phrase}".format(
-                    **vars(response)
-                ))
-        returnValue(Node(self, from_string(filecap)))
```
exarkun, line 987: I think we do need this functionality.  How will files be linked into the DMD?  It seems like they would be linked in with the HTTP API this function wraps.  Is there also another implementation of this functionality somewhere?  I think @meejah was hitting his head against directory interactions in #217 - maybe there's some work there?  Or would this method help with that?

vu3rdd:
Closing this for now, we can revisit removal of dead code once the project stabilize a bit.

src/magic_folder/cli.py
meejah: This doesn't just "link a file into a DMD" like the `add_file` API (in the "other" client :( ) does .. this accepts an "uploadable" (a Tahoe hold-over interface, I think, since it's not documented?) and uploads it PLUS puts it in the DMD.

Also, I don't think this method as it sits is a good idea/implementation since it sucks the entire file into memory on the first line, so it won't work for anything "big" anyway (and duplicates any functionality for "upload a file").

PR #252, opened 2020-08-17
client api endpoint

### LINKED ISSUE #251: Store and configure "client string" for API endpoint separately
We currently configure an "API Endpoint" as a Twisted server-string. This allows us to listen in a generic fashion for API requests. However, we also need a way to *connect* to that endpoint but Twisted doesn't have a way to convert a server endpoint (or its string) to a corresponding client endpoint (or its string).

meejah:
fixes #251

### DIFF
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -63,9 +63,6 @@ from .snapshot import (
 from .common import (
     atomic_makedirs,
 )
-from .endpoints import (
-    server_endpoint_str_to_client,
-)
 
 # Export this here since GlobalConfigDatabase is what it's for.
 from ._endpoint_parser import (
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -190,8 +190,8 @@ def create_global_configuration(basedir, api_endpoint, tahoe_node_directory,
             (_global_config_version, )
         )
         cursor.execute(
-            "INSERT INTO config (api_endpoint, tahoe_node_directory) VALUES (?, ?)",
-            (api_endpoint, tahoe_node_directory.path)
+            "INSERT INTO config (api_endpoint, tahoe_node_directory, api_client_endpoint) VALUES (?, ?, ?)",
+            (api_endpoint, tahoe_node_directory.path, api_client_endpoint)
         )
 
     config = GlobalConfigDatabase(
@@ -460,7 +460,7 @@ class GlobalConfigDatabase(object):
         cursor.execute("SELECT api_client_endpoint FROM config")
         return cursor.fetchone()[0].encode("utf8")
 
-    @api_endpoint.setter
+    @api_client_endpoint.setter
     def api_client_endpoint(self, ep_string):
         # confirm we have a valid endpoint-string
         from twisted.internet import reactor  # uhm...
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -46,6 +46,7 @@ from nacl.encoding import (
 
 from twisted.internet.endpoints import (
     serverFromString,
+    clientFromString,
 )
 from twisted.python.filepath import (
     FilePath,
@@ -98,7 +99,8 @@ CREATE TABLE magic_folders
 CREATE TABLE config
 (
     api_endpoint TEXT,                -- Twisted server-string for our HTTP API
-    tahoe_node_directory TEXT         -- path to our Tahoe-LAFS client state
+    tahoe_node_directory TEXT,        -- path to our Tahoe-LAFS client state
+    api_client_endpoint TEXT          -- Twisted client-string for our HTTP API
 );
 """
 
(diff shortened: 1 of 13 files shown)


src/magic_folder/config.py
```diff
@@ -131,7 +133,8 @@ class SnapshotNotFound(Exception):
 ## sure how to do that w/o docs here
 
 
-def create_global_configuration(basedir, api_endpoint, tahoe_node_directory):
+def create_global_configuration(basedir, api_endpoint, tahoe_node_directory,
+                                api_client_endpoint=None):
```
exarkun, line 137: I'd consider making this a required argument and doing the conversion in the option parsing code.  That way if automatic conversion fails and no value is supplied, it's the option parsing code that can fail - reporting a usage error.  And this code is safe and free of the extra conditionals for handling bogus input/combination of inputs.

Put another way (more formally or maybe even objectively), the option parsing code is already partial (it will fail in some inputs) but this function is currently total (except for I/O failures).  Moving the checks on this value into the option parsing code leaves that code partial and leaves this function total instead of making them both partial as they are here.

Also, it looks like this is a string description and not an endpoint itself so a different name might be preferable.  The same goes for `api_endpoint`.

src/magic_folder/test/cli/test_magic_folder.py
```diff
@@ -463,6 +468,75 @@ def test_poll_interval(self):
...
+    def setUp(self):
+        super(ClientEndpoint, self).setUp()
+        self.basedir = FilePath(self.mktemp())
+        self.nodedir = self.useFixture(
+            NodeDirectory(FilePath(self.mktemp()))
+        )
+
+    def test_convert_tcp(self):
+        """
+        a tcp: endpoint can be autoconverted
```
exarkun, line 485: The syntax threw me at first.  Maybe `A "tcp:" endpoint can be autoconverted."`

src/magic_folder/test/cli/test_magic_folder.py
```diff
@@ -463,6 +468,75 @@ def test_poll_interval(self):
...
+    def test_set_client_endpoint(self):
+        """
+        setting the client endpoint succeeds with a valid endpoint
+        """
+        config = create_global_configuration(self.basedir, u"tcp:5555", self.nodedir.path)
+        config.api_client_endpoint = u"tcp:localhost:5555"
+        self.assertThat(
+            config.api_client_endpoint,
+            Equals(u"tcp:localhost:5555")
+        )
```
exarkun, line 529: Maybe we want even more than this.  The new value is also persisted and available to a different `GlobalConfigDatabase` instance sharing the same basedir.

Also, this is sort of a weird idea and probably a bad one but as soon as I read this I wanted `config.api_client_endpoint` to be an actual endpoint instead of a string description.  This would make the setter and getter asymmetric though (since making the setter accept an actual endpoint object would have all kinds of problems).

src/magic_folder/test/cli/test_magic_folder.py
```diff
@@ -463,6 +468,75 @@ def test_poll_interval(self):
...
+
+    def test_convert_tcp(self):
+        """
+        a tcp: endpoint can be autoconverted
+        """
+        config = create_global_configuration(self.basedir, u"tcp:5555", self.nodedir.path)
+        self.assertThat(
+            config.api_client_endpoint,
+            Equals(u"tcp:127.0.0.1:5555")
```
exarkun, line 490: I wonder what `u"tcp:555555"` does?  I expect it makes it into the database?  Not sure there's much to do about this here, just noting it.

src/magic_folder/test/cli/test_magic_folder.py
```diff
@@ -463,6 +468,75 @@ def test_poll_interval(self):
...
+            Equals(u"tcp:localhost:5555")
+        )
+
+    def test_set_invalid_client_endpoint(self):
+        """
+        setting the client endpoint fails (if it is invalid)
+        """
+        config = create_global_configuration(self.basedir, u"tcp:5555", self.nodedir.path)
+        with ExpectedException(ValueError, "Unknown endpoint type.*"):
+            config.api_client_endpoint = u"invalid:"
```
exarkun, line 537: Similarly to the above, we probably also want to know that the persisted value hasn't changed.

src/magic_folder/test/cli/test_magic_folder.py
meejah: Hmm, interesting.

I wonder if it should maybe be a `create_api_client_endpoint()` or similar even .. since it should actually take the `reactor` arg ideally. One issue with NOT returning the actual string would be e.g. `magic-folder show-config` or any config-viewing API -- because we can't go from `IEndpoint`-instance to text :(

src/magic_folder/test/cli/test_magic_folder.py
meejah: It does make it in, yeah :/ (I guess Twisted doesn't check port ranges etc?) .. but even if we e.g. tried to listen there (to see if it's a valid port on this OS), failing to do so isn't *necessarily* bad (e.g. maybe the port is just busy right now, or maybe it's completely invalid or maybe you haven't fixed AppArmor yet or ...). There will be errors that just won't be found/apparent until you `magic-folder run` anyway.

All that said though, probably the HTTP API for "change the config" *should* actually try and listen first and generate an error if it can't .. because if it can't then you'll be unable to change the config back to something that does work.

PR #257, opened 2020-08-18
Only accept valid endpoints for 'magic-folder migrate' or 'magic-folder init'

### LINKED ISSUE #226: `magic-folder migrate` accepts invalid endpoint descriptions
From https://github.com/LeastAuthority/magic-folder/pull/223#discussion_r464158318:

> I just learned that magic-folder migrate --listen-endpoint 6001 --node-directory ../path/to/node accepts that non-endpoint ar argument without complaining. Thus magic folder configuration will have just 6001 for api_endpoint, and passing that on will result in an error here.

meejah:
fixes #226

### DIFF
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -194,6 +194,9 @@ def create_global_configuration(basedir, api_endpoint_str, tahoe_node_directory,
         database=connection,
         api_token_path=basedir.child(b"api_token"),
     )
+    # check that these are valid by setting them
+    config.api_endpoint = api_endpoint_str
+    config.api_client_endpoint = api_client_endpoint_str
     # make sure we have an API token
     config.rotate_api_token()
     return config
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -51,6 +51,9 @@ from twisted.internet.endpoints import (
 from twisted.python.filepath import (
     FilePath,
 )
+from twisted.python.compat import (
+    nativeString,
+)
 
 from allmydata.uri import (
     from_string as tahoe_uri_from_string,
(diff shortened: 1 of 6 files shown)


meejah:
I don't know what "blocked" means for the fedora build causing "ci" to fail

src/magic_folder/config.py
```diff
@@ -183,6 +186,21 @@ def create_global_configuration(basedir, api_endpoint_str, tahoe_node_directory,
 
     :returns: a GlobalConfigDatabase instance
     """
+
+    # our APIs insist on endpoint-strings being unicode, but Twisted
+    # only accepts "str" .. so we have to convert on py2. When we
```
exarkun, line 191: "When" or "if" ...

src/magic_folder/config.py
```diff
@@ -222,6 +240,9 @@ def create_global_configuration(basedir, api_endpoint_str, tahoe_node_directory,
         database=connection,
         api_token_path=basedir.child(b"api_token"),
     )
+    # check that these are valid by setting them
+    config.api_endpoint = api_endpoint_str
+    config.api_client_endpoint = api_client_endpoint_str
```
exarkun, line 245: It would be cool if we *only* inserted values into the database through a Python setter interface (ie, if we didn't need the `INSERT INTO` SQL statement a few lines above this code).

A runner-up might be if there were an explicit validation API that could be invoked separate from the setter so we don't need to invoke the setter just for the validation side-effect.

src/magic_folder/test/test_cli.py
```diff
@@ -138,6 +144,39 @@ def test_good(self):
             Equals([u"test-folder"]),
         )
 
+    def test_bad_listen_string(self):
```
exarkun, line 147: This could use a docstring.

src/magic_folder/test/test_cli.py
```diff
@@ -138,6 +144,39 @@ def test_good(self):
...
+                self.temp.child("new_magic"),
+                "1234",
+                self.node_dir.path,
+                u"alice",
+                None,
+            )
+
+    def test_invalid_listen_string(self):
+        """
+        Pass the wrong type of object for listen-string
```
exarkun, line 159: This docstring could use a description of expected behavior.

src/magic_folder/test/test_cli.py
```diff
@@ -138,6 +144,39 @@ def test_good(self):
...
+        with ExpectedException(ValueError):
+            magic_folder_migrate(
+                self.temp.child("new_magic"),
+                b"1234",  # only accepts unicode
+                self.node_dir.path,
+                u"alice",
+                b"invalid",
+            )
+
+    def test_bad_connect_string(self):
```
exarkun, line 170: This could use a docstring.

src/magic_folder/config.py
meejah: Thinking optimistically ;)

src/magic_folder/config.py
meejah: Might just be my lack of SQL/SQLite understanding .. I believe my thinking there is that the `INSERT INTO` above is to put a row in so that the setters can always just `UPDATE`.

But, okay so yeah if I move the validators separately we can call them here before the INSERT (and call the same ones in the setters).

PR #259, opened 2020-08-18
258.signatures

meejah:


docs/snapshots.rst
```diff
@@ -91,3 +93,33 @@ they next poll our mutable directory.
...
+Signature Details
+-----------------
+
+As above, we upload the content and metadata as two distinct immutable
+capability. There is also the "mangled" name of the file. This mangled
+name will also point at the Snapshot from the Personal DMD.
+
+The signature scheme is to concatenate all three of the above
+(separated by newlines) and then produce a signature over it using the
+author's signing key (which is a `nacl.signing.SigningKey` from the
```
exarkun, line 107: My cryptographic question here is whether this is what `nacl.signing.SigningKey` is for.  It *seems* like it is.  https://pynacl.readthedocs.io/en/latest/signing/ says "A valid digital signature gives a recipient reason to believe that the message was created by a known sender such that they cannot deny sending it (authentication and non-repudiation) and that the message was not altered in transit (integrity)." which I think is what you want from a signature system.  It would be good to confirm this understanding somehow.

src/magic_folder/snapshot.py
```diff
@@ -216,35 +216,39 @@ def create_author_from_json(data):
...
-    # cryptograhphers yet).
+    # this hasn't been looked at by our cryptograhphers yet).
     data_to_sign = (
         u"{content_capability}\n"
+        u"{metadata_capability}\n"
         u"{name}\n"
     ).format(
         content_capability=content_capability,
+        metadata_capability=metadata_capability,
         name=snapshot.name,
```
exarkun, line 246: This is more of our path confusion.  I *don't* think `snapshot.name` will be mangled in the long run.  It will be a normal relative path.  I think it is *sometimes* the mangled path currently, but probably the wrong mangled path (absolute instead of relative).  This is a separate issue that I'm not suggesting be addressed by this PR, just pointing it out.

Maybe the unmangled path is a better thing to sign anyway?  The mangling is sort of a protocol/storage detail that it might be nice to be able to change.  The *real* relative file name is part of the user data and we shouldn't want to (or be able to) change it without breaking signatures, I think.

docs/snapshots.rst
exarkun: Okay I have another cryptographic question, actually.  Should we tag the parts of the value to be signed?  eg

```
content-cap:<content cap>
metadata-cap:<metadata cap>
path:<path>
```

I have some vague idea about the purpose of tagging ... it prevents accidental collisions between different "kinds" of strings that might accidentally have the same value sometimes?  I'm not sure if that's actually true or not, it's just what's in my head.

I'm not exactly sure why it is important here either, except maybe that it somehow does the thing I described above in some circumstance.

src/magic_folder/snapshot.py
```diff
@@ -253,9 +257,11 @@ def verify_snapshot_signature(remote_author, alleged_signature, content_capabili
     # See comments about "data_to_sign" in sign_snapshot
     data_to_verify = (
         u"{content_capability}\n"
+        u"{metadata_capability}\n"
         u"{name}\n"
     ).format(
         content_capability=content_capability,
+        metadata_capability=metadata_capability,
         name=snapshot_name,
     )
```
exarkun, line 266: Maybe the implementation could be refactored so we don't have two duplicate bits of code constructing the string that was signed.

src/magic_folder/snapshot.py
```diff
@@ -216,35 +216,39 @@ def create_author_from_json(data):
...
     # https://github.com/LeastAuthority/magic-folder/issues/190 as
-    # this is likely insufficient (and hasn't been looked at by our
-    # cryptograhphers yet).
+    # this hasn't been looked at by our cryptograhphers yet).
     data_to_sign = (
         u"{content_capability}\n"
+        u"{metadata_capability}\n"
         u"{name}\n"
     ).format(
         content_capability=content_capability,
```
exarkun, line 244: Formatting bytes into unicode might not be ideal.  The capabilities are bytes, per the docstring.  The path is unicode, I think.  Maybe encoding the path first and then interpolating into bytes is better?

The only practice difference this might make is if the capabilities have non-ascii in them somehow, then the interpolation will fail with a `UnicodeDecodeError`, I think.

src/magic_folder/snapshot.py
```diff
@@ -420,15 +426,17 @@ def create_snapshot_from_capability(snapshot_cap, tahoe_client):
...
-
-        author = create_author_from_json(snapshot_author)
+        # NB: once we communicate authors + public-keys some other
+        # way, we could iterate all our known authors and attempt to
+        # verify a signature against the snapshot + metdata
+        # capabilities at this point .. that might be better anyway?
+        # (A key advantage is not even trying to deserialize anything
+        # that's not verified by a signature).
+        metadata_cap = snapshot["metadata"][1]["ro_uri"]
+        author_signature = snapshot["metadata"][1]["metadata"]["magic_folder"]["author_signature"]
```
exarkun, line 436: Demeter is weeping.  It's a Tahoe-LAFS data structure though, so too bad for Demeter, I guess?

src/magic_folder/snapshot.py
```diff
@@ -420,15 +426,17 @@ def create_snapshot_from_capability(snapshot_cap, tahoe_client):
...
-        author_json = yield tahoe_client.download_capability(author_cap)
-        snapshot_author = json.loads(author_json)
-
-        author = create_author_from_json(snapshot_author)
+        # NB: once we communicate authors + public-keys some other
+        # way, we could iterate all our known authors and attempt to
+        # verify a signature against the snapshot + metdata
+        # capabilities at this point .. that might be better anyway?
+        # (A key advantage is not even trying to deserialize anything
+        # that's not verified by a signature).
```
exarkun, line 434: Perhaps we could put the author pubkey into the snapshot as well so we don't have to try every known author, we can just look up the one with the matching key.

Or maybe looking up the author by the "name" field is good enough.

docs/snapshots.rst
```diff
@@ -91,3 +93,33 @@ they next poll our mutable directory.
...
+PyNaCl library (which uses `libsodium` under the hood). See
+https://pynacl.readthedocs.io/en/latest/signing/
+
+The resulting signature is base32-encoded and included in the "tahoe
```
exarkun, line 111: The implementation uses base64.  Either of these seems fine.  I'm not sure why we would pick base32 over base64 except for some kind of consistency with Tahoe-LAFS' choice of text-ish encoding function.

src/magic_folder/snapshot.py
```diff
@@ -696,22 +709,25 @@ def write_snapshot_to_tahoe(snapshot, author_key, tahoe_client):
...
+            "filenode", {
+                "ro_uri": content_cap,
+            },
+        ],
+        "metadata": [
+            "filenode", {
+                "ro_uri": metadata_cap,
+                "metadata": {
+                    "magic_folder": {
+                        "author_signature": author_signature_base64,
```
exarkun, line 723: All the strings in this object should be unicode, I think.

src/magic_folder/snapshot.py
```diff
@@ -723,15 +739,12 @@ def write_snapshot_to_tahoe(snapshot, author_key, tahoe_client):
     # failure, we'll try again.
     returnValue(
         RemoteSnapshot(
-            # XXX: we are copying over the name from LocalSnapshot, it is not
-            # stored on tahoe at the moment. This means, when we read back a snapshot
-            # we cannot create a RemoteSnapshot object from a cap string.
             name=snapshot.name,
             author=create_author(  # remove signing_key, doesn't make sense on remote snapshots
```
exarkun, line 743: Sort of unrelated but this could be `author.to_remote_author()` instead.

src/magic_folder/snapshot.py
```diff
@@ -723,15 +739,12 @@ def write_snapshot_to_tahoe(snapshot, author_key, tahoe_client):
...
                 name=snapshot.author.name,
                 verify_key=snapshot.author.verify_key,
             ),
-            metadata=content_metadata,  # XXX not authenticated by signature...
+            metadata=snapshot_metadata,
             parents_raw=parents_raw,  # XXX FIXME (at this point, will have parents' immutable caps .. parents don't ork yet)
```
exarkun, line 748: So ... I think this means the parent references aren't signed?  Does that mean history is forgeable?  (eg I could take your snapshot objects and rewrite the parent pointers, and your signature would still verify?)

If so, I guess parents belong in `snapshot_metadata` instead, so this isn't possible (still could have an accessor on `RemoteSnapshot` so application code doesn't have to go digging through a dict directly).

docs/snapshots.rst
```diff
@@ -91,3 +93,33 @@ they next poll our mutable directory.
...
+capability-strings when they download the Snapshot object and the
+mangled name from that participant's Personal DMD. They can thus
+verify the signature using the participant's public-key before
+downloading the content or metadata.
+
+**Note**: "metadata" is used in two ways above; Tahoe allows you to
+add arbitrary metadata for each entry in a directory. This is where
+we locate the base32-encoded signature itself. The "metadata about
+the Snapshot" is its own capability (consisting of JSON-encoded
+metadata).
```
exarkun, line 125: I didn't realize until very near the end of the diff but it seems like we should specify what "metadata about the Snapshot" consists of, exactly.  Since we're relying on the signature to validate the metadata for us, we should be sure everything we need to validate about a  snapshot (except for content) is included in the "metadata about the Snapshot".

Noted elsewhere it may make sense to include the parent capabilities in this metdata.  Also, maybe the path should be part of the metadata as well, instead of a third top-level-ish thing.

src/magic_folder/snapshot.py
meejah: I don't think it really matters either way here (of course, we need to consistently use one). Whatever "name" happens to be, we will sign it ... and only accept one on the other side that's got a valid signature. How that is interpreted by the rest of the system doesn't (shouldn't!) matter here.

It does seem *slightly* better, maybe, to sign the "real actual relative path" .. but also if the name mangling/de-mangling is consistent, does it matter? If "sign the relative path" is best we should just make sure that `name` refers to that I think so this code doesn't have to "know" anything about what's in the name.

src/magic_folder/snapshot.py
meejah: Tahoe actually stores this stuff differently -- but yeah, we're consistent with the JSON output of the Tahoe WebUI here.

src/magic_folder/snapshot.py
meejah: The author _is_ in the snapshot currently (including pubkey). However, there is no way to **verify** this key yet. The "real" solution is to communicate the keys along with participants..

src/magic_folder/snapshot.py
meejah: Hmm, yeah good point. The capability of the **Snapshot** includes the parents but yes that isn't (can't be) part of the signature of THIS snapshot.

meejah: okay, I moved the parents to the metadata

meejah:
okay I believe I have addressed everything from the review.

meejah:
I'm still really confused about what codecov is trying to say. It seems to be sad about `snapshot.py`

There's literally only one uncovered line in `snapshot.py` (an error-case) but I don't think this PR touches it (and in any case, codecov also says there's 100% patch coverage). Is it just because snapshot.py shrunk enough in this PR and so that one uncovered line + branch is now enough to drop coverage?

docs/snapshots.rst
```diff
@@ -74,20 +94,65 @@ even if we can't communicate to Tahoe-LAFS.
...
 The upload is a multi-part process:
 
-- upload the content, yielding an immutable read-capability
+- upload the content yielding an immutable read-capability
+
+- upload the metadata yielding an immutable read-capability
 
 - use the local keypair to sign the snapshot (requires the
-  read-capability from the first step)
+  read-capabilities from from the first two steps)
```
exarkun, line 102: duplicated `from` here

src/magic_folder/snapshot.py
```diff
@@ -628,60 +600,51 @@ def write_snapshot_to_tahoe(snapshot, author_key, tahoe_client):
...
+        "snapshot_version": SNAPSHOT_VERSION,
+        "name": snapshot.name,
+        "author": snapshot.author.to_remote_author().to_json(),
+        "parents": [
+            parent_cap.encode("utf8")
+            for parent_cap in parents_raw
+        ]
+    }
+    metadata_cap = yield tahoe_client.create_immutable(
+        json.dumps(snapshot_metadata)
```
exarkun, line 614: Not necessarily for this PR but the metadata and content could be uploaded in parallel instead of serially.

src/magic_folder/snapshot.py
```diff
@@ -628,60 +600,51 @@ def write_snapshot_to_tahoe(snapshot, author_key, tahoe_client):
...
     )
-    # print("author_cap: {}".format(author_cap))
+
+    # sign the snapshot (which can only happen after we have the
+    # content-capability and metadata-capability)
+    author_signature = sign_snapshot(author_key, snapshot, content_cap, metadata_cap)
+    author_signature_base64 = base64.b64encode(author_signature.signature)
```
exarkun, line 620: The `sign_snapshot` docs say it returns `bytes` but it looks like it returns something else instead.

src/magic_folder/test/test_local_snapshot.py
```diff
@@ -244,7 +244,7 @@ def test_create_snapshots(self, filenames, data):
         files = []
         for filename in filenames :
             file = self.magic.child(filename)
-            content = data.draw(binary())
+            content = snapshot_data.draw(binary())
             file.asBytesMode("utf-8").setContent(content)
```
exarkun, line 248: The changes here might be superfluous.  For what it's worth, the parameter that was named `data` before isn't particularly tied to the snapshot data itself.  It is a thing that can draw examples from strategies.  In the `data.draw(binary())` expression, it's `binary()` that is the snapshot data strategy.

If the motivation for the change was the collision between the module-scope `data` name and the `data` parameter, I'm sympathetic.  I'm not sure what a good name improvement is though.  A good non-name-improvement would be to replace the strategy for `filenames` with something like:

```
dictionaries(
    path_segments().map(lambda p: p.encode("utf-8")),
    binary(),
)
```
to produce a dictionary mapping filenames to their contents without using the `data` strategy at all.  The `unique` constraint remains implicitly by virtue of dict keys having to be unique always.

I don't think there is a good reason for the way the test is written now apart from unfamiliarity with different strategy tricks like this.

src/magic_folder/snapshot.py
meejah: True. Not sure if it'll make much difference because Tahoe does uploads in serial (but downloads in parallel). However, maybe the WebUI returns before all the data has been uploaded to all the storage-servers? (I don't recall).

src/magic_folder/test/test_local_snapshot.py
meejah: yes, i was just trying to get rid of the name shadowing

PR #262, opened 2020-08-20
261.in memory magic folder config

### LINKED ISSUE #261: Support in-memory use of `MagicFolderConfig`
This will make a number of unit tests much faster (one I measured ran in 20% of the time w/ in-memory vs on filesystem).

exarkun:
Fixes #261

### DIFF
--- a/src/magic_folder/test/test_config.py
+++ b/src/magic_folder/test/test_config.py
@@ -199,10 +199,13 @@ class EndpointDescriptionConverterTests(SyncTestCase):
         )
 
 
-class TestMagicFolderConfig(SyncTestCase):
-
+class GlobalConfigDatabaseMagicFolderTests(SyncTestCase):
+    """
+    Tests for the ``GlobalConfigDatabase`` APIs that deal with individual
+    ``MagicFolderConfig`` instances.
+    """
     def setUp(self):
-        super(TestMagicFolderConfig, self).setUp()
+        super(GlobalConfigDatabaseMagicFolderTests, self).setUp()
         self.temp = FilePath(self.mktemp())
         self.node_dir = FilePath(self.mktemp())
         self.tahoe_dir = self.useFixture(NodeDirectory(self.node_dir))
--- a/src/magic_folder/test/test_config.py
+++ b/src/magic_folder/test/test_config.py
@@ -50,6 +50,8 @@ from .strategies import (
     magic_folder_filenames,
 )
 from ..config import (
+    SQLite3DatabaseLocation,
+    MagicFolderConfig,
     endpoint_description_to_http_api_root,
     create_global_configuration,
     load_global_configuration,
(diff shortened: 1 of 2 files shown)


src/magic_folder/test/test_config.py
```diff
@@ -323,25 +328,23 @@ class StoreLocalSnapshotTests(SyncTestCase):
...
     def setup_example(self):
         self.temp = FilePath(self.mktemp())
-        self.global_db = create_global_configuration(
-            self.temp.child(b"global-db"),
-            u"tcp:12345",
-            self.temp.child(b"tahoe-node"),
-            u"tcp:localhost:12345",
-        )
+        self.stash = self.temp.child("stash")
+        self.stash.makedirs()
```
meejah, line 336: This is redundant (I think) since `initialize` will do atomic_makedirs on the stash-path?

exarkun:
> Looks good.
> 
> Does `SQLite3DatabaseLocation` allow a second instance of a config (or whatever) to open the **same** in-memory database? (I'm thinking of those tests that "do something" and then check if the "actual" database was changed -- or not -- by loading a second instance). Maybe this is more just a question about what `":memory:"` actually does :)

SQLite3 has this functionality but the sqlite3 module in the Python 2.7 stdlib doesn't let you get at it.  The sqlite3 module in the Python 3.something stdlib does, and `SQLite3DatabaseLocation` will be a good place to shove that functionality if we want it, if we switch to Python 3.

exarkun:
> > Looks good.
> > Does `SQLite3DatabaseLocation` allow a second instance of a config (or whatever) to open the **same** in-memory database? (I'm thinking of those tests that "do something" and then check if the "actual" database was changed -- or not -- by loading a second instance). Maybe this is more just a question about what `":memory:"` actually does :)
> 
> SQLite3 has this functionality but the sqlite3 module in the Python 2.7 stdlib doesn't let you get at it. The sqlite3 module in the Python 3.something stdlib does, and `SQLite3DatabaseLocation` will be a good place to shove that functionality if we want it, if we switch to Python 3.

Although I guess we could do this without relying on SQLite3 features at all.  We could have an alternate implementation where `connect` always returns the same connection object.

src/magic_folder/test/test_config.py
exarkun: The `atomic_makedirs` calls got left behind in `create_magic_folder` because they are trying to encapsulate both the `MagicFolderConfig` initialization _and_ the global database configuration changes.  Coordinating the three difference pieces of state (filesystem, global db, per-folder db) is definitely awkward :/  Seems like there should be a better way to manage it but I don't know what that is yet.

exarkun:
> > Looks good.
> > Does `SQLite3DatabaseLocation` allow a second instance of a config (or whatever) to open the **same** in-memory database? (I'm thinking of those tests that "do something" and then check if the "actual" database was changed -- or not -- by loading a second instance). Maybe this is more just a question about what `":memory:"` actually does :)
> 
> SQLite3 has this functionality but the sqlite3 module in the Python 2.7 stdlib doesn't let you get at it. The sqlite3 module in the Python 3.something stdlib does, and `SQLite3DatabaseLocation` will be a good place to shove that functionality if we want it, if we switch to Python 3.

The SQLite3 functionality I'm referring to here is the `memory` value for the `mode` flag in SQLite3 database URIs:

> If two or more distinct but shareable in-memory databases are needed in a single process, then the mode=memory query parameter can be used with a URI filename to create a named in-memory database:
> rc = sqlite3_open("file:memdb1?mode=memory&cache=shared", &db);

(https://www.sqlite.org/inmemorydb.html)

But Python 2.7 doesn't expose URI filenames to us at all.

PR #270, opened 2020-08-24
HTTP API for local snapshot creation

### LINKED ISSUE #266: Add an HTTP API to create a local snapshot of a file
This will allow manual triggering of snapshot creation independent of any filesystem-change-notification system that might eventually be hooked up.  This mode of operation may be interesting to end users eventually or it may just be a useful stepping stone towards full "magic" operation.

exarkun:
Fixes #266

### DIFF
--- a/docs/interface.rst
+++ b/docs/interface.rst
@@ -80,12 +80,14 @@ The values for the ``local`` property can be resolved against the ``identifier``
 ``GET /v1/snapshot/:folder-name``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-        Get all snapshots for one folder:
+Not yet implemented.
+Get all snapshots for one folder.
 
 ``GET /v1/snapshot/:folder-name?path=:some-path``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-        Get all snapshots for one folder beneath a certain path:
+Not yet implemented.
+Get all snapshots for one folder beneath a certain path.
 
 ``POST /v1/snapshot/:folder-name?path=:some-path``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/interface.rst
+++ b/docs/interface.rst
@@ -1,10 +1,29 @@
 Interface
 =========
 
-HTTP Blah Blah
---------------
+HTTP API
+--------
 
-When enabled ...
+Magic-Folder exposes all of its supported functionality in an HTTP API.
+The address of the HTTP server is part of the `daemon configuration`_.
+
+Authorization
+~~~~~~~~~~~~~
+
+The HTTP API is protected by Bearer token-style authorization scheme.
+Only requests which include the correct token will receive successful responses.
+Other requests will receive **401 Unauthorized** responses which omit the requested resource.
+The token value should be included with the **Bearer** scheme in the **Authorization** header element.
+For example::
+
+  Authorization: Bearer abcdefghijklmnopqrstuvwxyz
+
+The correct token value can be found in the *api_token* file inside the Magic-Folder daemon configuration directory.
+The token value is periodically rotated so clients must be prepared to receive an **Unauthorized** response even when supplying the token.
+In this case,
+the client should re-read the token from the filesystem to determine if the value held in memory has become stale.
+
+.. _`daemon configuration`: :ref:`config`
 
 ``GET /v1/magic-folder``
 ~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/interface.rst
+++ b/docs/interface.rst
@@ -1,0 +1,67 @@
+Interface
+=========
+
+HTTP Blah Blah
+--------------
+
+When enabled ...
+
+``GET /v1/magic-folder``
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+This endpoint returns ...
+
+The response is **OK** with an ``application/json`` **Content-Type**::
+
+  { ...
+  }
+
+``GET /v1/snapshot``
+~~~~~~~~~~~~~~~~~~~~
+
+This endpoint returns information about all snapshots known to exist.
+
+The response is **OK** with an ``application/json`` **Content-Type**::
+
+  {"folder name": {
+      "foo/bar": [
+        { "type": "local"
+        , "identifier": 2
+        , "parents": [{"local": 1}, {"remote": "URI:CHK:..."}]
+        }
+      ]
+  }}
+
+Properties of the top-level object map the name of a magic-folder to second-level objects describing snapshots for each file in that folder.
+Properties of second-level objects map relative file paths to lists of snapshots for that magic-folder.
+Elements of the list are objects with describe a single snapshot.
+
+Local snapshots are represented like this::
+
+  { "type": "local"
+  , "identifier": "0d585a0e-c39c-4dec-affb-cbab34245370"
+  , "parents": [{"local": "a3eb3d57-5272-45f3-ba5f-04a52024785b"}]
+  ,
+  }
+
+The parents property gives a list of references to snapshots which are parents of the containing snapshot.
+A local snapshot reference in this list is represented like this::
+
+  { "local": "a3eb3d57-5272-45f3-ba5f-04a52024785b" }
+
+The values for the ``local`` property can be resolved against the ``identifier`` described above.
+
+``GET /v1/snapshot/:folder-name``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+        Get all snapshots for one folder:
+
+``GET /v1/snapshot/:folder-name?path=:some-path``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+        Get all snapshots for one folder beneath a certain path:
+
+``POST /v1/snapshot/:folder-name?path=:some-path``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+        Create a new snapshot for a file at a certain path in a certain folder:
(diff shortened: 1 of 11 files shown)


src/magic_folder/magic_folder.py
```diff
@@ -298,21 +302,36 @@ def from_config(cls, reactor, tahoe_client, name, config):
...
+        return u"magic-folder-{}".format(self.folder_name)
+
+    def __init__(self, client, config, name, local_snapshot_service, initial_participants, clock):
+        super(MagicFolder, self).__init__()
+        self.folder_name = name
+        self._clock = clock
         self._config = config  # a MagicFolderConfig instance
         self._participants = initial_participants
+        self.local_snapshot_service = local_snapshot_service
+        local_snapshot_service.setServiceParent(self)
```
vu3rdd, line 334: don't we also need to call `startService()` on `local_snapshot_service`?

vu3rdd: >  don't we also need to call startService() on local_snapshot_service?

Looks like we don't need to. Starting the MagicFolder service seem to start the children services.

docs/interface.rst
```diff
@@ -0,0 +1,106 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+This endpoint returns a list of all individual magic-folders managed by this daemon.
+
+The response code **OK** and the **Content-Type** is ``application/json``.
+The response body follows the form of this example::
+
+  { "folders":
+    [ { "name": "Alice's music", "local-path": "/home/alice/Music" }
+    , { "name": "Secret docs", "local-path": /home/alice/secrets" }
```
meejah, line 38: Perhaps not actionable for this PR, but is there a reason we show only these two bits of config? What about the directory-capabilities? Author name, keys?

Internally we call the path `magic_directory` so it might be better to be consistent (versus the state-location versus the stash-path for example).

docs/interface.rst
```diff
@@ -0,0 +1,106 @@
...
+~~~~~~~~~~~~~~~~~~~~
+
+Retrieve information about all snapshots known to exist.
+
+The response is **OK** with an ``application/json`` **Content-Type**::
+
+  {"folder name": {
+      "foo/bar": [
+        { "type": "local"
+        , "identifier": "06be2d83-2d86-402d-ae2a-81b3779d72d9"
```
meejah, line 52: Is this leading-comma syntax .. normal? (I've not seen it before).

docs/interface.rst
```diff
@@ -0,0 +1,106 @@
...
+  { "type": "local"
+  , "identifier": "0d585a0e-c39c-4dec-affb-cbab34245370"
+  , "parents": [{"local": "a3eb3d57-5272-45f3-ba5f-04a52024785b"}]
+  ,
+  }
+
+The parents property gives a list of references to snapshots which are parents of the containing snapshot.
+A local snapshot reference in this list is represented like this::
+
+  { "local": "a3eb3d57-5272-45f3-ba5f-04a52024785b" }
```
meejah, line 76: So there will be both `"local"` and `"remote"` parents, represented by changing the string here? Might be worth some words here (I do see an example of that above though).

docs/interface.rst
```diff
@@ -0,0 +1,106 @@
...
+Not yet implemented.
+Get all snapshots for one folder beneath a certain path.
+
+``POST /v1/snapshot/:folder-name?path=:some-path``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Create a new snapshot for a certain file in a certain magic-folder.
+
+The ``path`` query argument is required.
+It must be a filesystem path relative to the selected magic-folder.
```
meejah, line 98: Is there a leading `":"` in this relative-path? If so, that should be mentioned here; if not, the ":" in the example/titles should probably be removed.

docs/interface.rst
```diff
@@ -0,0 +1,106 @@
...
+A new snapshot will be created for the file it identifies.
+
+The response is delayed until the local state tracking the snapshot has been created.
```
meejah, line 101: :+1:

src/magic_folder/test/test_web.py
```diff
@@ -246,7 +267,7 @@ def authorized_request(treq, auth_token, method, url):
     )
 
 
-def treq_for_folders(basedir, auth_token, folders):
+def treq_for_folders(reactor, basedir, auth_token, folders, start_folder_services):
```
meejah, line 270: "reactor" is pretty obvious, but I guess it should be added to the docstring too..

src/magic_folder/test/test_web.py
```diff
@@ -260,13 +281,22 @@ def treq_for_folders(basedir, auth_token, folders):
...
+        # Make this endpoint string and the one below parse but make them
+        # invalid, too, because we don't want anything to start listening on
+        # these during this set of tests.
+        u"tcp:-1",
+        # It wants to know where the Tahoe-LAFS node directory is but we don't
+        # have one and we don't want to invoke any functionality that requires
+        # one.  Give it something bogus.
         FilePath(u"/non-tahoe-directory"),
-        u"non-endpoint:",
+        u"tcp:-1",
```
meejah, line 299: I don't know if this is viable, but can we register an endpoint plugin for testing? So then we could do "real endpoint stuff" (like listen) but avoid doing "real network stuff" too, maybe. I don't want to hold up this PR for that, just asking/noting it here after seeing the other comment below (about starting just some services).

src/magic_folder/test/test_web.py
```diff
@@ -385,3 +431,126 @@ def test_list_folders(self, auth_token, folders):
...
+
+        self.assertThat(
+            authorized_request(
+                treq,
+                AUTH_TOKEN,
+                b"POST",
+                url_to_bytes(self.url.child(folder_name).set(u"path", path_in_folder)),
+            ),
+            has_no_result(),
+        )
```
meejah, line 477: It's not clear to me how this test achieves what the docstring says. How do we know if there is (or is not yet) an entry in the local database? Thinking out load, I'd expect a test like this to:
- "start the request"
- (maybe something here, but what? do we care if the DB is or is not modified at this point?)
- wait for the response
- confirm that we have now modified the local database

Maybe I'm reading the test wrong, but all it seems to assert is that the POST doesn't immediately return (but nothing about the local effects).

I guess the flip-side of this would be a test that does something such that there _should_ be no local effects (like maybe cancelling the request right at the last byte? or the second-last byte?). This might be tricky ;)

src/magic_folder/web.py
```diff
@@ -96,10 +103,182 @@ class APIv1(Resource, object):
...
+        # FilePath so it's a bad idea.  However, add_file is going to apply an
+        # equivalent safety check to the one we're bypassing.  So ... it's
+        # okay.
+        #
+        # Maybe we should have a "relative path" type that has safety features
+        # in it and then any safety checks can be performed early, near the
+        # code accepting user input, instead of deeper in the model.
+        path = self._folder_config.magic_path.preauthChild(path_u)
+
+        # TODO error handling?
```
meejah, line 167: Yeah, probably .. what will happen here? I guess we're depending on the Twisted's "unhandled error in deferred" code to write a traceback somewhere?

docs/interface.rst
exarkun: No particular reason, just the two obvious pieces of state.  This should be easy to expand when it becomes clear what's needed - by GridSync, I suppose?  Will GridSync need to know the caps?  I guess it does create some files on the grid somewhere but I don't think it creates them inside the collective or dmd directories ... I can't remember exactly right now.  And maybe it will change anyway?  We should get @crwood involved here.

It does seem like any changes here can wait for a future PR though.  At that point it'll just be small twiddles to the resource, I think.

docs/interface.rst
exarkun: It's the usual syntax but I guess the formatting is uncommon.  It's motivated by two things:

  * Cleaner diffs since the comma and the element that requires it end up on the same line.  This means deleting a line doesn't require changing the line above it (and similar improvement for adding a line).
  * JSON doesn't allow a dangling comma (final comma with no element after it) so the pattern that works for, eg, Python import lists (with a comma after every element, even if there is no subsequent element) doesn't work here.

docs/interface.rst
exarkun: Ah, there isn't.  Some API doc systems use `:` to introduce a parameter but maybe it's not sufficiently widespread to assume it will make sense without explanation.  `<folder-name>` would probably be clearer?  I can switch to that notation.

src/magic_folder/test/test_web.py
exarkun: We can do that, yea, and it's a big strategy I used in Tahoe-LAFS to fix a lot of the "address already in use" errors.  It's a lot of work though, which is why I skipped it for now here and did this cheesy thing.

I'll leave a comment here explaining this situation, though, linked to a ticket.

src/magic_folder/test/test_web.py
exarkun: > It's not clear to me how this test achieves what the docstring says. How do we know if there is (or is not yet) an entry in the local database? Thinking out load, I'd expect a test like this to:

In this case, there's a tricky, indirect way that it knows which should probably be improved.  The way it knows is that it hasn't started the snapshot creation service so the API call can *never* result in a local snapshot being created.  I'll see if I can improve the test to make this more clear.

> I guess the flip-side of this would be a test that does something such that there should be no local effects (like maybe cancelling the request right at the last byte? or the second-last byte?). This might be tricky ;)

Perhaps, but I think that would just be a Twisted Web test that demonstrates incomplete requests are never dispatched to application code.

src/magic_folder/web.py
exarkun: Yea.  Meanwhile the request will just be left hanging.  This should be better.  I think it's awkward to test because how do I get `add_file` to fail?  Define an interface for that object and build a new verified fake for it ...

exarkun: Eh I went the cheesy route and just found an error case of the real implementation and exploited it.

docs/interface.rst
meejah: Sure, I guess it can be future. I think it'll need to export ALL the information (e.g. to support moving the CLI over, which does list all those things). For example, via `magic-folder list --include-secret-information`

PR #271, opened 2020-08-26
normalize snapshot schema

### LINKED ISSUE #197: Normalize the `local_snapshots` database schema
Currently we have this schema:

```
CREATE TABLE local_snapshots
(
 path               TEXT PRIMARY KEY,             -- UTF-8 relative filepath that the snapshot represents
 snapshot_blob      BLOB                          -- a JSON blob representing the snapshot instance
);
```

`snapshot_blob` holds all of the rest of the snapshot structure.  Since this is all hidden inside a BLOB it's effectively unqueryable at the SQL level.  Also, because of the way the blob is constructed, most rows of the table contain redundant information (since the parent of each snapshot is serialized along with it).

Here's a different possible schema:

```
CREATE TABLE local_snapshots
(
 id                    INTEGER PRIMARY KEY -- unique snapshot identifier
 path               TEXT,             -- relative filepath that the snapshot represents
 author            TEXT,             -- the author name
 content_path TEXT,             -- stashed content filesystem path 
);

CREATE TABLE local_snapshot_parents
(
 child_id    INTEGER -- the local_snapshots.id of the child in the relationship
 parent_id INTEGER -- the local_snapshots.id of the parent in the relationship
);
```

This leaves out metadata.  Metadata could be represented in a few ways:

* As an opaque blob in the local_snapshots table.  This has the same drawbacks as the previous opaque blob but at least it's restricted to metadata.
* As first-class columns in the local_snapshots table.  This is most useful from the SQL perspective.  If metadata is relatively fixed (changing from release to release, perhaps, but not from snapshot to snapshot or user to user) then this is pretty good.  Things like "created date" fit well here.
* As elements in a separate local_metadata table with two columns - "key" and "value".  This is a kind of compromise.  It's queryable though only awkwardly but still allows fully dynamic metadata.
* Some hybrid of the two - for example, keeping everything we want to query on as columns in the local_snapshots table and shoving the rest into a JSON blob.  These two kinds of representation could be fused/split to obscure the storage representation choice from the Python application layer - unless we thought preserving it would improve the Python code (which it might... consider `snapshot.created` vs `snapshot.metadata[u"ctime"]`).

exarkun:
Fixes #197

~~This is stacked on #266~~

### DIFF
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -133,7 +133,7 @@ _magicfolder_config_schema = Schema([
 
             -- The magicpath-mangled name of the file this snapshot is for,
             -- UTF-8-encoded.
-            [path]             TEXT,
+            [name]             TEXT,
 
             -- A local filesystem path where the content can be found.
             [content_path]     TEXT
@@ -405,7 +405,7 @@ def _get_snapshots(cursor, path):
         FROM
             [local_snapshots]
         WHERE
-            path = ?
+            [name] = ?
         """,
         (path,),
     )
@@ -435,7 +435,7 @@ def _get_metadata(cursor, path):
         WHERE
             [metadata].[snapshot_identifier] = [local_snapshots].[identifier]
         AND
-            [local_snapshots].[path] = ?
+            [local_snapshots].[name] = ?
         """,
         (path,),
     )
@@ -472,7 +472,7 @@ def _get_parents(cursor, path):
         WHERE
             [parents].[snapshot_identifier] = [local_snapshots].[identifier]
         AND
-            [local_snapshots].[path] = ?
+            [local_snapshots].[name] = ?
         """,
         (path,),
     )
@@ -755,7 +755,7 @@ class MagicFolderConfig(object):
             cursor.execute(
                 """
                 INSERT INTO
-                    [local_snapshots] ([identifier], [path], [content_path])
+                    [local_snapshots] ([identifier], [name], [content_path])
                 VALUES
                     (?, ?, ?)
                 """,
@@ -841,7 +841,7 @@ class MagicFolderConfig(object):
         Retrieve a set of all relpaths of files that have had an entry in magic folder db
         (i.e. that have been downloaded at least once).
         """
-        cursor.execute("SELECT [path] FROM [local_snapshots]")
+        cursor.execute("SELECT [name] FROM [local_snapshots]")
         rows = cursor.fetchall()
         return set(r[0] for r in rows)
 
@@ -856,8 +856,8 @@ class MagicFolderConfig(object):
             relpath=path,
         )
         with action:
-            cursor.execute("DELETE FROM local_snapshots"
-                           " WHERE path=?",
+            cursor.execute("DELETE FROM [local_snapshots]"
+                           " WHERE [name]=?",
                            (path,))
 
     @with_cursor
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -798,7 +798,7 @@ class MagicFolderConfig(object):
                     [parent_identifier]
                 )
             VALUES
-                (?, ?, False, ?)
+                (?, ?, 0, ?)
             """,
             list(
                 (unicode(snapshot.identifier), index, parent_identifier)
@@ -826,7 +826,7 @@ class MagicFolderConfig(object):
                     [parent_identifier]
                 )
             VALUES
-                (?, ?, True, ?)
+                (?, ?, 1, ?)
             """,
             list(
                 (unicode(snapshot.identifier), index, unicode(parent.identifier))
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -389,11 +389,11 @@ class SQLite3DatabaseLocation(object):
         return sqlite3.connect(self.location, *a, **kw)
 
 
-def _get_snapshots(cursor, path):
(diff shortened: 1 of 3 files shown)


src/magic_folder/config.py
```diff
@@ -118,10 +126,60 @@
...
+        CREATE TABLE [local_snapshots]
         (
-            path          TEXT PRIMARY KEY,  -- the (mangled) name in UTF8
-            snapshot_blob BLOB               -- a JSON blob representing the snapshot instance
+            -- A random, unique identifier for this particular snapshot.
+            [identifier]       TEXT PRIMARY KEY,
+
+            -- The magicpath-mangled name of the file this snapshot is for,
+            -- UTF-8-encoded.
+            [path]             TEXT,
```
meejah, line 136: basically bike-shed-ing here, but maybe `name` would be better? (We use "name" everywhere else to describe the flattened snapshot name)

src/magic_folder/config.py
```diff
@@ -118,10 +126,60 @@
...
+            -- Offer marginal additional data integrity by requiring it to be
+            -- 0 or greater.
+            [index]                 INTEGER CHECK ([index] >= 0) NOT NULL,
+
+            -- If this parent is local only then its identifier is only
+            -- meaningful for looking up local snapshots in our local
+            -- database.  If it is not only local then its identifier is a
+            -- capability refering to a remote snapshot which we may or may
+            -- not have a local cache of.
+            [local_only]            BOOL NOT NULL,
```
meejah, line 175: Just to capture some real-time comments from elsewhere, this table strikes me as trying to mash two things together: local and remote parents. It might be better to have them separate so there isn't this "bool" column that essentially everything using this will have to consult to figure out how to interpret the "id" column.

The actual work of splitting this could be done when we actually have remote AND local parents? Could be worth a comment here though.

src/magic_folder/config.py
```diff
@@ -321,6 +389,229 @@ def connect(self, *a, **kw):
...
+    """
+    return LocalSnapshot(
+        identifier=UUID(hex=identifier),
+        name=path,
+        author=author,
+        content_path=FilePath(content_paths[identifier]),
+        metadata=metadata.get(identifier, {}),
+        parents_remote=_get_remote_parents(identifier, parents),
+        parents_local=_get_local_parents(
+            identifier,
```
meejah, line 602: Per the comment about the "parents" table (above), since we're *already* splitting these up anyway why not just store them in their own tables? Not that "efficiency"/optimization really matters here, but: 50% of the queries! ;)

src/magic_folder/config.py
```diff
@@ -451,31 +750,90 @@ def store_local_snapshot(self, cursor, snapshot):
...
+            ),
+        )
+        # Associate all of the parents with it.  First remote, then local.
+        #
+        # XXX What is the relative ordering of parents_local and
+        # parents_remote?  It's not preserved in `LocalSnapshot`.  Since local
+        # parents turn into remote parents I guess I'll put remote parents
+        # first here.  That way as local snapshots disappear they'll disappear
+        # from the middle and appear at the end of the remote snapshots list,
+        # also in the middle?
```
meejah, line 790: I don't believe we should be depending upon the *order* of parents anywhere. In the vast majority of cases there will be just 1 parent .. so it'll flow from local to remote. Only when there are conflicts will there be 2+ parents .. but they don't have any "order".

src/magic_folder/config.py
```diff
@@ -321,6 +389,229 @@ def connect(self, *a, **kw):
...
+        parents_local=_get_local_parents(
+            identifier,
+            parents,
+            partial(
+                _construct_local_snapshot,
+                path=path,
+                author=author,
+                content_paths=content_paths,
+                metadata=metadata,
+                parents=parents,
```
meejah, line 610: Okay, so this *doesn't* recurse forever because .. we do the `get()` based on identifiers (above) ... right?

src/magic_folder/config.py
exarkun: Sure.  Consistency is good.  fwiw, "path" makes more sense to me than "name" but since it's name in a lot of other places I'll adjust the database to match that.

src/magic_folder/config.py
exarkun: Added a comment.  I agree it makes sense to defer implementation changes here until there's application code depending on this stuff and we can see what facilitates the higher level goals better.

src/magic_folder/config.py
exarkun: Okay.  This would have been good to figure out earlier in the process, I guess.  I preserved order in the database because the Python API preserves order and there was never a good argument articulated (that I saw) that order doesn't matter.

Since this code works, I'm going to leave it (because it's easier to do so at this point).  In the future, if order doesn't matter, we can get rid of all of the ordering support in the database layer *and* change the Python representation so it doesn't appear to represent an order (probably switch some lists to some sets).

src/magic_folder/config.py
exarkun: Right.  Eventually a recursive call will encounter a snapshot with no local parents and the function can complete.  As long as snapshots are in a DAG. :)

I don't know how long this code will live, either.  The higher-level application code needs to get dis-entangled from the recursive `LocalSnapshot` representation now.  Maybe it will still be useful to have code to construct the full tree?  Or maybe not.  I don't know.

PR #275, opened 2020-08-27
Add a "fast" Hypothesis setting

### LINKED ISSUE #274: Use hypothesis profile to speed up tests
Currently, the unit-test suite takes close to 4minutes to run on my laptop. Using hypothesis settings to allow it just a single example per test brings this down to about 1minute.

meejah:
fixes #274

### DIFF
--- a/src/magic_folder/test/__init__.py
+++ b/src/magic_folder/test/__init__.py
@@ -17,6 +17,7 @@ from sys import (
 
 from foolscap.logging.incident import IncidentQualifier
 
+
 class NonQualifier(IncidentQualifier, object):
     def check_event(self, ev):
         return False
@@ -49,6 +50,11 @@ def _configure_hypothesis():
     from hypothesis import (
         HealthCheck,
         settings,
+    )
+
+    settings.register_profile(
+        "fast",
+        max_examples=1,
     )
 
     settings.register_profile(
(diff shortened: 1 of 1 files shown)


src/magic_folder/test/__init__.py
```diff
@@ -51,6 +52,11 @@ def _configure_hypothesis():
         settings,
     )
 
+    settings.register_profile(
+        "fast",
```
exarkun, line 56: It should be `magic-folder-fast` I think, for the same reason we have `magic-folder-ci` below.  Profile names aren't namespaced in any other way and Hypothesis allows profile name collisions to pass silently, then more or less randomly picks one definition to use when you try to activate the name.

I don't know of any other "fast" profile that's defined but considering how confusing and difficult to debug the situation is I think it's worth being defensive about it.

Maybe I should have written that as a comment above the `"magic-folder-ci"` profile below...

meejah: ahh, okay, thanks. I added (some of) the above as a comment too

meejah:
times now (running under coverage on my laptop which has 4 cores):

normal profile: 4:01
fast profile: 2:05
normal profile with -j4: 1:45
fast profile with -j4: 0:45

fast profile with -j4 and no coverage: 0:44

PR #277, opened 2020-08-27
265.local to remote snapshot

### LINKED ISSUE #265: Process local snapshots into remote snapshots as part of normal operation
Local snapshots might exist sometimes.  We want them to be remote snapshots.  There is code to create a remote snapshot from a local snapshot.

What is missing is some code to glue the local-to-remote uploader in to the overall magic folder service so that the uploader actually runs.

Implement that.

vu3rdd:
Fixes #265

### DIFF
--- a/src/magic_folder/tahoe_client.py
+++ b/src/magic_folder/tahoe_client.py
@@ -20,6 +20,9 @@ from hyperlink import (
 from treq.client import (
     HTTPClient,
 )
+from treq.testing import (
+    StubTreq,
+)
 
 import attr
 
@@ -125,7 +128,14 @@ class TahoeClient(object):
     """
 
     url = attr.ib(validator=attr.validators.instance_of(DecodedURL))
-    http_client = attr.ib(validator=attr.validators.instance_of(HTTPClient))
+
+    # treq should provider an interface but it doesn't ...  HTTPClient and
+    # StubTreq are both the same kind of thing.  HTTPClient is the one that
+    # does real networking, StubTreq is the one that operates in-memory on a
+    # local resource object.
+    http_client = attr.ib(
+        validator=attr.validators.instance_of((HTTPClient, StubTreq)),
+    )
 
     @inlineCallbacks
     def create_immutable_directory(self, directory_data):
--- a/src/magic_folder/tahoe_client.py
+++ b/src/magic_folder/tahoe_client.py
@@ -129,7 +129,7 @@ class TahoeClient(object):
 
     url = attr.ib(validator=attr.validators.instance_of(DecodedURL))
 
-    # treq should provider an interface but it doesn't ...  HTTPClient and
+    # treq should provide an interface but it doesn't ...  HTTPClient and
     # StubTreq are both the same kind of thing.  HTTPClient is the one that
     # does real networking, StubTreq is the one that operates in-memory on a
     # local resource object.
--- a/src/magic_folder/test/test_magic_folder_service.py
+++ b/src/magic_folder/test/test_magic_folder_service.py
@@ -206,12 +206,9 @@ class MagicFolderFromConfigTests(SyncTestCase):
         relative_paths(),
         relative_paths(),
         just(LOCAL_AUTHOR),
-        one_of(
-            tahoe_lafs_immutable_dir_capabilities(),
-            tahoe_lafs_dir_capabilities(),
-        ),
         tahoe_lafs_dir_capabilities(),
-        integers(min_value=1),
+        tahoe_lafs_dir_capabilities(),
+        integers(min_value=1, max_value=10000),
     )
     def test_uploader_service(self, name, relative_magic_path, relative_state_path, author, collective_dircap, upload_dircap, poll_interval):
         """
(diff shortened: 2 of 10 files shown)


src/magic_folder/tahoe_client.py
```diff
@@ -125,7 +128,14 @@ class TahoeClient(object):
     """
 
     url = attr.ib(validator=attr.validators.instance_of(DecodedURL))
-    http_client = attr.ib(validator=attr.validators.instance_of(HTTPClient))
+
+    # treq should provide an interface but it doesn't ...  HTTPClient and
+    # StubTreq are both the same kind of thing.  HTTPClient is the one that
+    # does real networking, StubTreq is the one that operates in-memory on a
+    # local resource object.
```
meejah, line 135: Well, so we also (now?) have an `HTTPClient` that does the in-memory stuff too via `magic_folder.client.create_testing_http_client`. I'm not suggesting immediate changes here -- just noting that. So e.g. we could standardize around "HTTPClient" but still do in-memory testing.

src/magic_folder/test/test_magic_folder_service.py
```diff
@@ -35,15 +44,34 @@
...
     relative_paths,
+    path_segments,
+    local_authors,
+    tahoe_lafs_dir_capabilities,
+    tahoe_lafs_readonly_dir_capabilities,
+    folder_names,
 )
 from .test_local_snapshot import (
-    MemorySnapshotCreator,
+    MemorySnapshotCreator as LocalMemorySnapshotCreator,
```
meejah, line 74: Is there a need for this aliasing? If `MemorySnapshotCreator` doesn't conflict, it might be easier to search etc if the name stays the same..

src/magic_folder/test/test_magic_folder_service.py
```diff
@@ -117,6 +148,182 @@ def test_create_local_snapshot(self, relative_target_path, content):
...
+    @given(
+        folder_names(),
+        relative_paths(),
+        path_segments(),
+        relative_paths(),
+        just(LOCAL_AUTHOR),
+        one_of(
+            tahoe_lafs_dir_capabilities(),
+            tahoe_lafs_readonly_dir_capabilities(),
+        ),
```
meejah, line 207: nice :)

src/magic_folder/test/test_magic_folder_service.py
```diff
@@ -117,6 +148,182 @@ def test_create_local_snapshot(self, relative_target_path, content):
...
+        reactor = task.Clock()
+
+        root = create_fake_tahoe_root()
+        http_client = create_tahoe_treq_client(root)
+        tahoe_client = TahoeClient(
+            DecodedURL.from_text(U"http://example.invalid./"),
+            http_client,
+        )
+
+        root._uri.data[upload_dircap] = json.dumps([
```
meejah, line 237: `_FakeTahoeRoot` has an `add_data` method which is probably better than twiddling internals like this directly.

I guess here you're doing something a bit different, though, which is adding data at an explicit key. The `add_data` API returns you the key (like Tahoe would). So a way around that would be to get rid of the Hypothesis generator for the upload-dircap and just generate it in the test like `_, upload_dircap = root.add_data("URI:DIR2-CHK", json.dumps(...))` (approximately).

src/magic_folder/test/test_magic_folder_service.py
```diff
@@ -117,6 +148,182 @@ def test_create_local_snapshot(self, relative_target_path, content):
...
+        )
+
+        self.assertThat(
+            d,
+            succeeded(Always()),
+        )
+
+        def children():
+            return json.loads(root._uri.data[upload_dircap])[1][u"children"]
+
```
meejah, line 313: I guess *ideally* this would be confirmed via the API instead of internals .. that is, by calling something on `tahoe_client` like `TahoeClient.list_directory(upload_dircap`).

So I think that could be like `children_d = TahoeClient.list_directory(upload_dircap)` and then `self.assertThat(children_d, succeeded(AfterPreprocessing(result=Equals({})))` or similar?

src/magic_folder/test/test_magic_folder_service.py
vu3rdd: Yes, there is no need. I also had the `MemorySnapshotCreator` from `test_upload` module imported but it was removed in a later refactoring. But the above code remained.

src/magic_folder/tahoe_client.py
exarkun: It does seem like that should be possible, yea.  Actually, looking more closely, I don't even know what `StubTreq` exists.  Maybe it was a precursor to `RequestTraversalAgent`.

src/magic_folder/test/test_magic_folder_service.py
exarkun: Maybe ... on the other hand, this is a test for code that does uploads.  If there is a coherent model underlying the HTTP interface then I don't think there's a compelling reason to force the code that checks to see if the upload has happened properly to also go via HTTP.  It is possible to argue that `json.loads(root._uri.data[upload_dircap])[1][u"children"]` is not the *most* coherent underlying model though ...

PR #278, opened 2020-08-28
'magic-folder list' uses HTTP API instead

### LINKED ISSUE #207: `magic-folder list` should use the HTTP API instead of reaching into internal Magic-Folder state

meejah:
Fixes #207 

This makes "magic-folder list" use the HTTP API (instead of reading the config itself). To do so requires a "magic-folder HTTP API client", which is included in this (using the new "client api endpoint" config).

To facilitate testing, the configuration abstractions are changed to construct more things in-memory and to hook up an in-process HTTP treq client to the HTTP API. This piece introduces a "token provider" so that API tokens can be in-memory only as well.

Some of the integration tests/infrastructure needed a little adjustment too.

### DIFF
--- a/docs/interface.rst
+++ b/docs/interface.rst
@@ -28,16 +28,33 @@ the client should re-read the token from the filesystem to determine if the valu
 ``GET /v1/magic-folder``
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-This endpoint returns a list of all individual magic-folders managed by this daemon.
+This endpoint returns a dict of all individual magic-folders managed
+by this daemon. The keys of the dict are the folder name and the
+values are themselves dicts.
+
+You may include the query argument ``?include_secret_information=1`` to
+include values for each folder which should be kept hidden (and are
+not shown by default). These are: ``upload_dircap``,
+``collective_dircap``, and the ``signing_key`` inside ``author``.
 
 The response code **OK** and the **Content-Type** is ``application/json``.
-The response body follows the form of this example::
+The response body follows the form of this example (containing a single
+magic-folder named "documents")::
 
-  { "folders":
-    [ { "name": "Alice's music", "local-path": "/home/alice/Music" }
-    , { "name": "Secret docs", "local-path": /home/alice/secrets" }
-    ]
-  }
+    {
+        "documents": {
+            "name": "documents",
+            "author": {
+                "name": "alice",
+                "verify_key": "OY7FCVPCOJXDNHQLSDTTJFONTROMQQED5Q6K33T3NBGGQHKLV73Q===="
+            },
+            "poll_interval": 60,
+            "is_admin": true,
+            "stash_path": "/home/alice/.config/magic-folder/documents",
+            "magic_path": "/home/alice/Documents"
+        }
+    }
+
 
 ``GET /v1/snapshot``
 ~~~~~~~~~~~~~~~~~~~~
--- a/integration/test_list.py
+++ b/integration/test_list.py
@@ -1,4 +1,10 @@
 
+import json
+
+from twisted.python.filepath import (
+    FilePath,
+)
+
 import pytest_twisted
 
 from eliot import (
@@ -40,3 +46,41 @@ def test_list(request, reactor, temp_dir, introducer_furl, flog_gatherer):
     output = yield proto.done
     assert output.strip() == "No magic-folders"
 
+    magic_dir = FilePath(temp_dir).child("zelda-magic")
+    magic_dir.makedirs()
+
+    proto = util._CollectOutputProtocol()
+    util._magic_folder_runner(
+        proto, reactor, request,
+        [
+            "--config", zelda.magic_config_directory,
+            "add",
+            "--author", "laptop",
+            "--name", "workstuff",
+            magic_dir.path,
+        ],
+    )
+    output = yield proto.done
+
+    proto = util._CollectOutputProtocol()
+    util._magic_folder_runner(
+        proto, reactor, request,
+        [
+            "--config", zelda.magic_config_directory,
+            "list",
+            "--json",
+        ],
+    )
+    output = yield proto.done
+    data = json.loads(output)
+
+    assert list(data.keys()) == ["workstuff"]
+    assert data["workstuff"]["name"] == "workstuff"
+    assert int(data["workstuff"]["poll_interval"]) == 60
+    assert data["workstuff"]["magic_path"] == magic_dir.path
+    assert data["workstuff"]["is_admin"] == True
+
+    # make sure we didn't reveal secrets
+    assert "signing_key" not in data["workstuff"]["author"]
+    assert "upload_dircap" not in data["workstuff"]
+    assert "collective_dircap" not in data["workstuff"]
--- a/integration/test_list.py
+++ b/integration/test_list.py
@@ -1,0 +1,51 @@
+from json import (
+    loads,
+)
+
+import pytest_twisted
+
+from eliot import (
+    start_action,
+)
+
+import util
+
+# see "conftest.py" for the fixtures (e.g. "magic_folder")
+
+
+@pytest_twisted.inlineCallbacks
+def test_list(request, reactor, temp_dir, introducer_furl, flog_gatherer):
+    """
+    'magic-folder list' happy-path works
+    """
+
+    print("-"*80)
+    print("create zelda")
+    with start_action(action_type=u"integration:test_list:zelda", include_args=[], include_result=False):
+        zelda = yield util.MagicFolderEnabledNode.create(
+            reactor,
+            request,
+            temp_dir,
+            introducer_furl,
+            flog_gatherer,
+            name="zelda",
+            tahoe_web_port="tcp:9982:interface=localhost",
+            magic_folder_web_port="tcp:19982:interface=localhost",
+            storage=True,
+        )
+    print("zelda", zelda)
+
+    proto = util._CollectOutputProtocol()
+    util._magic_folder_runner(
+        proto, reactor, request,
+        [
+            "--config", zelda.magic_config_directory,
+            "list",
+        ],
+    )
+    output = yield proto.done
+    print("-" * 80)
+    print(output)
+    print("-" * 80)
+    assert output.strip() == "No magic-folders"
+
--- a/integration/test_list.py
+++ b/integration/test_list.py
@@ -19,8 +19,6 @@ def test_list(request, reactor, temp_dir, introducer_furl, flog_gatherer):
     'magic-folder list' happy-path works
     """
 
-    print("-"*80)
-    print("create zelda")
     with start_action(action_type=u"integration:test_list:zelda", include_args=[], include_result=False):
         zelda = yield util.MagicFolderEnabledNode.create(
             reactor,
@@ -33,7 +31,6 @@ def test_list(request, reactor, temp_dir, introducer_furl, flog_gatherer):
             magic_folder_web_port="tcp:19982:interface=localhost",
             storage=True,
         )
-    print("zelda", zelda)
 
     proto = util._CollectOutputProtocol()
     util._magic_folder_runner(
@@ -44,8 +41,5 @@ def test_list(request, reactor, temp_dir, introducer_furl, flog_gatherer):
         ],
     )
     output = yield proto.done
-    print("-" * 80)
-    print(output)
-    print("-" * 80)
     assert output.strip() == "No magic-folders"
 
--- a/integration/test_list.py
+++ b/integration/test_list.py
@@ -1,6 +1,3 @@
-from json import (
-    loads,
-)
 
 import pytest_twisted
 
--- a/integration/test_list.py
+++ b/integration/test_list.py
@@ -7,6 +7,9 @@ from twisted.python.filepath import (
 
 import pytest_twisted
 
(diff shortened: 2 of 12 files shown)


src/magic_folder/config.py
```diff
@@ -219,14 +243,76 @@ def create_global_configuration(basedir, api_endpoint_str, tahoe_node_directory,
...
+        token_provider=FilesystemTokenProvider(
+            basedir.child(b"api_token"),
+        )
     )
     # make sure we have an API token
     config.rotate_api_token()
     return config
 
 
+def create_testing_configuration(basedir, tahoe_node_directory, global_service):
```
exarkun, line 257: I think the need for this function goes away if `magic_folder_list` and others accept an http client to use as an argument.  The tests can then use a regular global config and `treq_for_folders` or similar to make an http client to supply.

src/magic_folder/config.py
```diff
@@ -590,28 +679,43 @@ def is_admin(self):
...
 
 
+class ITokenProvider(Interface):
+    """
+    How the configuration retrieves and sets API access tokens.
+    """
+
+    def get():
+        """
+        Retrieve the current token.
```
exarkun, line 689: Return type would be good to document here.

src/magic_folder/config.py
```diff
@@ -590,28 +679,43 @@ def is_admin(self):
         return not collective_dmd.is_readonly()
 
 
+class ITokenProvider(Interface):
```
exarkun, line 682: Nice.

src/magic_folder/config.py
```diff
@@ -590,28 +679,43 @@ def is_admin(self):
...
-    def api_token(self):
+    def get(self):
         """
-        Current API token
+        Retrieve the current token.
         """
         if self._api_token is None:
             with self.api_token_path.open('rb') as f:
                 self._api_token = f.read()
         return self._api_token
```
exarkun, line 714: Not really related to this branch but just to share - At some point I found it surprising that if I rewrote the token file then the implementation *didn't* respect the change due to the caching here.  I don't think rewriting the token file outside of the control of these classes is a common or even particularly good use case but there's no setter right now and it was a convenient thing to try at one point (until I realized it didn't work).

src/magic_folder/config.py
```diff
@@ -620,6 +724,56 @@ def rotate_api_token(self):
...
+    Keep an in-memory token.
+    """
+    _api_token = attr.ib(default=None)
+
+    def get(self):
+        """
+        Retrieve the current token.
+        """
+        if self._api_token is None:
+            self.rotate()
```
exarkun, line 741: What does the other implementation do in this case?  My guess is it raises OSError.  Maybe it should auto-rotate too?

src/magic_folder/list.py
```diff
@@ -0,0 +1,76 @@
...
+    :param output: a file-like object to which the output will be written
+
+    :param bool as_json: return answer in JSON
+
+    :param bool include_secret_information: include sensitive private
+        information (such as long-term keys) if True (default: False).
+
+    :return: JSON response from `GET /v1/magic-folder`.
+    """
+    from twisted.internet import reactor
```
exarkun, line 35: I'd suggest reactor as a parameter.  Let the CLI entrypoint import the global reactor and require everything else to receive it as a parameter.

src/magic_folder/test/cli/common.py
```diff
@@ -63,7 +63,6 @@ def cli(config_directory, argv):
     try:
         try:
             options.parseOptions([
-                b"--debug",
```
exarkun, line 66: Does this mean we're going to get less debug output when a cli test fails?  Is this change necessary?

src/magic_folder/test/cli/test_magic_folder.py
```diff
@@ -70,98 +80,111 @@ def setUp(self):
...
     @defer.inlineCallbacks
     def test_list_none(self):
         """
         When there are no Magic Folders at all, the output of the list command
         reports this.
         """
-        outcome = yield cli(
-            self.config_dir,
-            [b"list"],
+        output = StringIO()
```
exarkun, line 113: This is supposed to be like sys.stdout right?  `BytesIO` might be more realistic.  Although if this is just mirroring what `cli` was doing then nevermind, let's not blow up the scope of this PR.

src/magic_folder/test/cli/test_magic_folder.py
```diff
@@ -70,98 +80,111 @@ def setUp(self):
...
-                b"add",
-                b"--name", b"list-some-folder",
-                b"--author", b"alice",
-                folder_path.asBytesMode().path,
-            ],
-        )
-        self.assertThat(
-            outcome.succeeded(),
-            Equals(True),
+        self.config.create_magic_folder(
```
exarkun, line 142: much nicer

src/magic_folder/test/cli/test_magic_folder.py
```diff
@@ -70,98 +80,111 @@ def setUp(self):
...
 
-        outcome = yield cli(
-            self.config_dir,
-            [b"list"],
-        )
-        self.expectThat(outcome.stdout, Contains(b"list-some-folder"))
-        self.expectThat(outcome.stdout, Contains(folder_path.path))
+        output = StringIO()
+        yield magic_folder_list(self.config, output)
+        self.expectThat(output.getvalue(), Contains(b"list-some-folder"))
```
exarkun, line 154: I think looking for `b"..."` in a `StringIO` only works until there's any non-ASCII anywhere.  I'm inclined to suggest `BytesIO` here but same note as above about not unnecessarily expanding this PR's scope...

src/magic_folder/test/test_web.py
```diff
@@ -407,17 +414,25 @@ def test_list_folders(self, folders):
...
+            AUTH_TOKEN,
+            {
                 name: magic_folder_config(self.author, FilePath(self.mktemp()), path_u)
                 for (name, path_u)
                 in folders.items()
             },
             False,
         )
 
+        config = load_global_configuration(basedir)
```
exarkun, line 430: It took me a minute to realize there is a global configuration at that path as a side-effect of `treq_for_folders`.  Maybe worth a comment here to that effect.

src/magic_folder/test/test_web.py
```diff
@@ -429,13 +444,21 @@ def test_list_folders(self, folders):
                     body_matcher=AfterPreprocessing(
                         loads,
                         Equals({
-                            u"folders": list(
-                                {u"name": name, u"local-path": path_u.path}
-                                for (name, path_u)
-                                in sorted(folders.items())
-                            ),
+                            name: {
+                                u"name": name,
```
exarkun, line 448: Reading these two lines, it occurs to me that maybe the name property is now redundant?

src/magic_folder/web.py
```diff
@@ -313,17 +317,35 @@ def render_GET(self, request):
...
+                },
+                u"stash_path": mf.stash_path.path,
+                u"magic_path": mf.magic_path.path,
+                u"poll_interval": mf.poll_interval,
+                u"is_admin": mf.is_admin(),
+            }
+            if include_secret_information:
+                info[u"author"][u"signing_key"] = mf.author.signing_key.encode(Base32Encoder)
+                info[u"collective_dircap"] = mf.collective_dircap.encode("ascii")
+                info[u"upload_dircap"] = mf.upload_dircap.encode("ascii")
```
exarkun, line 338: I think JSON is happier with unicode than bytes ... although since caps are always ascii I don't know how much it matters here.  Still, there's no "bytes" type in JSON.

docs/interface.rst
```diff
@@ -28,16 +28,33 @@ the client should re-read the token from the filesystem to determine if the valu
...
+        "documents": {
+            "name": "documents",
+            "author": {
+                "name": "alice",
+                "verify_key": "OY7FCVPCOJXDNHQLSDTTJFONTROMQQED5Q6K33T3NBGGQHKLV73Q===="
+            },
+            "poll_interval": 60,
+            "is_admin": true,
+            "stash_path": "/home/alice/.config/magic-folder/documents",
+            "magic_path": "/home/alice/Documents"
```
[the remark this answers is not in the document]
meejah: okay, I filed https://github.com/LeastAuthority/magic-folder/issues/279 because there are a lot of `under_score`-using ones that aren't in this PR

integration/test_list.py
```diff
@@ -0,0 +1,42 @@
...
+# see "conftest.py" for the fixtures (e.g. "magic_folder")
+
+
+@pytest_twisted.inlineCallbacks
+def test_list(request, reactor, temp_dir, introducer_furl, flog_gatherer):
+    """
+    'magic-folder list' happy-path works
+    """
+
+    with start_action(action_type=u"integration:test_list:zelda", include_args=[], include_result=False):
```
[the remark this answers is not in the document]
meejah: The pytest decorator just marks the function .. so it seems like both decorators can be used. Are there symptoms of the "not safe" part that I can see to confirm it's behaving safely?

integration/util.py
```diff
@@ -23,7 +26,7 @@
     ProcessDone,
 )
 
-import requests
+import treq
```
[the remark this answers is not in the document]
meejah: The previous code was wrong; it was doing synchronous waiting for the API to become ready. However, that only works in pytest fixtures (which use `pytest_twisted.blockon` to await any deferreds synchronously) .. but methods in `util` can also be used by tests which are (or can be) just "normal" async functions. So, the `await_client_ready` call would only happen to work when called inside fixtures not "in general".

integration/util.py
```diff
@@ -593,87 +596,74 @@ def _check_status(response):
...
-            time.sleep(1)
-            continue
 
         # check that all times are 'recent enough'
-        if any([time.time() - t > liveness for t in server_times]):
-            print("waiting because at least one server too old")
-            time.sleep(1)
+        if all([time.time() - t > liveness for t in server_times]):
+            print("waiting because no server new enough")
+            yield twisted_sleep(reactor, 1)
```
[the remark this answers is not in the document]
meejah: Because `happy=1`. So, since we're using a single Introducer .. any client that is started (with storage) but then stopped will no longer report a connection here at all (i.e. their time will be `None`). Also the behavior described in the docstring wasn't what this was actually doing.

So basically a client is "ready" if there are enough recent storage-servers so that a Tahoe operation will succeed. In this case, 1.

src/magic_folder/client.py
```diff
@@ -0,0 +1,137 @@
...
+
+    :ivar callable get_api_token: returns the current API token
+    """
+
+    # we only use the path-part not the domain
+    base_url = DecodedURL.from_text(u"http://invalid./")
+    http_client = attr.ib(validator=attr.validators.instance_of(HTTPClient))
+    get_api_token = attr.ib()
+
+    def list_folders(self, include_secret_information=None):
```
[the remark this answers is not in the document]
meejah: Well, I wanted to be able to distinguish in case we actually wanted to send the query-arg with "=0" whereas `None` means to not include the query-arg at all. Although, there isn't a code-path that includes `=0` so I guess False or None is  kind-of equivalent here as it sits.

Also, "in general" I usually always use `None` for "I want the default" in kwargs .. probably because "sometimes" (i.e. mutable defaults) it's wrong to put the "actual" default there.

src/magic_folder/test/cli/common.py
meejah: Well, so the output you do get with `--debug` turned on isn't very useful -- that's why I changed `cli.run_magic_folder_options` to let the exception "escape". Without *that* change you just always get a traceback to the `maybeDeferred` call .. not to the actual problem.

So, but then instead the CLI runner will get "real" exceptions out instead of printing what the user will see. So I figured since this is all in the "cli testing" helper code the tests should really be expecting to see what the user will. Thinking now, it's certainly possible to have this helper catch the exception and dump the stack "just like python might have done" ... but also none of our tests currently look for printed-out-exceptions from `--debug`.

integration/test_list.py
```diff
@@ -0,0 +1,42 @@
...
+    proto = util._CollectOutputProtocol()
+    util._magic_folder_runner(
+        proto, reactor, request,
+        [
+            "--config", zelda.magic_config_directory,
+            "list",
+        ],
+    )
+    output = yield proto.done
+    assert output.strip() == "No magic-folders"
```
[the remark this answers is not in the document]
meejah: I kept this in because it's what I wrote first (and it fixed some other broken stuff in the integration tests). Can we keep both kinds of tests? (There are unit-tests as well).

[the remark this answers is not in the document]
exarkun: I don't think this kind of test adds a lot of value compared to its cost.  It's fragile (it encodes a machine dependency on a human-facing piece of english prose) and slow to run (14 seconds for this single test on my laptop).  It may also be hard to debug (often these tells tell you just one thing "it worked" or "it didn't" and you're left to try to figure out why;  I think because of the `--debug`/traceback change this one might sometimes give you more, though I don't think that's necessarily the case - it depends where the problem occurs).  If there's a clear path to a testing strategy without these drawbacks we should take it.

PR #282, opened 2020-08-31
Magic UTF-8 dust cleanup

### LINKED ISSUE #281: There is much repetition of the policy to interact with the platform filesystem using UTF-8
Throughout the codebase we now have dozens of expressions like `filepath.asBytesMode("utf-8")` and `filemode.asTextMode("utf-8")`.  "utf-8" here is a policy decision (based on the premise that it can encode everything and all filesystems that matter will accept UTF-8 and round-trip it more or less correctly back to us).

Unfortunately:
  * It's easy to forget this sometimes.  Either because it has to happen *everywhere* filepath and a filesystem interact or because those interactions aren't always obvious/local.
  * It makes every filesystem interaction harder.
  * It bloats the code.
  * It increasingly builds up a barrier to making any significant improvements to our filesystem interaction code by spreading the UTF-8 decision far and wide.

Fortunately:
  * FilePath has built-in support for automatically selecting an encoding.
  * There is a single chokepoint for this: `sys.getfilesystemencoding`.

Furthermore, on some release of Python 3, the "look at locale settings" behavior of `sys.getfilesystemencoding` which causes it to return "ascii" sometimes has been removed and it simply *always* returns "UTF-8".

We can backport this behavior to magic-folder processes with a monkey-patch and then stop sprinkling magic utf-8 dust over the entire rest of the codebase.

exarkun:
Fixes #281

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -177,7 +177,6 @@ jobs:
     environment:
       <<: *UTF_8_ENVIRONMENT
       MAGIC_FOLDER_TOX_ENVIRONMENT: "pypy27-coverage"
-      ALLOWED_FAILURE: "yes"
 
 
   c-locale:
--- a/setup.py
+++ b/setup.py
@@ -38,6 +38,12 @@ install_requires = [
 
     "importlib_metadata",
 
+    # We use this to "fix" sys.getfilesystemencoding().  This is also an
+    # indirect requirement through python-cryptography so we're not making our
+    # dependency list any heavier by this approach.  We list this here though
+    # to be clear that we also directly depend on it, though.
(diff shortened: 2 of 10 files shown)


exarkun:
None of the strategies I have explored can simultaneously support all of:

  * "Regular" CPython builds
  * Debian-style builds
  * Any released PyPy
  * `magic-folder ...`-style usage
  * `import magic_folder`-style usage
  * `python -m ...`-style usage

And I'm out of ideas for more strategies.  It's possible a lot more digging and special-casing could make one of the strategies finally succeed on all our of tested platforms but it doesn't seem worth the effort anymore.

Issue #287, opened 2020-08-31
Explicit validators


meejah:
Per https://github.com/LeastAuthority/magic-folder/pull/257#discussion_r480242001 we end up using config setters to get the validation side-effect.

It also means we're doing an SQL `INSERT` and then soon after an `UPDATE` of the same information.

One way around this would be to have (separate, internal) validator methods that the setters AND the initializer calls. Perhaps there is some other way to achieve the same effect.

PR #288, opened 2020-08-31
move validators to helper-functions

meejah:
fixes https://github.com/LeastAuthority/magic-folder/issues/287


src/magic_folder/config.py
```diff
@@ -1369,3 +1357,25 @@ def create_magic_folder(self, name, magic_path, state_path, author,
...
+
+
+def _validate_listen_endpoint_str(ep_string):
+    """
+    confirm we have a valid endpoint-string
+    """
+    from twisted.internet import reactor  # uhm...
+    # XXX so, having the reactor here sucks. But if we pass in an
+    # IStreamServerEndpoint instead, how can we turn that back
+    # into an endpoint-string?
```
exarkun, line 1369: not sure this comment still makes sense in this context, in particular about turning it back into a string, since this function is just concerned with validating the string.

PR #290, opened 2020-09-01
Configure flake8 with pre-commit and fix some errors

### LINKED ISSUE #289: Add "pre-commit" flake8 hook
https://pre-commit.com/ is a project to glue checks into the git pre-commit hook system.  It is a handy way to do things like run flake8 prior to every commit.  The motivation here is similar to the motivation for systems like emacs' flycheck - learn about errors early.  The idea is not to replace CI but to learn about problems that CI will eventually find but only after much more time has passed.

exarkun:
Fixes #289

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,4 @@
+- repo: "https://github.com/PyCQA/flake8"
+  rev: "3.8.3"
+  hooks:
+  - id: "flake8"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,9 @@
-- repo: "https://github.com/PyCQA/flake8"
-  rev: "3.8.3"
-  hooks:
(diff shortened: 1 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,4 @@
+- repo: "https://github.com/PyCQA/flake8"
+  rev: "3.8.3"
+  hooks:
+  - id: "flake8"
```
meejah, line 4: I think this should say `stage: push` or similar .. I don't know about others, but I will often have commits that aren't "ready" in the sense of "production-quality code" or so but are still "a thing I want to commit".

Or, maybe I'm wrong about what a commit-hook does, but I think the above says "you can't make a commit unless flake8 passes".

exarkun:
After some practical experience with pre-commit, on Tahoe-LAFS I'm not so interested in this anymore.

PR #294, opened 2020-10-07
CircleCI vs Docker Hub

### LINKED ISSUE #293: New Docker Hub pull limits will likely negatively impact CircleCI jobs
Docker Inc. is introducing limits on unauthenticated image pulls from Docker Hub - https://docs.docker.com/docker-hub/download-rate-limit/

Our CircleCI configuration uses Docker images for most jobs.  CircleCI pulls these images from Docker Hub, unauthenticated.  The result is that our jobs will share the unauthenticated pull limits with all other CircleCI users and no doubt rapidly be denied service, causing jobs failures.

exarkun:
Fixes #293

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,6 +118,8 @@ jobs:
           username: $DOCKERHUB_USERNAME
           password: $DOCKERHUB_PASSWORD
 
+    steps: []
+
   lint:
     docker:
       - <<: *DOCKERHUB_AUTH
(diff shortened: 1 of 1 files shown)


exarkun:
It's hard to tell what difference this is making now.  CircleCI output doesn't seem to give any indication of whether the credentials are being used or not.  The docs *say* they'll be used ... so I guess they are?  And the pulls are succeeding...

I also don't see anything on the Docker Hub side that would tell us if we're using the credentials or not.

exarkun:
So I guess I'll merge this  for now on the premise that it does what the CircleCI docs say we should do.  Hopefully that means we won't ever hit the problem we're trying to avoid.  If we do, I guess we have to revisit.

PR #310, opened 2021-04-29
309.add file

### LINKED ISSUE #309: "add new snapshot" CLI command
In order to exercise the "create a Snapshot" HTTP API it would be good to have a CLI subcommand to trigger this.

It might be worth considering how to lay out "plumbing" versus "porcelain" commands in the future; this one is more "plumbing" than anything.

meejah:
fixes #309 

Probably still needs:
- [x] tests
- [x] merge #311

Note that this is based on #311 so that needs to be reviewed / merged first.

### DIFF
--- a/newsfragments/309.feature
+++ b/newsfragments/309.feature
@@ -1,0 +1,1 @@
+Add a "magic-folder-api add-snapshot" command
--- a/setup.py
+++ b/setup.py
@@ -314,6 +314,7 @@ setup(name="magic_folder",
       entry_points = {
           "console_scripts": [
               "magic-folder = magic_folder.cli:_entry",
+              "magic-folder-api = magic_folder.api_cli:_entry",
           ],
       },
       **setup_args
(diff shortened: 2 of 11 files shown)


meejah:
Perhaps it makes sense to have a completely separate `magic-folder-api` command or something where we can just drop in straight-up convenience methods for using the API (i.e. instead of messing with `curl`). That is, it just calls the API and prints the result. This is better than `curl` because: you don't have to figure out the URL (just the config-directory); you don't have to load the api-token; you don't have to use bash to construct a syntactically-valid `Authorization` header; it has relevant documentation.

Obviously, some of the API endpoints are used for other commands etc, but that would leave `magic-folder *` a bit more "pure" (i.e. straight porcelain).

meejah:
Discussed last week and consensus was reached that having a `magic-folder-api` (well, name could be different, but "separate command tree") for direct wrappers of API endpoints is good. This is for testing or other lower-level access. Commands in this tree shouldn't do anything "smart" -- they're essentially just `curl` except you don't have to find the token, build headers, etc.

I'm going to switch this command to be the first such command. There may be a smarter (i.e. more generic) way to write these once we add more.

PR #321, opened 2021-05-05
unifiy tahoe-clients

### LINKED ISSUE #311: unify tahoe clients
There are kind of two similar "tahoe client" implementations. They should be unified (i.e. missing parts from the one added to the other).

One is `magic_folder.cli.TahoeClient` and the other is in `magic_folder.tahoe_client`

The one in `cli` also depends on a `Node` object that presents (some of) the tahoe-lafs legacy `Node` API. I assume this was a way to help port magic-folder out of tahoe -- however the "Node" object doesn't make much sense for stand-alone magic-folder per se.

meejah:
Fixes #311 

Merges the two tahoe-clients into one and adds missing tests.


meejah:
Merging this after self-review as I need it for other in-progress PRs and it's 6 days old already. I'm still happy to receive a "hindsight" review though!

Issue #331, opened 2021-05-11
Participants don't track verify-keys


meejah:
Currently, participants in the Collective DMD aren't associated with their public-key.

We could use tahoe's `metadata` for this:
- use `POST /uri/<collective-dmd>?t=set-children` with an appropriate body
- set `metadata` to e.g. `{"verify_key": <base32>}`
- get the `verify_key` back of out the metadata on download/scan.

This would give all participants a way to verify signatures (of Snapshots). This might give the folder-admin some slight ambient authority: they could lie and put in their own keypair, but I think all they could accomplish is that this client's Snapshots would all be rejected.

PR #332, opened 2021-05-19
327.add participant

### LINKED ISSUE #327: "add participant" HTTP API
This isn't quite "invites", but can be used as a building-block for that (or for part of it).

Adding a participant can also be used for testing (either manual or automated) to set up known participants without having to use magic-wormhole. Having multiple participants is necessary to test downloader. This API  can also be used by a "recovery" workflow that wishes to syncrhonize from a previous device's folder -- the old folder is simply added as a participant and synchronization proceeds as normal.

meejah:
Fixes #327 

"add participant" HTTP API, associated support and tests.

### DIFF
--- a/newsfragments/327.feature
+++ b/newsfragments/327.feature
@@ -1,0 +1,1 @@
+There is now an HTTP API to add a new participant
--- a/newsfragments/327.feature
+++ b/newsfragments/327.feature
@@ -1,1 +1,1 @@
-There is now an HTTP API to add a new participant
+There is now an HTTP API to add and list new participants, along with magic-folder-api subcommands
--- a/newsfragments/327.feature
+++ b/newsfragments/327.feature
@@ -1,1 +1,1 @@
-There is now an HTTP API to add and list new participants, along with magic-folder-api subcommands
+There is now an HTTP API to add and list new participants, along with corresponding magic-folder-api subcommands
--- a/src/magic_folder/api_cli.py
+++ b/src/magic_folder/api_cli.py
@@ -101,15 +101,15 @@ class AddParticipantOptions(usage.Options):
     optParameters = [
         ("folder", "n", None, "Name of the magic-folder to add a participant to"),
         ("author-name", "a", None, "Name of the new participant"),
-        ("author-verify-key", "k", None, "Base32-encoded verify key of the new participant"),
         ("personal-dmd", "p", None, "Read-capability of the participant's Personal DMD"),
+        # not yet
+        # ("author-verify-key", "k", None, "Base32-encoded verify key of the new participant"),
     ]
 
     def postOptions(self):
         required_args = [
             ("folder", "--folder / -n is required"),
             ("author-name", "--author-name / -a is required"),
-            ("author-verify-key", "--author-verify-key / -k is required"),
             ("personal-dmd", "--personal-dmd / -p is required"),
         ]
         for (arg, error) in required_args:
@@ -125,7 +125,6 @@ def add_participant(options):
     res = yield options.parent.get_client().add_participant(
         options['folder'].decode("utf8"),
         options['author-name'].decode("utf8"),
-        options['author-verify-key'].decode("utf8"),
         options['personal-dmd'].decode("utf8"),
     )
     print("{}".format(res), file=options.stdout)
--- a/src/magic_folder/api_cli.py
+++ b/src/magic_folder/api_cli.py
@@ -2,6 +2,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import sys
+import json
 
 from twisted.internet.task import (
     react,
@@ -96,6 +97,65 @@ def dump_state(options):
         print("    {}: {}".format(snap_name, snap), file=options.stdout)
 
 
+class AddParticipantOptions(usage.Options):
+    optParameters = [
+        ("folder", "n", None, "Name of the magic-folder to add a participant to"),
+        ("author-name", "a", None, "Name of the new participant"),
+        ("author-verify-key", "k", None, "Base32-encoded verify key of the new participant"),
+        ("personal-dmd", "p", None, "Read-capability of the participant's Personal DMD"),
+    ]
+
+    def postOptions(self):
+        required_args = [
+            ("folder", "--folder / -n is required"),
+            ("author-name", "--author-name / -a is required"),
+            ("author-verify-key", "--author-verify-key / -k is required"),
+            ("personal-dmd", "--personal-dmd / -p is required"),
+        ]
+        for (arg, error) in required_args:
+            if self[arg] is None:
+                raise usage.UsageError(error)
+
+
+@inlineCallbacks
+def add_participant(options):
+    """
+    Add one new participant to an existing magic-folder
+    """
+    res = yield options.parent.get_client().add_participant(
+        options['folder'].decode("utf8"),
+        options['author-name'].decode("utf8"),
+        options['author-verify-key'].decode("utf8"),
+        options['personal-dmd'].decode("utf8"),
+    )
+    print("{}".format(res), file=options.stdout)
+
+
+class ListParticipantsOptions(usage.Options):
+    optParameters = [
+        ("folder", "n", None, "Name of the magic-folder participants to list"),
+    ]
+
+    def postOptions(self):
+        required_args = [
+            ("folder", "--folder / -n is required"),
+        ]
+        for (arg, error) in required_args:
+            if self[arg] is None:
+                raise usage.UsageError(error)
+
+
+@inlineCallbacks
+def list_participants(options):
+    """
+    List all participants in a magic-folder
+    """
+    res = yield options.parent.get_client().list_participants(
+        options['folder'].decode("utf8"),
+    )
+    print("{}".format(json.dumps(res, indent=4)), file=options.stdout)
+
+
 class MagicFolderApiCommand(usage.Options):
     """
     top-level command (entry-point is "magic-folder-api")
@@ -150,6 +210,8 @@ class MagicFolderApiCommand(usage.Options):
     subCommands = [
         ["add-snapshot", None, AddSnapshotOptions, "Add a Snapshot of a file to a magic-folder."],
         ["dump-state", None, DumpStateOptions, "Dump the local state of a magic-folder."],
+        ["add-participant", None, AddParticipantOptions, "Add a Participant to a magic-folder."],
+        ["list-participants", None, ListParticipantsOptions, "List all Participants in a magic-folder."],
     ]
     optFlags = [
         ["debug", "d", "Print full stack-traces"],
@@ -248,6 +310,8 @@ def run_magic_folder_api_options(options):
     main_func = {
         "add-snapshot": add_snapshot,
         "dump-state": dump_state,
+        "add-participant": add_participant,
+        "list-participants": list_participants,
     }[options.subCommand]
 
     # we want to let exceptions out to the top level if --debug is on
--- a/src/magic_folder/participants.py
+++ b/src/magic_folder/participants.py
@@ -158,8 +158,7 @@ class _CollectiveDirnodeParticipants(object):
         """
         uri = tahoe_uri_from_string(value)
         if IDirectoryURI.providedBy(uri):
-            if not uri.is_readonly():
-                return
+            return
         raise TypeError(
             "Upload dirnode was {!r}, must be a read-write directory node.".format(
                 value,
--- a/src/magic_folder/participants.py
+++ b/src/magic_folder/participants.py
@@ -32,6 +32,13 @@ from allmydata.uri import (
 from .magicpath import (
     magic2path,
 )
+from .snapshot import (
+    RemoteAuthor,
+)
+from .tahoe_client import (
+    CannotAddDirectoryEntryError,
+)
+
 
 class IParticipant(Interface):
     """
@@ -142,6 +149,57 @@ class _CollectiveDirnodeParticipants(object):
                 value,
             ),
         )
+
+    @inlineCallbacks
+    def add(self, author, personal_dmd_cap):
+        """
+        Add a new participant to this collective.
+
+        :param IRemoteAuthor author: personal details of new participant
+
+        :param bytes personal_dmd_cap: the read-capability of the new
+            participant (if it is a write-capability then it's "us"
+            but by definition we already have an "us" participant in
+            any existing Magic Folder and so that's an error here).
+
+        :returns IParticipant: the new participant
+        """
+        uri = tahoe_uri_from_string(personal_dmd_cap)
+        if not IDirnodeURI.providedBy(uri) or not uri.is_readonly():
+            raise ValueError(
+                "New participant Personal DMD must be read-only dircap"
+            )
+        if not isinstance(author, RemoteAuthor):
+            raise ValueError(
+                "Author must be a RemoteAuthor instance"
+            )
+        # semantically, it doesn't make sense to allow a second
+        # participant with the very same Personal DMD as another (even
+        # if the name/author is different). So, we check here .. but
+        # there is a window for race between the check and when we add
+        # the participant. The only real solution here would be a lock
+        # or to serialize all Tahoe operations.
+        participants = yield self.list()
+        if personal_dmd_cap in [p.dircap for p in participants]:
+            raise ValueError(
+                "Already have a participant with Personal DMD '{}'".format(personal_dmd_cap)
+            )
+
+        # NB: we could check here if there is already a participant
+        # for this name .. however, there's a race between that check
+        # succeeding and adding the participant so we just try to add
+        # and let Tahoe send us an error by using "replace=False"
+        try:
+            yield self._tahoe_client.add_entry_to_mutable_directory(
+                self._collective_cap,
+                author.name,
+                personal_dmd_cap.encode("ascii"),
+                replace=False,
+            )
+        except CannotAddDirectoryEntryError:
+            raise ValueError(
+                "Already have a participant called '{}'".format(author.name)
+            )
 
     @inlineCallbacks
     def list(self):
--- a/src/magic_folder/participants.py
+++ b/src/magic_folder/participants.py
@@ -73,6 +73,21 @@ class IParticipants(Interface):
             associated with..
         """
 
+    @inlineCallbacks
+    def add(author, personal_dmd_cap):
+        """
+        Add a new participant to this collective.
+
+        :param IRemoteAuthor author: personal details of new participant
+
+        :param bytes personal_dmd_cap: the read-capability of the new
+            participant (if it is a write-capability then it's "us"
+            but by definition we already have an "us" participant in
+            any existing Magic Folder and so that's an error here).
+
+        :returns IParticipant: the new participant
+        """
+
 
 def participant_from_dmd(name, dirnode, is_self, tahoe_client):
     """
@@ -153,16 +168,7 @@ class _CollectiveDirnodeParticipants(object):
     @inlineCallbacks
     def add(self, author, personal_dmd_cap):
         """
-        Add a new participant to this collective.
-
-        :param IRemoteAuthor author: personal details of new participant
-
-        :param bytes personal_dmd_cap: the read-capability of the new
-            participant (if it is a write-capability then it's "us"
-            but by definition we already have an "us" participant in
-            any existing Magic Folder and so that's an error here).
-
-        :returns IParticipant: the new participant
+        IParticipants API
         """
         uri = tahoe_uri_from_string(personal_dmd_cap)
         if not IDirnodeURI.providedBy(uri) or not uri.is_readonly():
@@ -177,8 +183,9 @@ class _CollectiveDirnodeParticipants(object):
         # participant with the very same Personal DMD as another (even
         # if the name/author is different). So, we check here .. but
         # there is a window for race between the check and when we add
-        # the participant. The only real solution here would be a lock
-        # or to serialize all Tahoe operations.
+        # the participant. The only real solution here would be a
+        # magic-folder-wide write-lock or to serialize all Tahoe
+        # operations (at least across one magic-folder).
         participants = yield self.list()
         if personal_dmd_cap in [p.dircap for p in participants]:
             raise ValueError(
--- a/src/magic_folder/participants.py
+++ b/src/magic_folder/participants.py
@@ -187,7 +187,7 @@ class _CollectiveDirnodeParticipants(object):
         # magic-folder-wide write-lock or to serialize all Tahoe
         # operations (at least across one magic-folder).
         participants = yield self.list()
-        if personal_dmd_cap in [p.dircap for p in participants]:
+        if any(personal_dmd_cap == p.dircap for p in participants):
             raise ValueError(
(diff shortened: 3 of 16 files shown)


src/magic_folder/web.py
```diff
@@ -114,11 +134,158 @@ class APIv1(Resource, object):
     """
     _global_config = attr.ib()
     _global_service = attr.ib()
+    _tahoe_client = attr.ib()
 
     def __attrs_post_init__(self):
         Resource.__init__(self)
         self.putChild(b"magic-folder", MagicFolderAPIv1(self._global_config))
         self.putChild(b"snapshot", SnapshotAPIv1(self._global_config, self._global_service))
+        self.putChild(b"participants", ParticipantsAPIv1(self._global_config, self._tahoe_client))
```
tomprince, line 143: nit: Given that we are adding a participant to a folder, I'd be inclined to put this endpoint as a child of the(a) folder endpoint. So something like `/v1/magic-folder/<folder-name>/participants`.

src/magic_folder/test/__init__.py
```diff
@@ -62,6 +62,11 @@ def _configure_hypothesis():
     settings.register_profile(
         "magic-folder-fast",
         max_examples=1,
+        # see magic-folder-ci profile below for justification
+        suppress_health_check=[
+            HealthCheck.too_slow,
+        ],
+        deadline=None,
```
tomprince, line 69: I'm not sure this makes sense?

src/magic_folder/test/cli/test_magic_folder.py
```diff
@@ -111,6 +125,7 @@ def get_folder_service(s, name):
             self.config,
             self.service,
             lambda: self.config.api_token,
+            create_tahoe_client(DecodedURL.from_text(u""), StubTreq(Resource())),
```
tomprince, line 128: Maybe have a comment on why `StubTreq` with an empty resource is sufficient here?

src/magic_folder/util/encoding.py
```diff
@@ -0,0 +1,4 @@
+import unicodedata
```
tomprince, line 1: My fault but this file should probably have:
- copyright header
- docstring (module and for function0
- `from __future__ import (unicode_literals, print_function, absolute_import, division)`

src/magic_folder/web.py
```diff
@@ -114,11 +134,158 @@ class APIv1(Resource, object):
...
+
+    def __attrs_post_init__(self):
+        Resource.__init__(self)
+
+    def getChild(self, name, request):
+        name_u = name.decode("utf-8")
+        try:
+            folder_config = self._global_config.get_magic_folder(name_u)
+        except ValueError:
+            return NoResource(b"{}")
```
tomprince, line 166: I suspect we'll want something smarter than this, but we can file an issue and handle it in a follow up.

src/magic_folder/web.py
```diff
@@ -114,11 +134,158 @@ class APIv1(Resource, object):
...
+            if set(participant["author"].keys()) != required_author_keys:
+                raise ValueError("'author' requires: {}".format(", ".join(sorted(required_author_keys))))
+
+            author = create_author(
+                participant["author"]["name"],
+                # we don't yet properly track keys but need one
+                # here .. this won't be correct, but we won't use
+                # it .. following code still only looks at the
+                # .name attribute
+                VerifyKey(os.urandom(32)),
```
tomprince, line 249: Can you file an issue for this and link it here?

src/magic_folder/web.py
```diff
@@ -114,11 +134,158 @@ class APIv1(Resource, object):
...
+            )
+
+            dmd = tahoe_uri_from_string(participant["personal_dmd"])
+            if not IDirnodeURI.providedBy(dmd):
+                raise ValueError("personal_dmd must be a directory-capability")
+            if not dmd.is_readonly():
+                raise ValueError("personal_dmd must be read-only")
```
tomprince, line 256: Same comments as elsewhere regarding `IDirnodeURI`.

src/magic_folder/web.py
```diff
@@ -114,11 +134,158 @@ class APIv1(Resource, object):
...
+            personal_dmd_cap = participant["personal_dmd"]
+        except ValueError as e:
```
tomprince, line 258: Does  `ValueError` get raised by anything other than our code? I wonder if if it would make sense to have a custom exception that gets handled here, instead of using `ValueError`.

tomprince (approved):
This looks reasonable. I've got a bunch of minor questions and comments, but all of them could be addressed as followup, if you'd prefer. I've not yet looked at the tests at all, but will do that after I have some food.

src/magic_folder/web.py
meejah: I figured I was following the convention of the other things (e.g. `snapshot` does it this way around too). I kind of agree that putting the folder-name higher-up might make more sense -- but we should maybe switch `/snapshot` around too?

I added a "question" ticket about this: https://github.com/LeastAuthority/magic-folder/issues/359 so perhaps could address in a followup (or not, depending on the conclusion ;)

src/magic_folder/api_cli.py
```diff
@@ -96,6 +97,64 @@ def dump_state(options):
...
+
+    def postOptions(self):
+        required_args = [
+            ("folder", "--folder / -n is required"),
+            ("author-name", "--author-name / -a is required"),
+            ("personal-dmd", "--personal-dmd / -p is required"),
+        ]
+        for (arg, error) in required_args:
+            if self[arg] is None:
+                raise usage.UsageError(error)
```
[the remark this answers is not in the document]
meejah: I would personally rather use `Click` for everything .. but I don't know that we have consensus on that.

src/magic_folder/api_cli.py
```diff
@@ -150,6 +209,8 @@ def get_client(self):
     subCommands = [
         ["add-snapshot", None, AddSnapshotOptions, "Add a Snapshot of a file to a magic-folder."],
         ["dump-state", None, DumpStateOptions, "Dump the local state of a magic-folder."],
+        ["add-participant", None, AddParticipantOptions, "Add a Participant to a magic-folder."],
+        ["list-participants", None, ListParticipantsOptions, "List all Participants in a magic-folder."],
```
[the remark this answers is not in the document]
meejah: @exarkun and I discussed this: the `magic-folder` CLI is a user-facing interface and often needs more consideration of its shape, options, etc. We decided it would be good to have a "low level" CLI that essentially just wraps the available HTTP APIs .. that is, you *could* do basically everything in the `magic-folder-api` with a properly-crafted `curl` command. This API needs fewer considerations around usability etc since it "just" represents the available endpoints.

So, it's more of a convenience for doing debugging or other low-level stuff where you just want to call the API endpoints. I guess we probably both were thinking of the `tahoe *` CLI which has an overwhelming number of sub-commands and options (although I'm not sure that part was made explicit).

We ultimately do want some high-level (that is `magic-folder *`) commands for user-management but those will most likely be centered around "invites" (e.g. see https://github.com/LeastAuthority/magic-folder/pull/312) and we definitely don't want users messing around with `Personal DMD` capability-strings if we can avoid it. (IMO). That said, `list` probably does make sense higher-level but for this PR I just wanted "plumbing" commands to poke at the API (and, arguably / ideally, the whole CLI part probably should have been a separate PR anyway).

src/magic_folder/api_cli.py
```diff
@@ -96,6 +97,64 @@ def dump_state(options):
...
+        options['folder'].decode("utf8"),
+        options['author-name'].decode("utf8"),
+        options['personal-dmd'].decode("utf8"),
+    )
+    print("{}".format(res), file=options.stdout)
+
+
+class ListParticipantsOptions(usage.Options):
+    optParameters = [
+        ("folder", "n", None, "Name of the magic-folder participants to list"),
```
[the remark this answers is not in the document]
meejah: maybe ... I guess I was going for "blind consistency" here?

src/magic_folder/participants.py
```diff
@@ -65,6 +72,21 @@ def list():
             associated with..
         """
```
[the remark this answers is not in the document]
meejah: @exarkun, do you remember? (I can't think of a reason for other implementations except for possibly testing considerations).

src/magic_folder/participants.py
```diff
@@ -142,6 +164,49 @@ def mutable_dirnode(self, attribute, value):
             ),
         )
 
+    @inlineCallbacks
+    def add(self, author, personal_dmd_cap):
+        """
+        IParticipants API
+        """
+        uri = tahoe_uri_from_string(personal_dmd_cap)
+        if not IDirnodeURI.providedBy(uri) or not uri.is_readonly():
```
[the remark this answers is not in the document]
meejah: This is a good point .. maybe we _do_ want to insist on `IReadonlyDirectoryURI` here.

Mostly we can ignore every `Literal` thing as "just" a special-case/optimization of `Immutable` .. but I don't see any use-case where we'd want a Personal DMD to be immutable. *Maaaaaybe* in the recovery flow (but then someone would have to go out of their way to turn a formerly-mutable Personal DMD into an immutable).

src/magic_folder/test/__init__.py
meejah: I did this to match with the `-ci` one so that the only difference between the two is the number of examples run

src/magic_folder/web.py
meejah: Not sure what you mean by "smarter"? Do you mean a 404 on a non-existent magic-folder name is insufficient?

src/magic_folder/web.py
meejah: https://github.com/LeastAuthority/magic-folder/issues/331 already exists (but yeah linking it in the comment is a good idea).

src/magic_folder/api_cli.py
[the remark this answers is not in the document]
tomprince: I definitely like the shape of click, using decorators to declare commands, on the functions that implement them. It does use global (or thread-local) data though, which I'm not a fan of. I've used a similar style API, built on top of argparse for a few projects like [this one](https://github.com/tomprince/tabletop-tools/blob/main/src/tts/cli.py).

src/magic_folder/test/__init__.py
tomprince: My thinking is that we want `magic-folder-fast` to be fast, and if something is triggering this health check, it isn't fast.

src/magic_folder/web.py
tomprince: I'm thinking:
- if the resource we return in the successful case isn't a leaf, will this handle requests to the subresource requests as expected
- do we want richer JSON error messages, particularly in the case with subresources.

src/magic_folder/test/test_web.py
```diff
@@ -654,3 +673,518 @@ def test_create_snapshot_fails(self, author, folder_name, path_outside_folder, s
...
+                treq,
+                AUTH_TOKEN,
+                b"POST",
+                self.url.child(folder_name),
+                dumps({
+                    "not-the-author": {"name": "kelly"},
+                }).encode("utf8")
+            ),
+            succeeded(
+                AfterPreprocessing(
```
tomprince, line 865: We should probably be checking the http response code in all these tests.

src/magic_folder/test/test_web.py
```diff
@@ -654,3 +673,518 @@ def test_create_snapshot_fails(self, author, folder_name, path_outside_folder, s
...
+            {
+                u"children": {
+                    author: format_filenode(folder_config["upload-dircap"]),
+                },
+            },
+        ])
+        tahoe_client = create_tahoe_client(
+            DecodedURL.from_text(u"http://invalid./"),
+            create_tahoe_treq_client(root),
+        )
```
tomprince, line 1088: This is dead code.

src/magic_folder/test/test_web.py
```diff
@@ -654,3 +673,518 @@ def test_create_snapshot_fails(self, author, folder_name, path_outside_folder, s
...
+        root = create_fake_tahoe_root()
+        # put our Collective DMD into the fake root
+        root._uri.data[folder_config["collective-dircap"]] = dumps([
+            u"dirnode",
+            {
+                u"children": {
+                    author: format_filenode(folder_config["upload-dircap"]),
+                },
+            },
+        ])
```
tomprince, line 1151: This is dead code.

src/magic_folder/test/test_web.py
```diff
@@ -654,3 +673,518 @@ def test_create_snapshot_fails(self, author, folder_name, path_outside_folder, s
...
+        )
+
+    @given(
+        author_names(),
+        folder_names(),
+        tahoe_lafs_readonly_dir_capabilities(),
+    )
+    def test_add_participant_internal_error(self, author, folder_name, personal_dmd):
+        """
+        An internal error on participant adding is logged
```
tomprince, line 1132: I don't see any code related to logging in this test.

tomprince (approved):

src/magic_folder/test/test_api_cli.py
```diff
@@ -599,3 +599,193 @@ def test_happy(self):
...
+                 {
+                     b'Host': [b'invalid.'],
+                     b'Connection': [b'close'],
+                     b'Authorization': [b'Bearer ' + self.global_config.api_token],
+                     b'Accept-Encoding': [b'gzip']
+                 },
+                 b"",
+                ),
+                # expected response
+                (200, {}, b"{}"),
```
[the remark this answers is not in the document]
meejah: We do test "actual responses" on the HTTP endpoint in the `test_web` tests for Participants .. I believe my thinking here is that the point (for the `magic-folder-api` CLI test) is merely to ensure that the CLI calls the right HTTP endpoint.

src/magic_folder/test/test_web.py
```diff
@@ -266,6 +281,8 @@ def treq_for_folders(reactor, basedir, auth_token, folders, start_folder_service
     :param bool start_folder_services: If ``True``, start the Magic Folder
         service objects.  Otherwise, don't.
 
+    :param TahoeClient tahoe_client: if provided, used as the tahoe-client
+
```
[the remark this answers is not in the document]
meejah: Yeah we probably do need to return more than the treq stand-in .. the thing we need to call `stopService()` on is the `global_service` instance.

Conceptually it might be possible to set up the correct tahoe data for the given "folders" mapping .. but I feel like it's probably more-explicit/easier to read if that setup happens in the test itself?

(`test_add_participant` is a test that actually sets up some Tahoe data, for comparison .. so I guess the alternative is to provide a `tahoe_client_for_participants()` or similar that sets up the Collective DMD properly and move those lines into that function instead...?)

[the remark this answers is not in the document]
tomprince: > Yeah we probably do need to return more than the treq stand-in .. the thing we need to call stopService() on is the global_service instance.

I think this isn't true, given that we don't start it, but only some children.

> Conceptually it might be possible to set up the correct tahoe data for the given "folders" mapping .. but I feel like it's probably more-explicit/easier to read if that setup happens in the test itself?
>
> (test_add_participant is a test that actually sets up some Tahoe data, for comparison .. so I guess the alternative is to provide a tahoe_client_for_participants() or similar that sets up the Collective DMD properly and move those lines into that function instead...?)

I don't have strong feelings on this either way. We can keep the code in the test, but updating `tahoe_root._uri.data` after the fact (or, we could thread passing that value through, and expose it as a parameter).

src/magic_folder/test/test_api_cli.py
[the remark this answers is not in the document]
tomprince: There is some printing that we don't test, if there are no participants. It may not be worth it to test that, though.

src/magic_folder/test/test_participants.py
```diff
@@ -206,6 +226,252 @@ def test_list(self, collective_contents, rw_collective_dircap):
...
+        for name, dircap in collective_contents.items():
+            if name == author:
+                continue
+            participants.add(
+                RemoteAuthor(name, VerifyKey(os.urandom(32))),
+                to_readonly_capability(dircap),
+            )
+
+        # confirm we added all the right participants by using the
+        # list() API
```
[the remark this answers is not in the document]
meejah: If I understand the suggestion properly, I think the answer is "no" --  because that's what the `Participants` API is supposed to be doing -- that is, abstracting away "how Tahoe represents things" .. 

That said, we're already somewhat tied to that representation because a few lines up we have to put fake Tahoe data in which means we're pretty commited to the JSON returned by Tahoe's WebUI (and really all we'd be testing there is that our Fake is correct still, which arguably should  be in other tests).

So, I think the Tahoe fake (with its tests) should handle checking that the shape of the JSON is good after e.g. calling "`TahoeClient.add_entry_to_directory()`" and this code/test can just look at the output of the higher-level API (`IParticipants.list()`)

src/magic_folder/test/test_web.py
```diff
@@ -654,3 +673,518 @@ def test_create_snapshot_fails(self, author, folder_name, path_outside_folder, s
...
+            FilePath(self.mktemp()),
+            local_path,
+        )
+        # we can't add a new participant if their DMD is the same as
+        # one we already have .. and because Hypothesis is 'sneaky' we
+        # have to make sure it's not our collective, either
+        assume(personal_dmd != folder_config["upload-dircap"])
+        assume(personal_dmd != to_readonly_capability(folder_config["upload-dircap"]))
+        assume(personal_dmd != folder_config["collective-dircap"])
+        assume(personal_dmd != to_readonly_capability(folder_config["collective-dircap"]))
```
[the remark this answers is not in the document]
meejah: yeah, I think I already tried to move the `assume()`s up higher before and hit that .. essentially the issue being that `magic_folder_config()` makes some fake dircaps itself and if Hypothesis finds those for the Personal DMDs it becomes an error.

src/magic_folder/test/test_web.py
meejah: cb32a91f (testing to see if github links to the commit automagically)

src/magic_folder/test/test_participants.py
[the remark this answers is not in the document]
tomprince: > If I understand the suggestion properly, I think the answer is "no" -- because that's what the Participants API is supposed to be doing -- that is, abstracting away "how Tahoe represents things" ..

Well, the thing we aren't testing is if we are representing the abstraction "reasonably" or "as we expect". Which, granted, depends somewhat on our fake being accurate. Given that we want to maintain forward and backwards compatibility, at some point, we probably do want tests that make assertions about what calling our abstractions does on the underlying tahoe level.

That said, this may not be the right place for those tests, and they should instead by in the integration tests, where we inspect the state in a real tahoe cluster after our operations.

src/magic_folder/web.py
meejah: I *thought* I was copying that style from elsewhere, but now I don't see where .. I don't *think* that `ValueError` comes from any of the called functions .. but even if it did, returning that as an error still seems appropriate?

tomprince: It isn't clear that we want to expose the message of a random other `ValueError` unchanged to the user, though.

src/magic_folder/api_cli.py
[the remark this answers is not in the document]
meejah: (Punting on these `magic-folder-api` considerations in the context of this PR: see Slack discussion instead).

src/magic_folder/api_cli.py
```diff
@@ -96,6 +97,64 @@ def dump_state(options):
...
+
+@inlineCallbacks
+def add_participant(options):
+    """
+    Add one new participant to an existing magic-folder
+    """
+    res = yield options.parent.get_client().add_participant(
+        options['folder'].decode("utf8"),
+        options['author-name'].decode("utf8"),
+        options['personal-dmd'].decode("utf8"),
```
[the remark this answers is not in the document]
meejah: See #360

src/magic_folder/api_cli.py
```diff
@@ -248,6 +309,8 @@ def run_magic_folder_api_options(options):
     main_func = {
         "add-snapshot": add_snapshot,
         "dump-state": dump_state,
+        "add-participant": add_participant,
+        "list-participants": list_participants,
```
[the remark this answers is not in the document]
meejah: See #360

src/magic_folder/api_cli.py
```diff
@@ -96,6 +97,64 @@ def dump_state(options):
...
+        ("personal-dmd", "p", None, "Read-capability of the participant's Personal DMD"),
+        # not yet
+        # ("author-verify-key", "k", None, "Base32-encoded verify key of the new participant"),
+    ]
+
+    def postOptions(self):
+        required_args = [
+            ("folder", "--folder / -n is required"),
+            ("author-name", "--author-name / -a is required"),
+            ("personal-dmd", "--personal-dmd / -p is required"),
```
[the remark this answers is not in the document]
meejah: See #360

src/magic_folder/test/test_web.py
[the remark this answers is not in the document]
meejah: > I think this isn't true, given that we don't start it, but only some children.

yes, that's more-accurate ;) but we do get the sub-services from it..

newsfragments/327.feature
```diff
@@ -1 +1 @@
-There is now an HTTP API to add a new participant
+There is now an HTTP API to add and list new participants, along with magic-folder-api subcommands
```
tomprince, line 1:
```suggestion
There is now an HTTP API to add and list new participants, along with corresponding magic-folder-api subcommands
```

src/magic_folder/participants.py
```diff
@@ -156,7 +157,7 @@ def mutable_dirnode(self, attribute, value):
         The Upload DMD must be a writable directory capability
         """
         uri = tahoe_uri_from_string(value)
-        if IDirnodeURI.providedBy(uri):
+        if IDirectoryURI.providedBy(uri):
             if not uri.is_readonly():
```
tomprince, line 161: I *think* this is implied by `IDirectoryURI`, though the documentation doesn't make it clear.

tomprince (approved):

src/magic_folder/participants.py
meejah: hmmm, yeah reading the code I believe you're correct.

Tempted to leave it there so future-me doesn't have to re-read the Tahoe code every time ... maybe I should have put all this stuff in `capabilities.py` anyway? (e.g. `is_readonly_directory_cap()` and `is_readwrite_directory_cap()` approximately) ...

meejah: or `is_directory_capability()` along with `is_readonly_capability()` or so ....

meejah: took out redundant check at least

src/magic_folder/web.py
meejah: yeah, fair point .. I'll make a custom exception

src/magic_folder/test/__init__.py
meejah: Moved to PR: https://github.com/LeastAuthority/magic-folder/pull/361

PR #337, opened 2021-05-20
Replaces uses of `allmydata` re-exports with the thing actual source.

tomprince:


meejah (approved):
LGTM :+1:

meejah:
BTW, the inotify/watchdog etc stuff was kind of a mess in old-magic-folders so .. still TBD how exactly that stuff will work in the future. For now, there's an HTTP API and `magic-folder-api add-snapshot` to call it to add work. Point being: not worried about test-coverage in there for immediate term.

Issue #170, opened 2020-07-01
The "Project" coverage CI job is flaky


exarkun:
According to [codecov](https://codecov.io/gh/LeastAuthority/magic-folder/branch/master) coverage bounces around in the ~87% range.  It changes even when changesets clearly include nothing which could change the coverage level.  This causes PRs to get marked as having a failing codecov job when they are not at fault.

We should have stable, deterministic coverage results and then mark the codecov project job as required.

PR #338, opened 2021-05-20
Run macOS testing on pull-requests as well.

tomprince:
The unit-tests are failing, but they are failing on main as well.

### DIFF
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -2,6 +2,7 @@ name: macOS Testing
 
 on:
   - "push"
+  - "pull_request"
 
 jobs:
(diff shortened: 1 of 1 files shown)


tomprince:
This may partly address #170.

meejah (approved):
No opposed to this .. but is there a way to mark that builder is "optional" or so (as it is, this change will mean that every PR is marked as failed, right?)

tomprince:
> No opposed to this .. but is there a way to mark that builder is "optional" or so (as it is, this change will mean that every PR is marked as failed, right?)

Based on https://github.com/actions/toolkit/issues/399 it looks like there isn't, currently.

PR #349, opened 2021-05-21
Pass dockerhub auth through circleci YAML correctly.

tomprince:
This enables using stored credentials to avoid dockerhub rate limits[1].
Previously, the contents of the `auth` key was being passed as keys to docker.

[1] https://support.circleci.com/hc/en-us/articles/360050623311-Docker-Hub-rate-limiting-FAQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -125,7 +125,7 @@ jobs:
 
   lint:
     docker:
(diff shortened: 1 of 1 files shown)


tomprince:
Unfortunately, we can't really verify this is working until we land it, since the auth is not passed to PRs.

meejah (approved):

PR #350, opened 2021-05-21
Build newer debian and pypy docker images.

tomprince:


tomprince:
Landing this now, so it is picked up by the next cron image build.

Issue #335, opened 2021-05-20
realtime status API


meejah:
To support GUI clients such as GridSync a status endpoint is desired. UI applications will connect via a WebSocket connection. In order that a UI doesn't miss any messages all messages will be buffered until at least one client connects to the status API; it will then receive all pending messages and begin receiving new ones in real time.

The status API shall report ongoing upload or download activity as a message indicating the total size of pending operations and their current percent completed. It is also desirable to have notification of new files added (that's "new LocalSnapshot added" I guess). Some notification of "idle" versus "synchronizing" is desired as well.

Currently the only concrete user is GridSync and it only currently has UI for showing a "sync-ing" indicator and the percent progress.

meejah:
The code in 340 adds the general infrastructure for this feature as well as a single sort of message:

```
{
    "kind": "synchronizing",
    "status": true
}
```

The "status" will be `true` whenever there is work to do and `false` when there isn't. All messages are buffered whenever zero clients are connected (so e.g. GridSync should never miss any messages).

meejah:
For lack of an immediate better place to put this, here is some example code that can use this API...

```python
import json

from autobahn.twisted.websocket import (
    WebSocketClientProtocol,
    WebSocketClientFactory,
)
from twisted.internet.endpoints import (
    clientFromString,
)
from twisted.internet.task import (
    react,
)
from twisted.internet.defer import (
    Deferred,
)


class StatusProtocol(WebSocketClientProtocol):
    """
    Client-side protocol that receives status updates from
    magic-folder
    """

    def onOpen(self):
        print("WebSocket connection open.")

    def onMessage(self, payload, isBinary):
        data = json.loads(payload)
        print("status: {kind}={status}".format(**data))

    def onClose(self, wasClean, code, reason):
        print("WebSocket connection closed: {0}".format(reason))


if __name__ == '__main__':

    def main(reactor):
        with open("alice/api_token", "r") as f:
            token = f.read().strip()
        factory = WebSocketClientFactory(
            "ws://127.0.0.1:4444/v1/status",
            headers={
                "Authorization": "Bearer {}".format(token),
            }
        )
        factory.protocol = StatusProtocol
        endpoint = clientFromString(reactor, "tcp:127.0.0.1:4444")
        endpoint.connect(factory)
        return Deferred()
    react(main)
```

meejah:
[...]

PR #352, opened 2021-05-22
monitor status

### LINKED ISSUE #351: magic-folder-api monitor
As part of #335 it would be great for debugging and example-code to have a command-line convenience to access the realtime status API.

meejah:
Fixes #351

Serves both as a debugging aid and place to have example WebSocket client code.

This PR is stacked on #335 so review/merge that one first.

### DIFF
--- a/newsfragments/351.feature
+++ b/newsfragments/351.feature
@@ -1,0 +1,1 @@
+Add 'magic-folder-api monitor' command
--- a/src/magic_folder/api_cli.py
+++ b/src/magic_folder/api_cli.py
@@ -7,6 +7,17 @@ import json
 from twisted.internet.task import (
     react,
 )
+from twisted.internet.defer import (
+    Deferred,
+)
+from twisted.internet.endpoints import (
+    clientFromString,
+)
+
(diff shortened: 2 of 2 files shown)


meejah:
Bit of a testing SNAFU here: although Autobahn does have a facility for testing a client/server interaction (see https://github.com/crossbario/autobahn-python/blob/master/autobahn/twisted/testing/__init__.py#L280) there was a bug in it that clobbered the factory (which we do need to provide here). See https://github.com/crossbario/autobahn-python/pull/1480

Although that bug is fixed, there is no release with containing the fix. _However_ even if there were a release Autobahn is now python3-only -- so a new release wouldn't help us because we have to use an older python2-supporting release anyway.

meejah:
(p.s. @tomprince thought of a workaround for the Autobahn thing, which was pushed into the "upstream" status-api PR)

tomprince (approved):
Looks good. I assume you've tested this manually?

meejah:
> Looks good. I assume you've tested this manually?

Yeah .. inclined to add at least a basic test first since you suggested the workaround for Autobahn bug.

PR #355, opened 2021-05-23
?t=json for dircaps from uri endpoint

### LINKED ISSUE #354: Tahoe's /uri endpoint sometimes needs ?t=json
When fetching capabilities from Tahoe, the `/uri` endpoint is used. Normally this returns the content. However, if we are downloading a directory-capability, the endpoint returns an HTML page .. so `?t=json` is required to get the JSON we expect.

It is not possible to *always* use `?t=json` because if you do that on a normal file you only get the JSON "structure" of the content, not the actual content.

The solution is for `TahoeClient.download_capability` must learn what sort of capability it has and use `?t=json` for and directory capabilities.

meejah:
Fixes #354

### DIFF
--- a/src/magic_folder/tahoe_client.py
+++ b/src/magic_folder/tahoe_client.py
@@ -335,9 +335,11 @@ class TahoeClient(object):
         returnValue(capability_string)
 
     @inlineCallbacks
-    def download_capability(self, cap):
+    def download_file(self, cap):
         """
-        Retrieve the raw data for a capability from Tahoe
+        Retrieve the raw data for a capability from Tahoe. It is an error
+        if the capability-string is a directory-capability, since the
+        Tahoe `/uri` endpoint treats those specially.
 
         :param cap: a capability-string
 
@@ -345,12 +347,18 @@ class TahoeClient(object):
         """
         # Visiting /uri with a directory-capability causes Tahoe to
         # render an HTML page .. instead adding ?t=json tells it to
-        # send JSON instead. However, using ?t=json for non-directory
-        # capabilities means the content is *not* downloaded .. so,
-        # "t=json" is only added for directory-capabilities.
-        query_args = [(u"uri", cap.decode("ascii"))]
+        # send the JSON instead. We can't just use ?t=json all the
+        # time though, because for non-directories it returns the
+        # metadata, not the data. So, the caller needs to know if they
+        # have a dir-cap or not and call list_directory() .. if
+        # anything really does need the "raw" directory JSON we need
+        # another method on TahoeClient for that.
         if is_directory_cap(cap):
-            query_args.append((u"t", u"json"))
+            raise ValueError(
+                "{} is a directory-capability not a regular file".format(cap)
+            )
+
+        query_args = [(u"uri", cap.decode("ascii"))]
 
         get_uri = self.url.child(u"uri").replace(query=query_args)
         res = yield _request(
--- a/src/magic_folder/tahoe_client.py
+++ b/src/magic_folder/tahoe_client.py
@@ -294,8 +294,8 @@ class TahoeClient(object):
         :returns dict: the JSON representing this directory
         """
         if not is_directory_cap(dir_cap):
-            raise VaueError(
-                "{} is not a directory-capability".format(dir-cap)
+            raise ValueError(
+                "{} is not a directory-capability".format(dir_cap)
             )
         api_uri = self.url.child(
             u"uri",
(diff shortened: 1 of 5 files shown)


src/magic_folder/tahoe_client.py
```diff
@@ -338,9 +343,16 @@ def download_capability(self, cap):
...
-            query=[(u"uri", cap.decode("ascii"))],
-        )
+        # Visiting /uri with a directory-capability causes Tahoe to
+        # render an HTML page .. instead adding ?t=json tells it to
+        # send JSON instead. However, using ?t=json for non-directory
+        # capabilities means the content is *not* downloaded .. so,
+        # "t=json" is only added for directory-capabilities.
+        query_args = [(u"uri", cap.decode("ascii"))]
+        if is_directory_cap(cap):
+            query_args.append((u"t", u"json"))
```
tomprince, line 353: I'm not convinced that we should be inspecting the cap to determine whether we should request the json format or not. My impression is, that every caller should know whether they are requesting a directory or not, so we could have a different function for handling directories, as compared to files (and that function could even load the returned json and maybe(??) parse it).

I'm not opposed to checking that we are requesting a file before making a call without `?t=json`, but I think we should error out if that is not the case (and should maybe check for `IFileURI` instead of `IDirectoryURI`).

I'm not sure if we want to separate getting file metadata from directory listing or not; they use the same API request, but return (somewhat) different data.

tomprince (dismissed):

src/magic_folder/tahoe_client.py
meejah: So you want like `download_directory_capability` versus `download_file_capability` or so...?

meejah: We really are downloading a capability, though (in both cases). "A directory" in Tahoe is just a text file that contains some JSON describing the directory structure. But, sure, we could leave that out (it should be clear from the arg-name anyway).

meejah: (Maybe the only real change we need here is to error-out from `download_capability` if its a directory-cap .. and maybe rename the method, too)...?

meejah: This PR I think was motivated from some download changes I was making and I saw HTML getting spit out of `download_capability()` .. so a (proper) error would be just as good (since we already have a `list_directory` that uses `/uri?t=json`)

tomprince: > We really are downloading a capability,

We are downloading what a capability is pointing at :upside_down_face:; we already have the capability. You wouldn't say `download_file_name`, this feels equivalent.

tomprince: > This PR I think was motivated from some Downloader changes I was making and I saw HTML getting spit out of `download_capability()` .. so a (proper) error would be just as good (since we already have a `list_directory` that uses `/uri?t=json`)

That makes sense.

meejah:
okay, I re-worked this so `download_capability` is now `download_file` and errors on directories, and added a `directory_data()` (as opposed to the existing `list_directory()`) that returns the full JSON of the data.

meejah:
(hmm, I clicked "dismiss review" which seemed like the right thing, but also had a popup warning so maybe it does bad things? anyway, just trying to note that the old review isn't relevant since I re-worked the code significantly)

tomprince:
> (hmm, I clicked "dismiss review" which seemed like the right thing, but also had a popup warning so maybe it does bad things? anyway, just trying to note that the old review isn't relevant since I re-worked the code significantly)

Instead of "dismiss review", you can click the little arrows in a circle next to the reviewer name in the sidebar to request a re-review.

src/magic_folder/tahoe_client.py
```diff
@@ -335,22 +367,29 @@ def add_entry_to_mutable_directory(self, mutable_cap, path_name, entry_cap, repl
...
-        # send JSON instead. However, using ?t=json for non-directory
-        # capabilities means the content is *not* downloaded .. so,
-        # "t=json" is only added for directory-capabilities.
-        query_args = [(u"uri", cap.decode("ascii"))]
+        # send the JSON instead. We can't just use ?t=json all the
+        # time though, because for non-directories it returns the
+        # metadata, not the data. So, the caller needs to know if they
+        # have a dir-cap or not and call list_directory() or
+        # directory_data() to get "raw" representation
         if is_directory_cap(cap):
```
tomprince, line 387: Instead of checking that it isn't a directory cap, can you check that it is a file cap? I think this means `IFileURI` or `IMutableFileURI`. After making this suggestion, I see that this avoid `IVerifierURI`, but I think generally allow-lists make more sense than deny-lists for this kind of check.

tomprince (approved):

src/magic_folder/tahoe_client.py
meejah: Strictly speaking, Tahoe could have new as-yet-unknown URIs that would download just fine (remember the point of this restriction/change is that the Tahoe endpoint renders HTML for dir-caps, but not for other things) .. 

.. but, okay, we say "file" in the API now so I suppose it makes sense to insist it's only one of the known file-cap types.

PR #356, opened 2021-05-23
snapshot names

### LINKED ISSUE #353: path-names in Snapshots are wrong
When Snapshots are uploaded they are getting a mangled pathname that is the full path. This is incorrect; it must be the relative path from the base of the magic-folder directory (and mangled).

meejah:
Fixes #353

### DIFF
--- a/src/magic_folder/magicpath.py
+++ b/src/magic_folder/magicpath.py
@@ -4,31 +4,40 @@ import os.path
 from pyutil.assertutil import precondition, _assert
 
 
-def mangle_path_segments(segments):
+_DEMANGLER = re.compile(u'@[_@]')
+
+
+def path_to_label(base, path):
     """
-    Perform name-flattening on a list of path segments. Each element
-    represents one segment of a path (e.g. (sub)directory or the
-    actual filename) and is unicode.
+    :param FilePath base: the base directory (for example, the "magic
+        folder dir")
 
-    The mangling consists of replacing path-separators with '@_' and
-    escaping '@' symbols inside a path-segment as '@@'.
+    :param FilePath path: the actual file (for example, a local file
+        we're turning into a Snapshot). Must be child if `base`.
 
-    :param list[unicode] segments: 1 or more unicode path segments
-
-    :returns unicode: flattened / mangled name
+    :returns unicode: the relative path from `base` to `path`
+        "mangled" into a Snapshot name. It is an error if `path` is
+        not a child of `base`. The mangled label can be turned back
+        into a FilePath using `label_to_path.
     """
+    segments = path.segmentsFrom(base)
     return u'@_'.join(
         segment.replace(u'@', u'@@')
         for segment in segments
     )
 
 
-def unmangle_relative_path(mangled_path):
+def label_to_path(base, mangled_name):
     """
-    Undo the name-mangling achieved by `mangle_path_segments`. That
-    is, split on `@_` and turn `@@` inside a segment back into `@`.
+    :param FilePath base: the base directory below which the
+        un-mangled name will exist.
 
-    :returns unicode: the un-mangled relative path separated by "/".
+    :param unicode mangled_name: the output of a previous call to
+        `path_to_label`, this encodes a relative path.
+
+    :returns FilePath: an absolute FilePath below `base` matching this
+       manged name. Will raise InsecurePath if the path would end up
+       above `base`.
     """
 
     def replace(match):
@@ -36,7 +45,8 @@ def unmangle_relative_path(mangled_path):
             u'@_': u'/',
             u'@@': u'@',
         }[match.group(0)]
-    return re.sub(u'@[_@]', replace, mangled_path)
+    relpath = _DEMANGLER.sub(replace, mangled_name)
+    return base.preauthChild(relpath)
 
 
 IGNORE_SUFFIXES = [u'.backup', u'.tmp', u'.conflict']
--- a/src/magic_folder/magicpath.py
+++ b/src/magic_folder/magicpath.py
@@ -3,11 +3,44 @@ import os.path
 
 from pyutil.assertutil import precondition, _assert
 
+
 def path2magic(path):
     return re.sub(u'[/@]',  lambda m: {u'/': u'@_', u'@': u'@@'}[m.group(0)], path)
 
 def magic2path(path):
     return re.sub(u'@[_@]', lambda m: {u'@_': u'/', u'@@': u'@'}[m.group(0)], path)
+
+
+def mangle_path_segments(segments):
+    """
+    Perform name-flattening on a list of path segments. Each element
+    represents a (sub)directory and is unicode.
+
+    The mangling consists of replacing path-separators with '@_' and
+    escaping '@' symbols inside a path-segment as '@@'.
+
+    :param list[unicode] segments: 1 or more unicode path segments
+
+    :returns unicode: flattened / mangled name
+    """
+    return u'@_'.join(
+        segment.replace(u'@', u'@@')
+        for segment in segments
+    )
+
+
+def unmangle_path_segments(mangled_path):
+    """
+    Undo the name-mangling achieved by `mangle_path_segments`. That
+    is, split on `@_` and turn `@@` inside a segment back into `@`.
+
+    :returns list[unicode]: the un-mangled path segments as a list of
+        unicode objects, one per segment. This is a relative path.
+    """
+    return [
+        segment.replace(u'@@', u'@')
+        for segment in mangled_path.split(u'@_')
+    ]
 
 
 IGNORE_SUFFIXES = [u'.backup', u'.tmp', u'.conflict']
--- a/src/magic_folder/magicpath.py
+++ b/src/magic_folder/magicpath.py
@@ -23,20 +23,19 @@ def mangle_path_segments(segments):
     )
 
 
-def unmangle_path_segments(mangled_path):
+def unmangle_relative_path(mangled_path):
     """
     Undo the name-mangling achieved by `mangle_path_segments`. That
     is, split on `@_` and turn `@@` inside a segment back into `@`.
 
-    :returns list[unicode]: the un-mangled path segments as a list of
-        unicode objects, one per segment. This is a relative path.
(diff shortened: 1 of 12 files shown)


meejah:
I haven't been able to re-produce the failures on (some) CI machines reliably locally .. but I've obviously got something misaligned unicode-wise ...

src/magic_folder/magicpath.py
```diff
@@ -3,11 +3,40 @@
 
 from pyutil.assertutil import precondition, _assert
 
-def path2magic(path):
-    return re.sub(u'[/@]',  lambda m: {u'/': u'@_', u'@': u'@@'}[m.group(0)], path)
 
-def magic2path(path):
-    return re.sub(u'@[_@]', lambda m: {u'@_': u'/', u'@@': u'@'}[m.group(0)], path)
+def mangle_path_segments(segments):
```
tomprince, line 7: I see why this and `unmangle_relative_path` are not mirrors (based on how they are used), but I wonder if that will be confusing down the line. Additionally, I wonder if the fact that this combines segments will be surprising.

src/magic_folder/magicpath.py
```diff
@@ -3,11 +3,40 @@
...
+
+    :returns unicode: flattened / mangled name
+    """
+    return u'@_'.join(
+        segment.replace(u'@', u'@@')
+        for segment in segments
+    )
+
+
+def unmangle_relative_path(mangled_path):
```
tomprince, line 26: Is there anything that makes this specific to relative paths?

src/magic_folder/magicpath.py
tomprince: One way to address this would be to have a `relative_to` function, that does `segmentsFrom` and then joints them with `/`.

src/magic_folder/test/test_local_snapshot.py
```diff
@@ -289,15 +290,17 @@ def test_create_snapshots(self, filenames, data_strategy):
 
         self.assertThat(self.db.get_all_localsnapshot_paths(), HasLength(len(files)))
         for (file, content) in files:
-            mangled_filename = path2magic(file.asTextMode(encoding="utf-8").path)
+            mangled_filename = mangle_path_segments(
+                file.asTextMode("utf8").segmentsFrom(self.magic.asTextMode("utf8"))
```
tomprince, line 294: I guess this may be fodder for a separate PR, but it feels like `.asTextMode` should be something that is happening once, closer to the edge of the system, rather than in the middle of slightly complex logic.

src/magic_folder/uploader.py
```diff
@@ -95,7 +96,10 @@ def store_local_snapshot(self, path):
             # Query the db to check if there is an existing local
             # snapshot for the file being added.
             # If so, we use that as the parent.
-            mangled_name = magicpath.mangle_path(path)
+
+            mangled_name = mangle_path_segments(
+                path.asTextMode("utf8").segmentsFrom(self._magic_dir.asTextMode("utf8"))
```
tomprince, line 101: I think we should be converting these paths to unicode closer to the edge of the system.

src/magic_folder/uploader.py
```diff
@@ -191,7 +196,7 @@ def add_file(self, path):
         textpath = path.asTextMode("utf-8")
 
         try:
-            bytespath.segmentsFrom(self._magic_path)
+            bytespath.segmentsFrom(self._magic_path.asBytesMode("utf-8"))
```
tomprince, line 199: Same here.

src/magic_folder/magicpath.py
tomprince: Reading through some of the tests and things, it seems there are enough places that start with a relative path, that I think pushing the joining code to the caller that does `.segementsFrom` probably does make sense.

tomprince (changes requested):

src/magic_folder/magicpath.py
meejah: "mangled" paths should never be absolute

src/magic_folder/magicpath.py
meejah: should i just put the original `magicpath` code back, then?

meejah: At least part of my motivation for changing it is that we should only ever be "mangling" relative paths, and the only FilePath API for that returns list-of-strings...maybe an even-different API would make it harder to screw up?

Something like `mangle_relative_to(base, path)` where base, path are both FilePaths and `unmangle_relative_to(base, mangled)` returns a FilePath (an absolute one, but starting at "base")

src/magic_folder/test/test_local_snapshot.py
meejah: I don't really get why we're calling `.asXMode()` where we are as it is .. so happy to have a better way to determine this (I added the `.asTextMode` to self.magic because I presume segmentsFrom needs them both in the same mode).

src/magic_folder/magicpath.py
tomprince: > should i just put the original magicpath code back, then?

If we decide to go with that API, it probably makes sense, though the formatting of `unmangle_relative_path` is definitely more readable than the one-line version.

> At least part of my motivation for changing it is that we should only ever be "mangling" relative paths, and the only FilePath API for that returns list-of-strings...maybe an even-different API would make it harder to screw up?

I definitely understand the desire for an interface that is hard to get wrong. I'm not sure how to balance that desire against having an unsymmetrical API and all the places where we call the API with an already relative path where we are now calling `.split`. 

> Something like mangle_relative_to(base, path) where base, path are both FilePaths and unmangle_relative_to(base, mangled) returns a FilePath (an absolute one, but starting at "base")

There are a number of places where we interact with mangled names where we don't have or want and absolute unmangled path, so I don't think these make sense.

src/magic_folder/test/test_local_snapshot.py
tomprince: > so happy to have a better way to determine this

My feeling would be:
- whenever we construct a FilePath (or receive one from library code), we use `asTextMode` on it (unless we know we are constructing it from a `unicode`)
- we only use `asBytesMode` before accessing `.path` when passing the result to a function that expects `bytes`.

But, as I said, we don't need to handle that in this PR.

src/magic_folder/magicpath.py
tomprince: I guess my thought here is that
1) the API "correctly" handles mangled absolute paths
2) there is nothing that enforces that the paths are relative.

so the name feels slightly misleading. I guess one solution would to verify that the path is relative before returning it?

src/magic_folder/magicpath.py
meejah: I only see "split" calls in tests .. so a different way to mitigate that would be to change how the `strategies` produce relative paths.

tomprince: After writing https://github.com/LeastAuthority/magic-folder/pull/356#discussion_r638951794, one thought that occurred to me is to have:
```python
def mangle_relative_to(path: FilePath, base: FilePath) -> unicode:
    ...
def mangle_relative_path(path: FilePath) -> unicode:
    ...
def unmangle_relative_path(mangled: unicode) -> FilePath:
    ...
```
and both the single argument versions could check that `.isabs` is `False` before proceeding. (Note that we'd want the first two implemented via the same internal function, or one in terms of the other)

tomprince: > I only see "split" calls in tests .. so a different way to mitigate that would be to change how the `strategies` produce relative paths.

I can imagine wanting to add the ability to access files in snapshots directly, which would also have the same kind of issue.

meejah: > There are a number of places where we interact with mangled names where we don't have or want and absolute unmangled path, so I don't think these make sense.

I guess there's kind of two uses here: in one sense the mangled-relative-name is just the "name" or "label" of a Snapshot; in another sense it is turned into a local filesystem path. _Conceptually_ we probably shouldn't ever care that the "name" happens to be the (mangled) path relative to the magic-folder *except* when we're about to open a file (for reading or writing).

So, where are we using an unmangled relative path? Perhaps those places should just not depend on that knowledge at all and simply use it like a label? (I think we do need to receive unmangled relative paths in APIs .. e.g. the "add a snapshot" API)

meejah:
Another suggestion for a different API (further to my comment about "name" / "label") is:

- `absolute_path_to_label(magic_folder_base, filename)` (where both args are `FilePath`s) returns a unicode that is the "mangled", relative pathname
- `label_to_absolute_path(magic_folder_base, label)` does the opposite of above

So anything that receives local filenames uses the first method to turn things into Snapshot names. Everything else uses "labels" / "names" of Snapshots, except code that needs to open the local file corresponding to a Snapshot: it uses `label_to_absolute_path` to get a `FilePath`.

@tomprince

meejah:
Okay, I re-worked the API and call-sites into something approximately between our two suggestions. The mangle/unmanle functions are mirrors and the outside only sees absolute `FilePath`s which I think is good because Snapshots always refer to "a thing inside your magic-folder".

meejah:
p.s. I tried to make the `.asTextMode` / `asBytesMode` make more sense, but that deserves a closer look since I'm not very confident about those changes.

meejah:
CI failure is unrelated to this PR, it seems: magic-wormhole is now producing warnings for using `attr.s(cmp=)` ...

tomprince:
As I'm reviewing this, I realized that I was confusing `pathlib.Path` and `FilePath`, in that the former can be relative whereas the latter can't be, so my suggestion to return relative `FilePath`s was non-nonsensical.

I think the return value of `_CollectiveDirnodeParticipant.files()` should not be local absolute paths, so I don't think we want to force local names to be absolute. I'm going to take a go at this, and restrict myself to fixing #353, without trying to come up with a harder-to-misuse interface. Once that is done, we can bikeshed on the interface.

meejah:
Why do you think `_CollectiveDirnodeParticipant.files()` shouldn't return absolute local paths? What use-case is there for anything else? (I think that literally nothing uses that API so maybe we should just delete it and ignore all this?)

tomprince:
Closing this in favour of #364 for now.

PR #362, opened 2021-05-26
Use klein for web API.

tomprince:


src/magic_folder/web.py
```diff
@@ -136,100 +140,55 @@ class APIv1(Resource, object):
...
+        except Exception:
             # probably should log this failure
             request.setResponseCode(http.INTERNAL_SERVER_ERROR)
             _application_json(request)
-            request.write(json.dumps({"reason": "unexpected error processing request"}))
-            return None
-        d.addErrback(failed)
-        d.addBoth(lambda ignored: request.finish())
-        return NOT_DONE_YET
+            returnValue(json.dumps({"reason": "unexpected error processing request"}, ensure_ascii=False).encode("utf-8"))
```
meejah, line 167: Why do we need to change `ensure_ascii=` from the default here?

meejah (changes requested):
Overall looks good. Couple questions inline -- and looks like we need several variations on the "folder not found" test now since this new layout needs to check that in multiple spots instead of one (i.e. a few untested error-cases).

src/magic_folder/web.py
tomprince: I dropped it. I tend to default to setting to false, since everything can deal with utf-8 encoded json, and it makes viewing it by hand easier.

src/magic_folder/web.py
```diff
@@ -119,13 +122,14 @@ def getChildWithDefault(self, path, request):
         if _is_authorized(request, self._get_auth_token):
             # Authorization checks out, let the protected resource do what it
             # will.
-            return self._resource.getChildWithDefault(path, request)
+            request.postpath.insert(0, request.prepath.pop())
```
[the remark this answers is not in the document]
tomprince: Done. The reason being, we want this to be transparent, and not consume paths, but the caller assumes we do. This is copied from the equivalent logic in `twisted/web/_auth/wrapper.py`.

tomprince:
> I just noticed this seems to delete all the Nix config? Is that a mistake or...?

It doesn't look like `klein` is packaged for nix. I spent some time to see if I could add it to the packaging, but didn't have any luck. I figure it isn't valuable enough to block this change.

tomprince:
> I just noticed this seems to delete all the Nix config? Is that a mistake or...?

I've switched it to just remove the nix CI configuration. I'm inclined to land this, and someone with more nix experience can get the nix configs working again.

meejah:
We'd have to lean on @exarkun for Nix help, basically ... would a pairing session (or two) pass enough Nix on to @tomprince or I so we could package a future Python dependency? (I'm thinking when we have an additional dependency or need to upgrade one).

I'm happy to land this, but I think we should at least file a bug about re-enabling Nix (and include a pointer to the deleted CI config so it's easy to find).

The last time I tried to use Nix I failed to find any way to make it work (on a Debian system .. including VMs) but of course things improve :)

hacklschorsch:
I'm a bit hesitant to stand up to this since I still feel less-than-great at Nix, but I think I could give it a try.  After having completed my over-due tasks for the "Orange" PrivateStorage milestone, which will take me probably two more weeks...

About the two sessions:  I wish you luck!  I have been wrestling with Nix since January and still don't feel very competent, but maybe you become friends with it quicker than me :)

exarkun:
I guess you should go ahead with adding the klein dep and removing the Nix packaging.  I will deal with the Nix packaging issues afterwards.

meejah (approved):
looks good, pending ticket for re-enabling Nix etc.

PR #363, opened 2021-05-26
http api layout

### LINKED ISSUE #359: HTTP API organization
See https://github.com/LeastAuthority/magic-folder/pull/332#discussion_r638115777 

This brings up a question about how to organize the HTTP API; currently there are the following endpoints:

-  `/v1/magic-folder` which response to GET and lists (all) magic-folders and their config
- `/v1/snapshot` which responds to GET and lists all snapshots in all folders
- `/v1/snapshot/<folder-name>`  which responds to POST and adds a snapshot

The comment ref'd above is for a PR which adds a `/v1/participants/<folder-name>` API which essentially follows the "snapshot" pattern.

However, maybe it makes more sense to hang some of those APIs behind a `/<folder-name>/` path piece. That is, like: `/v1/magic-folder/<folder-name>/participants` or similar.

meejah:
Fixes #359

(Note at the same time there is an exploration of whether klein is viable / easier .. that PR might be worth merging too / first / etc)

### DIFF
--- a/src/magic_folder/client.py
+++ b/src/magic_folder/client.py
@@ -122,12 +122,12 @@ class MagicFolderClient(object):
         return self._authorized_request("GET", api_url)
 
     def add_snapshot(self, magic_folder, path):
-        api_url = self.base_url.child(u'v1').child(u'snapshot').child(magic_folder)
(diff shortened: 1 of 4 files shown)


tomprince:
Closing in favour of #375.

PR #364, opened 2021-05-27
Store relative path names in snapshots.

### LINKED ISSUE #353: path-names in Snapshots are wrong

tomprince:
Fixes #353.

### DIFF
--- a/src/magic_folder/uploader.py
+++ b/src/magic_folder/uploader.py
@@ -84,7 +84,7 @@ class LocalSnapshotCreator(object):
     _stash_dir = attr.ib(validator=attr.validators.instance_of(FilePath))
 
     @inline_callbacks
-    def store_local_snapshot(self, path):
+    def store_local_snapshot(self, path, relpath):
         """
         Convert `path` into a LocalSnapshot and persist it to disk.
 
@@ -95,7 +95,7 @@ class LocalSnapshotCreator(object):
             # Query the db to check if there is an existing local
             # snapshot for the file being added.
             # If so, we use that as the parent.
-            mangled_name = magicpath.mangle_path(path)
+            mangled_name = magicpath.path2magic(relpath)
             try:
                 parent_snapshot = self._db.get_local_snapshot(mangled_name)
             except KeyError:
@@ -109,14 +109,13 @@ class LocalSnapshotCreator(object):
             # when we handle conflicts we will have to handle multiple
             # parents here (or, somewhere)
 
-            relpath_u = path.asTextMode("utf-8").path
-            action = SNAPSHOT_CREATOR_PROCESS_ITEM(relpath=relpath_u)
+            action = SNAPSHOT_CREATOR_PROCESS_ITEM(relpath=relpath)
             with action:
                 snapshot = yield create_snapshot(
                     name=mangled_name,
                     author=self._author,
                     data_producer=input_stream,
-                    snapshot_stash_dir=self._stash_dir.asBytesMode("utf-8"),
+                    snapshot_stash_dir=self._stash_dir,
                     parents=parents,
                 )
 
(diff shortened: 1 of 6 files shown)


src/magic_folder/uploader.py
```diff
@@ -84,7 +84,7 @@ class LocalSnapshotCreator(object):
     _stash_dir = attr.ib(validator=attr.validators.instance_of(FilePath))
 
     @inline_callbacks
-    def store_local_snapshot(self, path):
+    def store_local_snapshot(self, path, relpath):
         """
```
meejah, line 88: I don't really like this new API: it makes no connection between path and relpath; the caller could use "`FilePath('/etc/password')`" as the `path` and `"foo/bar"` as the `relpath` (which semantically doesn't make any sense). Since a newly-created snapshot must exist inside a particular magic-folder I think it makes more sense for the `LocalSnapshotCreator` to know its magic-folder path and keep the API the same (that is: pass in a single `FilePath`) and have this method compute the relpath piece. There will always be exactly one `LocalSnapshotCreator` for a particular magic-folder.

I see we check higher-up to report `ValueError`, but it still doesn't make sense for this piece to receive a `path` outside of its magic-directory. So basically all that boilds down to just moving the `segmentsFrom` call inside here so that the computed relpath is (still) definitely inside the magic-path.

meejah (changes requested):
One change inline, otherwise looks good.

src/magic_folder/uploader.py
tomprince: I disagree. I don't see any reason for them together. This is definitely coloured by thinking of this as a VCS, where interaction may be controlled by the user rather than automatic. Certainly we won't likely be doing anything like that for the MVP, but I think it is worthwhile keeping the distinction the in-process and stored representation (and the bits of code that interact with those) vs the on-disk representation (and the bits of code that want to interact with that).

tomprince: Partly, I'm thinking down the line when we are thinking about things like conflict resolution, we will want to deal with the remote state as distinct from the on-disk state.

One possible situation that we might run into is wanting to snapshot a path 'a/changed-file.local` with the name `a/changed-file` (without the `.local`), or maybe similarly a file that it outside the magic-folder in a temporary directory, but created by a merge tool.

meejah: I think this makes things less-clear right now.

In the future, we can easily track relative-paths separately (if need be) but as it sits everything in the uploader is tracked by a single thing: the `FilePath` of the item. We have no use-case to do the relative-path computation "early" like this. Again, all I'm suggesting is moving the `segmentsFrom` call closer to where we need it.

exarkun: I'm not sure I entirely follow this change or the discussion.  The new `relpath` isn't documented - but from the implementation change I'm guessing it is meant to be a string giving the path to the changed file relative to the top of the magic-folder directory (whereas `path` is just the absolute path to the file in the system, since it is a `FilePath`).

So having these two distinct values means you can specify any file as the data source for any snapshot - in the magic-folder or not, with the same relative/basename or not.

I don't feel very strongly one way or the other.  If you want a tie-breaker, I'd say to go with the smaller interface for now (because the functionality of unrelated relpath and path values isn't used or exposed to users anywhere yet).  If the need arises later we can spend a bit more time thinking about what use-cases will really drive such a need and what the best design to support them is.

tomprince: We need the call where it is now, to be able to return an error message in a timely fashion, moving it would put the check behind the queue.

meejah:
Closing in favour of https://github.com/LeastAuthority/magic-folder/pull/368 for now

PR #368, opened 2021-05-28
Store relative path names in snapshots.

meejah:
#364 without the `relpath` api changes


meejah:
per @exarkun's "tie breaker" comment (https://github.com/LeastAuthority/magic-folder/pull/364#discussion_r640886741) I'm merging this; we don't seem to have consensus on either the `relpath` changes in #364 nor on the "most APIs should use `FilePath`" changes in #356 and I need this bug fixed to move on with downloader etc changes and testing.

PR #377, opened 2021-06-01
Add linting for `__future__` imports.

tomprince:
This requires everything but `unicode_literals`, which will likely require some code adjustment to deal with.

Additionally, it enables flake8 linting (though it disables most of the currently failing lints.


.flake8
```diff
@@ -0,0 +1,20 @@
+[flake8]
+extend-ignore =
+    # existing errors
+    E30,E501,
+    E124,E26,E401,E402,E731
+    # flake8-future-import "__future__ import present"
```
meejah, line 6: What is `present`? (I don't see it in the docs)

meejah (approved):
Look fine.

For future PRs: I don't think these "self-referential" newsfragments fulfill the goal of requiring tickets. The idea is to have an Issue describing a bug/feature/etc (i.e. its motivation or a description of the problem etc) and then a separate PR which addresses it.

.flake8
tomprince: That is part of the error, not code, I've adjusted it to include a placeholder.

tomprince:
> For future PRs: I don't think these "self-referential" newsfragments fulfill the goal of requiring tickets. The idea is to have an Issue describing a bug/feature/etc (i.e. its motivation or a description of the problem etc) and then a separate PR which addresses it.

I'm not really sure how valuable newsfragments are before our first real release, though. It isn't like people will be looking for what changed since nothing.

meejah:
> I'm not really sure how valuable newsfragments are before our first real release, though. It isn't like people will be looking for what changed since nothing.

The point isn't about the newsfragment (or NEWS); the reason we require newsfragments is so that we write tickets describing changes separate from the changes themselves.

tomprince:
> > I'm not really sure how valuable newsfragments are before our first real release, though. It isn't like people will be looking for what changed since nothing.
> 
> The point isn't about the newsfragment (or NEWS); the reason we require newsfragments is so that we write tickets describing changes separate from the changes themselves.

I'm definitely a bit skeptical on the value of that always too. I definitely see value in that sometimes, but, I don't think a typo fix, say, benefits from a ticket. And, for this, I'd be inclined to lump it under a ticket for either python 3 porting, or pyinstaller support (or probably both), but towncrier doesn't support having multiple fragments pointing at the same issue (particularly not a `.minor` one).

meejah:
> I'm definitely a bit skeptical on the value of that always too

I think you should bring it up more widely then. I'm just trying to communicate our existing policy and why it's there.

tomprince:
> > I'm definitely a bit skeptical on the value of that always too
> 
> I think you should bring it up more widely then. I'm just trying to communicate our existing policy and why it's there.

As long as we are the only two people working on the project, I'm not sure we need a wider discussion? Or is there company policy somewhere?

meejah:
> As long as we are the only two people working on the project, I'm not sure we need a wider discussion? Or is there company policy somewhere?

We might be the only two contributing currently but I don't believe that gives us _carte blanche_ to do whatever. I'm trying to explain what the existing development policy is; if you want to change it then bring it up in a wider forum please.

PR #380, opened 2021-06-02
Run black and isort over the codebase.

tomprince:


meejah:
Overall running these tools will be good but can we please delay them until after our first milestone? (Or until we have some point with no outstanding PRs .. just to avoid disruption)

tomprince:
> Overall running these tools will be good but can we please delay them until after our first milestone? (Or until we have some point with no outstanding PRs .. just to avoid disruption)

The only open one I see that isn't mine is https://github.com/LeastAuthority/magic-folder/pull/352. I'd be happy to regenerate after that lands. Or, if you have some other branch, I'd be happy to rebase them on top of this before this lands.

meejah:
There's the downloader branch and the invites branch too (didn't check if those correspond the the draft PRs or not) and maybe some smaller ones i'm not remembering right now ... I'm mostly looking to avoid fallout / distractions in the short-term as we work to integrate with GridSync.

PR #381, opened 2021-06-02
Downloader

### LINKED ISSUE #128: download of files based on snapshots

meejah:
Fixes #128 

Documentation and implementation/u-tests of the Downloader along with some integration tests that covers the single-participant recovery case.

Supercedes #219


src/magic_folder/uploader.py
```diff
@@ -271,16 +286,21 @@ def upload_local_snapshots(self):
         # the new snapshot gets lost. Perhaps this can be solved by storing each
         # LocalSnapshot in its own row than storing everything in a blob?
         # https://github.com/LeastAuthority/magic-folder/issues/197
+
+        # XXX above comment obsolete, right? (197 is fixed / database is normalized)
```
tomprince, line 290: The comment is not obsolete. `MagicFolderConfig.get_local_snapshot` actually reads all snapshots for that path, and `MagicFolderConfig.delete_localsnapshot` deletes all snapshots for that path. So:

1. read all snapshots
2. write all snapshots that existed at point (1)
3. delete all snapshots that exist at point (3)

If snapshots are created between 1 and 3, they'll get lost. There are a couple of ways to address this:
- change `delete_localsnapshot` to only delete the snapshots passed to it (incidentally, that functions needs to be changed to delete data from related tables, probably)
- change `write_snapshot_to_tahoe` to not be recursive, and upload them one at a time, deleting from the database as we go.
I prefer the second option, even though it may be more work, as it lets us make progress, even if we can't upload all the snapshots at once.

tomprince: https://github.com/LeastAuthority/magic-folder/issues/383 for the cleanup in related tables issue

tomprince:
Reading through this (both the code and docs), I think we need some explicit documentation about our data-model, including filesystem, database and tahoe. In particular how data flows between them, and the invariants we want to maintain with them.

src/magic_folder/uploader.py
meejah: I changed the comment; just the part about "individual rows" is obsolete.

Most of the above should go into a ticket (it's related to the uploader not downloader anyway) so I clicked the "reference in new issue" button and it created #385

meejah:
> I think we need some explicit documentation about our data-model, including filesystem, database and tahoe. In particular how data flows between them, and the invariants we want to maintain with them.

Sure, that sounds useful.

meejah:
Removed "draft"; overall shape is good but still needs a few unit-tests (mostly for error-cases).

meejah:
(Not clear to me why "deprecations" fails .. on other builds I see those same deprecation warnings but no failures, so I guess it's related to the other messages...)

docs/downloader.rst
```diff
@@ -0,0 +1,169 @@
...
+.. -*- coding: utf-8 -*-
+
+.. _downloader:
+
+Downloader Operation
+====================
+
+Audience: developers working on magic-folder itself
+
+**Note:** some of this document describes features that aren't yet implemented; those are noted with (Not implemented).
```
tomprince, line 11: I've not yet read this, but it would be good to mark which parts are implemented, and which are aspirational. Partly to make seeing if the code below matches the description here.

meejah: Parts that don't say "(Not implemented)" are implemented.

src/magic_folder/downloader.py
```diff
@@ -0,0 +1,578 @@
...
+    def mark_overwrite(self, remote_snapshot, staged_content):
+        """
+        This snapshot is an overwrite. Move it from the staging area over
+        top of the existing file (if any) in the magic-folder.
+
+        :param FilePath staged_content: a local path to the downloaded
+            content.
+        """
+        local_path = self.magic_path.preauthChild(magic2path(remote_snapshot.name))
+        staged_content.moveTo(local_path)
```
tomprince, line 464: `moveTo` may involve copying all the content synchronously. We may want to avoid causing the process to block if we end up copying a multi-GB files here. (We should at least file an issue about this).

See also the comment above about race-conditions when this isn't synchronous.

docs/downloader.rst
```diff
@@ -0,0 +1,169 @@
...
+Briefly: a ``RemoteSnapshot`` is traced through its parents until a common ancestor is found.
+If the new Snapshot is a descendant of our latest Snapshot for that name, it's an overwrite.
+If it is not, there is a conflict (unless we don't yet have that name at all, then it's a creation).
+
+
+On Overwrite
+------------
+
+The ``content`` of the ``RemoteSnapshot`` is downloaded and moved into place in our Magic Folder.
+Our Personal DMD is updated to point at this Snapshot.
```
tomprince, line 100: I think we need to be explicit about the order things are done here, and how we ensure that we don't have inconsistencies or cause spurious conflicts if the process is racing creating a local snapshot.

In particular, we don't want to create a snapshot of a partially written remote file. This isn't *currently* a problem, but as I mention below, we may want to change it to avoid blocking while writing a large file, in which case it will definitely be an issue.

docs/downloader.rst
```diff
@@ -0,0 +1,169 @@
...
+
+On Conflict
+-----------
+
+The ``content`` of the ``RemoteSnapshot`` is downloaded and moved into a "conflict file" (see Leif Design) beside the conflicting content.
+The Personal DMD is **not** updated.
+Once the conflict is "resolved" then a new Snapshot is created with two parents: the latest
+Snapshot we had at conflict time and the conflicting Snapshot.
+Our Personal DMD is updated to point at this new Snapshot.
+
```
tomprince, line 115: Given that this is *not* implemented, we should describe what we currently do when we encounter a conflict.

tomprince (commented):
Some initial comments on the documentation of the downloader.

docs/downloader.rst
```diff
@@ -0,0 +1,169 @@
...
+If instead the resolution is ``mine`` then the file at ``<path>.theirs.<name>`` is deleted and a new (local) Snapshot is created (with two parents).
+
+The response is delayed until the local state tracking the new Snapshot has been created.
+
+The response code is **CREATED** and the **Content-Type** is ``application/json``.
+
+The response body follows the form of this example::
+
+  {}
```
tomprince, line 169: Not reviewed, since this isn't going to be part of v1. I wonder if there is value in including these docs, if they are not yet implemented.

docs/proposed/recovery.rst
```diff
@@ -0,0 +1,97 @@
...
+
+    $ magic-folder list --json --include-secret-information
+
+Running that command will produce a JSON ``dict`` containing one key for each Magic Folder.
+These will point to a ``dict`` containing all the information about this Magic Folder.
+The relevant keys here are:
+- ``"upload-dircap"``: our Personal DMD write-capability
+- ``"collective-dircap"``: a write-capability or read-capability to the Collective DMD (not used in this revision of this document but it is useful to save)
+
+You could choose to keep _all_ of the JSON from the ``magic-folder list`` incantation.
```
tomprince, line 55: Given my comment above about read vs write capabilities, I wonder if we should have a command and/or option to store all the relevant data with *just* the read-capabilities. And maybe also omitting the local file paths.

Will gridsync uses this, or does it do its own thing of storing that data in its own tahoe directory?

docs/downloader.rst
```diff
@@ -0,0 +1,169 @@
...
+If there are two or more parents this version is the resolution of a conflict.
+
+When reading the code, there is a method ``create_snapshot_from_capability`` which downloads a capability-string from Tahoe and returns a ``RemoteSnapshot`` instance.
+
+
+General Operation
+-----------------
+
+The ``DownloaderService`` is responsible for deciding which Snapshots to download.
+The capability strings of Snapshots to download are given to ``RemoteSnapshotCacheService`` which downloads a Snapshot and all its relevant parents into memory.
```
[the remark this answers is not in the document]
meejah: Yes, this probably needs additional error-handling to cover a case of "can't download a parent" .. I think that's best achieved as a followup?

docs/proposed/recovery.rst
```diff
@@ -0,0 +1,97 @@
...
+
+Further we assume that your new device or magic-folders instance is in ``~/magicfolder2`` and that you've correctly set this up and have the daemon running.
+
+First, we create a new Magic Folder into which we'll recover::
+
+    $ magic-folder --config ~/magicfolder2 add --author desktop --name recovered_cats --poll-interval 120 ~/favourite-kitties
+    Created magic-folder named 'recovered_cats'
+
+Next, we need to decide if we are the admin for this Magic Folder or not.
+Here we only describe the case where you are the admin and there is only a single participant in the original Magic Folder.
```
tomprince, line 89: This is unclear. I'm not sure what is meant by deciding if we are the admin for this magic folder. Whenever we create a magic-folder we are the admin for it.

docs/downloader.rst
```diff
@@ -0,0 +1,169 @@
...
+Once a Snapshot is downloaded it is passed to the ``MagicFolderUpdaterService`` which stages the actual content and determins if the change is an "overwrite" or a "conflict" and changes the local filesystem accordingly.
+
+Downloading new Snapshots from other participants ultimately causes changes to the local filesystem (in the magic-folder).
+
+
+What Snapshots to Download
+--------------------------
+
+Each configured magic-folder has a ``collective_dircap`` which is a Tahoe-LAFS Directory Capability ("dircap") for the list of participants.
```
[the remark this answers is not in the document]
meejah: Not really: we have a single method of describing the participants in a magic-folder right now: via the single Collective for that magic-folder. If we add more ways, then certainly this may need adjustment at that point.

docs/downloader.rst
```diff
@@ -0,0 +1,169 @@
...
+So, Snapshots to download are discovered by:
+
+- reading the Collective DMD
+- for each user in it:
+  - read their Personal DMD
+  - for each entry in the Personal DMD:
+    - queue the Snapshot for caching, including all parents
+    - after download, queue the Snapshot for updates
+
+Note that we don't always download *every* parent; we can stop early if a common ancestor is found.
```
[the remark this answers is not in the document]
meejah: The code already does this.

docs/downloader.rst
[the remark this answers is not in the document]
tomprince: I think we want to include a discussion of that case here, even if it is not yet implemented.

docs/downloader.rst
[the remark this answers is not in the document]
tomprince: Then I think the discussion above should explain what we are actually doing.

docs/downloader.rst
meejah: Adding another `(Not implemented)` to be more clear that we don't (yet) support resolving conflicts.

docs/proposed/recovery.rst
```diff
@@ -0,0 +1,97 @@
...
+**Note:** this document is only concerned with single-participant Magic Folders for now and does not cover recovery of other participants.
+
+
+Gathering the Information
+-------------------------
+
+In order to recover a Magic Folder on a new device you will need some information about it:
+
+- the Collective DMD (either write-capability or read-capability)
+- the Personal DMD (usually a write-capability but you could choose to only save the read-capability corresponding to it)
```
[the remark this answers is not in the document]
meejah: The point of being ambiguous here is that it's up to the integrator to decide that.

I think GridSync probably _should_ save it, since it's advertising it as "a backup" even though it only wants to use the read-capability with magic-folder. No user-data will be lost either way though.

docs/proposed/recovery.rst
```diff
@@ -0,0 +1,97 @@
...
+
+This can be gathered for all folders at once with a ``magic-folder`` command::
+
+    $ magic-folder list --json --include-secret-information
+
+Running that command will produce a JSON ``dict`` containing one key for each Magic Folder.
```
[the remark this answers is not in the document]
meejah: sure. after all, it's no necessary Python programmers doing integration..

docs/proposed/recovery.rst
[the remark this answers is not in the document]
tomprince: I am inclined to say that we should be opinionated on how people should integrate with us, as much as possible. Particularly around security sensitive things.

docs/proposed/recovery.rst
meejah: GridSync stores these data itself, somewhere. I think from magic-folders perspective it makes the most sense to dump the information we have as straight as possible. This lets integrators like GridSync do what they deem appropriate.

It could make sense at some point to add an _option_ to spit out only read-caps .. or e.g. a utility command like `magic-folder-api to-read-cap` or something. For now, no action needed until we have more guidance from at least GridSync on what it wants to store..?

docs/proposed/recovery.rst
meejah: Yes. This is trying to be clear about when there are multiple participants and proper invites (there is a PoC invites branch which also includes docs/words if you're interested).

So, you could indeed be "a participant in" a magic-folder for which you are not the admin and thus only have a read-cap for the Collective (but we don't "actually support" that yet).

docs/proposed/recovery.rst
tomprince: Well, we should perhaps have a conversation about @crwood about how we want this implemented between magic-folder and gridsync, and reflect that here.

docs/downloader.rst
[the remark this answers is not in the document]
tomprince: Well, that depends on the resolution of the issue.

docs/proposed/recovery.rst
tomprince: What impact would that have on the recovery process documented here?

It would probably make sense to:
- add some words at the beginning of the document indicating that this procedure is limited to handling the case of "...."
- if you want to have something here have something like ".. note:  When handling the multi-participant case we will need to consider .... and change ...."

In particular, don't frame it as a decision that needs to be made, but rather indicating what we might want to do in the future. It is fine if we don't know what we'll need to consider or change in the future, in which case, I'd leave the second note out entirely.

docs/downloader.rst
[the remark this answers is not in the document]
meejah: Okay, I can do that .. but also I suspect the answer is very likely to be, approximately, "we should renew leases on everything".

(Although discussion here of what to do if a parent _can't_ be downloaded would be good, because it could happen for other reasons)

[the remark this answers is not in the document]
tomprince: I'm more interested in what to do when we can't download a parent, since our discussion this morning suggested that for v1 we won't be renewing the leases on old snapshots.

I think it is reasonable to treat what we need to for renewal out-of-scope for this PR.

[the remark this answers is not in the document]
tomprince: >  "we should renew leases on everything"

If that is the answer, we may need to revisit either:
- the decision to not worry about renewing leases on old snapshots
- the design of recovery for v1 that depends on having those

docs/downloader.rst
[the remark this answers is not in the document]
meejah: It is doing what is discussed here: stopping early if it already finds the common ancestor.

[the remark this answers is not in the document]
tomprince: - We should include that above, rather than saying we get all parents and then have a note afterwards saying we don't
- We should how we find a common ancestor.

docs/downloader.rst
```diff
@@ -0,0 +1,169 @@
...
+For each one, the function ``create_snapshot_from_capability`` is used to download the capability and return a ``RemoteSnapshot`` instance.
+This ``RemoteSnapshot`` is stored in memory.
+
+Additionally, all parents of the Snapshot are also cached.
+We can stop caching parents once we find a "common ancestor"; this means a parent in the remote snapshot that matches the one in our Personal DMD.
+
+Conflict Resolution is described in :ref:`Multi-party Conflict Detection` under the Leif's Design section.
+Briefly: a ``RemoteSnapshot`` is traced through its parents until a common ancestor is found.
+If the new Snapshot is a descendant of our latest Snapshot for that name, it's an overwrite.
+If it is not, there is a conflict (unless we don't yet have that name at all, then it's a creation).
```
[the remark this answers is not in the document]
meejah: Yes. I added a sentence (basically: it's a conflict).

docs/proposed/recovery.rst
meejah: I thought I was being clear here that we only support the case where the answer is "yes, we are admin" -- changed this to a single sentence now, hopefully it's more-clear.

docs/downloader.rst
meejah: moved conflict stuff to proposal doc

docs/proposed/recovery.rst
meejah: Okay .. can we just file tickets against `magic-folder list` if what that outputs currently isn't sufficient for GridSync's needs?

docs/downloader.rst
meejah: The order is as-written there. Does it need more words?

There is no Snapshot created here, just the pointer in the Personal DMD is updated (but, yes, it shouldn't be updated until the on-disk data matches what is in that Snapshot).

tomprince: > The order is as-written there. Does it need more words?

We have pseudocode elsewhere for things that have a specific order. It wasn't obvious to me that this description in paragraph from was meant to be ordered/.

A couple of things that would be useful to have:
- where in the process we update the db, and with what
- some explanation of why this particular order. I assume it has to do with:
  - failing at any point in this process is recoverable (why, what invariants are being maintained that we can recover)
  - how it interacts with other process that may be going on at the same time, like a snapshot being taken at the same time, and the order that process takes as compared with this one mean that both maintain a consistent state.

>> In particular, we don't want to create a snapshot of a partially written remote file. This isn't currently a problem, but as I mention below, we may want to change it to avoid blocking while writing a large file, in which case it will definitely be an issue.
>
> There is no Snapshot created here, just the pointer in the Personal DMD is updated (but, yes, it shouldn't be updated until the on-disk data matches what is in that Snapshot).

What I'm talking about is if `LocalSnapshotCreator.store_local_snapshot` is running at the same time as `MagicFolderUpdaterService._process`. What I am asking about is if we are going to get a sensible result if the steps of those two functions are interleaved arbitrarily, are we going to get a reasonable result? One particular worry I have is whether we are going to get a spurious conflict because  `LocalSnapshotCreator.store_local_snapshot` creates a local snapshot of the new file data, after `MagicFolderUpdaterService._process` has written the new file, but before we've recorded any information about it.

docs/downloader.rst
```diff
@@ -0,0 +1,133 @@
...
+The entries in a user's Personal DMD are flat (no subdirectories) and point a (mangled) relative path-name to a Snapshot.
+
+So, Snapshots to download are discovered by:
+
+- reading the Collective DMD
+- for each user in it:
+  - read their Personal DMD
+  - for each entry in the Personal DMD:
+    - queue the Snapshot for caching, including all parents (see note)
+    - after download, queue the Snapshot for updates
```
tomprince, line 79: I'm curious why we handle each file of each participant-device individually, rather than for each file that is from any participant-device, handle all the remote snapshots corresponding to that file?

meejah: Is there a reason to prefer that order?

tomprince: That way, we only deal with any given file once, rather potentially multiple times, if multiple other participants have updated the file. The current order could potentially cause us to download several versions of a multi-gb file, only to overwrite it later in the process. It also means we may detect conflict in a given file later than we otherwise (after having already download one version an moved it in place, rather than before downloading any file.

meejah: Yes, I can imagine cases when there is a more-efficient order to do operations in.

I think it makes sense as a "later" / follow-up "optimization" ticket; there are other optimizations we could make and it's way harder to hit any "inefficient" scenario with just two participants anyway.

However, note that "caching the snapshot" does _not_ include downloading the content pointer; that only happens in the "do updates" phase.

tomprince: > Yes, I can imagine cases when there is a more-efficient order to do operations in.

I don't think it is *just* an optimization, but It also has implications for correctness.

> I think it makes sense as a "later" / follow-up "optimization" ticket;

I disagree. I think this has enough impact on behavior, and reasoning about whether our handling (or not) of conflicts is correct that we should deal with now. This is a blocking concern.

> it's way harder to hit any "inefficient" scenario with just two participants anyway.

We'll have (or should have) multiple participants as soon as somebody recovers twice. If we don't, they may lose data if recovery was not complete on the intermediate device.

> However, note that "caching the snapshot" does not include downloading the content pointer; that only happens in the "do updates" phase.

The relevant part isn't "cache the snapshot", but "queue the snapshot for update". We will queue multiple snapshots for the same file, if it is present in multiple remotes. It doesn't really matter which order we loop through participants, but the order we queue the snapshots we find (and how much they are chunked).

meejah: Okay, what correctness issues might there be from processing order?

docs/downloader.rst
meejah: Okay, I added words about when the database is updated.

I believe your worry won't happen: we move the file into place and update our database before doing any more async work (which is the "update our Personal DMD" step). A new LocalSnapshot can only be created before we move the file or after we've updated the database. If it happens before, we'll have the upstream parent (which is fine). If it happens right after the database update but before the Personal DMD update, we'll have our new snapshot as parent.

tomprince: > I believe your worry won't happen: [...]

We could crash between those two points, which would have similar issues. For the non-failure case, if it is important that those two steps are done synchronously (in other words, protected by the implicit global lock of running in the reactor thread), then that should be called out.

meejah: okay, I added a comment.

PR #388, opened 2021-06-04
Update towncrier version.

tomprince:
New versions allow using `towncrier.toml`, so we don't need the hacks around
`towncrier.pyproject.toml`.


meejah (approved):
:+1:

PR #389, opened 2021-06-04
Enable PEP 518 build isolation.

tomprince:


tomprince:
This appears to need pip's new resolver, which is apparently way to slow on py2.

sajith:
Ah, thank you.  I totally misunderstood the purpose of this PR. :-)

Issue #391, opened 2021-06-05
Get rid of all non `allmydata.uri` and `allmydata.interfaces` imports of `allmydata`.


tomprince:
The listed imports are for dealing with tahoe URIs, which we want to use as a library. The rest are holdovers from extracting the code base from tahoe.

tomprince:
After #392 and #393, the only remaining non-test imports will be:
- `magic_folder.cli`:
    - `from allmydata.util.encodingutil import argv_to_unicode, to_bytes,`
    - `from allmydata.client import read_config`
- `magic_folder.*.inotify`
    - I suspect these modules can be removed, or they'll be rewritten.

meejah:
Yeah, I'd not worry about the `inotify` modules for now, I think your prediction is accurate.

For the uncode-ish stuff in `cli` probably best to wait until we actually support Python3? I'd personally rather use `Click` for the CLI -- but also lets not do anything there until "release 2" or beyond.

For `read_config` it seems that's only for figuring out "shares needed" .. that's probably best gotten from some (maybe new) Tahoe API anyway. Perhaps there's already a non-config way to get it. But probably that can wait too?

(My assumption here is that we're still planning on a Python2 first release so depending on things from Tahoe is still fine short-term).

meejah:
Hmm, further to the `read_config` thing: that is really kind of a holdover, we could probably just remove the "ready enough" code since we do handle when Tahoe is offline -- and we have to handle that anyway, because our Tahoe client could fail to upload for many reasons besides "not connected to enough storage servers"...I filed https://github.com/LeastAuthority/magic-folder/issues/397 to discuss that aspect.

meejah:
I don't know that I wrote this down anywhere, but my thought is that if we can concentrate all our use of `allmydata.*` into `magic_folder.util.capabilities` that would be great. That is, all magic-folder code that needs to manipulate capabilities uses methods from there.

(At that point, we could decide to just copy in the hopefully-small amount of required Tahoe code if we wanted to eliminate the direct tahoe dependency at any moment).

PR #393, opened 2021-06-05
Use `cryptography.hazmat.primitives.constant_time.bytes_eq` instead of `allmydata.util.hashutil.timing_safe_compare`.

### LINKED ISSUE #336: Use `cryptography.hazmat.primitives.constant_time.bytes_eq` instead of `allmydata.util.hashutil.timing_safe_compare`.
The later looks slightly more complex, but should have the same observable behaviour. The main motivation is removing a dependency on `allmydata`.

The one use is in `magic_folder.web._is_authorized`.

tomprince:
Fixes #336. See #391.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,6 +86,9 @@ install_requires = [
     # We use them directly, rather than the re-exports from allmydata
     "pyutil >= 3.3.0",
 
+    # This is the version of cryptography required by tahoe-lafs
+    "cryptography >= 2.6",
+
     # last py2 release of klein
     "klein==20.6.0",
 ]
--- a/src/magic_folder/web.py
+++ b/src/magic_folder/web.py
@@ -53,9 +53,7 @@ from allmydata.uri import (
 from allmydata.interfaces import (
     IDirnodeURI,
 )
-from allmydata.util.hashutil import (
-    timing_safe_compare,
-)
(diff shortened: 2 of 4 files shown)


src/magic_folder/web.py
```diff
@@ -515,8 +513,13 @@ def _is_authorized(request, get_auth_token):
     if len(authorization) > 1:
         return False
     auth_token = get_auth_token()
-    expected = u"Bearer {}".format(auth_token).encode("ascii")
+    expected = b"Bearer {}".format(auth_token)
```
meejah, line 516: might want to leave the former version since `bytes` on python3 doesn't have a `.format()` .. no action required since we don't actually support py3 though :/

meejah (approved):
:+1:

setup.py
```diff
@@ -86,6 +86,9 @@ def read_version_py(infname):
     # We use them directly, rather than the re-exports from allmydata
     "pyutil >= 3.3.0",
 
+    # This is the version of cryptography required by tahoe-lafs
+    "cryptography >= 2.6",
```
tomprince, line 90: Add this explicitly, since we are importing it explicitly below.

exarkun (approved):

Issue #369, opened 2021-05-28
magic folders can't be named `.`, `..`, `README`, `api_token`, or `global.sqlite`.


tomprince:
This is because the name of folder is used as a sub-directory of the configuration folder.

meejah:
Hmm, yeah the `README` and other ones seem like weird limitations to have; what if we put the folder configs under `./folders/` or similar?

(Excluding `.` and `..` seems reasonable)

tomprince:
Thinking about this some more; if we are running on windows, we can't use any of `/\:\"?<>|` or have trailing `.` either.

exarkun:
> Thinking about this some more; if we are running on windows, we can't use any of `/\:\"?<>|` or have trailing `.` either.

Also CON, PRN, LPT1, etc. :)

meejah:
Hmm, does that mean we maybe need like UUIDs ("or something" .. truncated sha256 of the name?) for folder-directories and map folder-names -> those in the database?

PR #398, opened 2021-06-07
Don't leave API support for configuring state-directory name for magic-folders.

tomprince:
To address #369, I'm changing the pathname to be constructed in a specific way
from the folder name. There is no reason for this internal directory name to
be configurable.

### DIFF
--- a/newsfragments/398.minor
+++ b/newsfragments/398.minor
@@ -1,0 +1,1 @@
+Add your info here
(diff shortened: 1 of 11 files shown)


tomprince:
It turns out that macOS doesn't like some characters the test generate in filenames, so we need more of #369 to land at once.

PR #399, opened 2021-06-07
Use hashed folder name for state path.

tomprince:
See #369.

Switch to using a hashed folder name for the state path, to avoid filename collisions in our config directory, as well as issues with characters valid in filenames.


src/magic_folder/config.py
```diff
@@ -1340,9 +1341,10 @@ def create_magic_folder(self, name, magic_path, state_path, author,
             raise ValueError(
                 "'{}' does not exist".format(magic_path.path)
             )
+        state_path = self._get_state_path(name).asTextMode("utf-8")
         if state_path.asBytesMode("utf-8").exists():
             raise ValueError(
-                "'{}' already exists".format(state_path.path)
+                "magic-folder state directory {}' already exists".format(state_path.path)
```
meejah, line 1347: nit: unmatched single-quote now around the pathname

meejah (approved):
LGTM,
plus automatic state-paths and hashed names make a lot more sense.
(one inline nit).

meejah:
Might be best to avoid force-pushing on PRs where there are already comments etc; github seems to get confused by that.

tomprince:
> Might be best to avoid force-pushing on PRs where there are already comments etc; github seems to get confused by that.

I work best when keeping a clean commit history (i.e. each commit is small self-contained change). That requires force pushes.

PR #400, opened 2021-06-07
Add a YAML loader that always uses `unicode`.

tomprince:


meejah:
Could you please file a ticket explaining the motivation for this? We're only loading YAML from Tahoe, so at first it seems to me it'd be best to continue using "whatever Tahoe uses". (Does GridSync even need to use `magic-folder migrate` in the near term?)

tomprince:
> Could you please file a ticket explaining the motivation for this?

I ran into this working on #369, `migrate` is sometimes passing `bytes` to `GlobalConfigDatabase.create_magic_folder`, which expects `unicode`. I could have just changed it to conditionally convert, but I also want to use yaml for jsonschema.

tomprince:
> Does GridSync even need to use magic-folder migrate in the near term?

I'd be happy to tear out the migration support for now, instead.

meejah:
I don't think we should tear out the support; I just meant to note that we don't need to fix this _now_ if there's no immediate need from GridSync.

meejah (approved):

PR #402, opened 2021-06-07
Remove partial support for migration from tahoe-lafs magic-folder.

### LINKED ISSUE #401: Remove support for migrating from tahoe-lafs magic folders.
This is not required for the first release, and doesn't handle migrating anything other than the folder name and path.

tomprince:
Fixes #401.


meejah (changes requested):
I don't understand the motivation for removing a feature that works; see #401 for that discussion though.

PR #405, opened 2021-06-08
Klein error handling

### LINKED ISSUE #376: All error repsonse should have a JSON body.
There was code to do this originally in https://github.com/tp-la/magic-folder/pull/1
```diff
diff --git src/magic_folder/web.py src/magic_folder/web.py
--- src/magic_folder/web.py
+++ src/magic_folder/web.py
@@ -138,6 +138,9 @@ class BearerTokenAuthorization(Resource,
         return Unauthorized()
 
 
+from werkzeug.exceptions import HTTPException
+
+
 @attr.s
 class APIv1(object):
     """
@@ -153,6 +156,12 @@ class APIv1(object):
 
     app = Klein()
 
+    @app.handle_errors(HTTPException)
+    def internal_error(self, request, failure):
+        request.setResponseCode(failure.value.code or INTERNAL_SERVER_ERROR)
+        _application_json(request)
+        return json.dumps({"reason": failure.value.description}).encode("utf8")
+
     @app.route("/status")
     @inlineCallbacks
     def status(self, request):
```
But that isn't entirely correct, as `HTTPException` also encodes headers that we will need to add, if we override klein's handling of the exception. (Additionally we may want to special case `werkzeug.routing.RequestRedirect` which is a redirect not an error)

tomprince:
Fixes #376.

### DIFF
--- a/src/magic_folder/test/eliotutil.py
+++ b/src/magic_folder/test/eliotutil.py
@@ -26,6 +26,7 @@ from eliot import (
     Field,
 )
 from eliot.testing import capture_logging
+from eliot.twisted import DeferredContext
 
 from twisted.internet.defer import (
     maybeDeferred,
@@ -69,21 +70,15 @@ def eliot_logged_test(f):
         from eliot._output import _DEFAULT_LOGGER as default_logger
 
         def republish():
-            # This is called as a cleanup function after capture_logging has
-            # restored the global/default logger to its original state.  We
-            # can now emit messages that go to whatever global destinations
-            # are installed.
-
             # The memory logger captures both the raw messages and the
             # serializer. We pass them both to the global logger, so that the
             # expected serialization is preserved.
             logger = storage.logger
             for msg, serializer in zip(logger.messages, logger.serializers):
+                # We copy the message here, as the cleanup function that
+                # capture_logging installs will also try to serialize all
+                # messages.
                 default_logger.write(msg, serializer)
-
-            # And now that we've re-published all of the test's messages, we
-            # can finish the test's action.
-            storage.action.finish()
 
         @capture_logging(None)
         def run(self, logger):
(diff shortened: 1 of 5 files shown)


src/magic_folder/test/test_web.py
```diff
@@ -463,6 +472,86 @@ def test_list_folders(self, folders):
...
 
+
+class RedirectTests(SyncTestCase):
+    """
+    Test for handling redirects from werkzeug.
+    """
+    url = DecodedURL.from_text(u"http://example.invalid./v1/magic-folder")
+
+    @skipIf(
+        StrictVersion(werkzeug.__version__) < StrictVersion("1.0.0"),
```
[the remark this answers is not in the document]
tomprince: The version in the nixos snapshot we are using is < 1.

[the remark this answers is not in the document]
meejah: We should file a bug to upgrade it then .. and/or figure out some wider strategy about how to ensure Nix can track our Python requirements since I suspect this will be the case (i.e. "Nix has old version") for almost anything we want to upgrade..?

(A comment would be good here indicating that too...)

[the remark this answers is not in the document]
tomprince: #409

src/magic_folder/test/eliotutil.py
```diff
@@ -69,16 +69,14 @@ def run_and_republish(self, *a, **kw):
...
             # serializer. We pass them both to the global logger, so that the
             # expected serialization is preserved.
             logger = storage.logger
             for msg, serializer in zip(logger.messages, logger.serializers):
+                # We copy the message here, as the cleanup function that
+                # capture_logging installs will also try to serialize all
+                # messages.
                 default_logger.write(msg, serializer)
 
             # And now that we've re-published all of the test's messages, we
```
tomprince, line 82: I think the `action.finish` needs to be moved til after the default logger is restored

tomprince: Or, we need to pass an explicit logger into the action start.

src/magic_folder/test/test_web.py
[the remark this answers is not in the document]
exarkun: Maybe after the current round of priorities is out of the way, we can port this codebase to Python 3 and then a lot of Python packaging/dependency issues probably become simpler?  They certainly do wrt Nix.

src/magic_folder/test/eliotutil.py
```diff
@@ -69,22 +70,16 @@ def run_and_republish(self, *a, **kw):
         from eliot._output import _DEFAULT_LOGGER as default_logger
 
         def republish():
```
exarkun, line 72: There's a newer version of all of this Eliot/testtools integration logic in magic-folder which is both simpler and fixes the interaction with skips - which I'm guessing might have been your motivation here?

If the Tahoe version works here then perhaps it can replace most of what's going on here.  I haven't tried to understand these changes yet because I remember there are so many interactions this code tries to deal with and I'm sure it would take me a long time just to figure out what it was trying to do before, let alone what the effect of the changes are.  If we can get away with the Tahoe version then that would be nice. :)

exarkun (commented):
Thanks.  Sorry about that Eliot code. :/  Left a suggestion inline.

src/magic_folder/test/eliotutil.py
tomprince: I did have a look at the newer tahoe code, but one thing I noticed about it is that it only add logging to the test method itself, and not any of the setup/teardown/cleanup code, which is something that I was trying to avoid losing.

exarkun: :cry:

tomprince: That said, we could probably adapt some of the tahoe code (the split logger is probably better than the republishing, for example), but I'd rather getting this working now, and deal with adapting the tahoe changes later.

exarkun: Adapting later seems fine to me.  Sorry about the delayed reply.

PR #416, opened 2021-06-12
Remove tahoe windows fixups from magic-folder tests.

tomprince:
The fixups appear to do a few things:
- Call `SetErrorMode`[1]. This appears to be for errorhandling for `GetDiskFreeSpaceExW` in `allmydata.util.fileutil`.
- Various things for dealing with unicode output on the console.
- Handling unicode command-line arguments.

Given that we are only doing this in tests, I don't think any of those mater.

[1] https://docs.microsoft.com/en-ca/windows/win32/api/errhandlingapi/nf-errhandlingapi-seterrormode

See #391.

### DIFF
--- a/newsfragments/416.minor
+++ b/newsfragments/416.minor
@@ -1,0 +1,1 @@
+Add your info here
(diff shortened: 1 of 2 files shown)


meejah (approved):
lgtm

Issue #410, opened 2021-06-10
Pin all of our requirements, and automate updating them to new versions.


tomprince:
Since we are planning to distribute this as a bundled application, we should pin all of our dependencies exactly, so that we are testing exactly what we ship to users.

We should additionally automate updating those dependencies, using something like [dependabot](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/about-dependabot-version-updates)

meejah:
We should also bear in mind that the bundle is a PyInstaller bundle so we should do whatever is best for that target (cc @crwood if there's anything to keep in mind here?)

tomprince:
It looks like gridsync uses this for the other commands it bundles (see https://github.com/gridsync/gridsync/tree/master/requirements), but not yet magic-folder.

If we were to include a `requirements.txt` in our repo, gridsync could just directly consume that.

tomprince:
It looks like dependabot doesn't support python 2.7 anymore, so we won't have automatic updates until we switch to python 3.

PR #419, opened 2021-06-14
Multiple tahoe versions in integration tests

meejah:
Add a ``--tahoe-version`` option to the integration tests specifying which Tahoe version to install in a separate virtualenv for the tahoe grid (storage-servers, client, etc).

Also use that option to do 3 runs (1.14.0, 1.15.1 and master) of integration tests


integration/conftest.py
```diff
@@ -274,6 +322,7 @@ def bob(reactor, temp_dir, introducer_furl, flog_gatherer, request):
     return pytest_twisted.blockon(
         MagicFolderEnabledNode.create(
             reactor,
+            tahoe_venv,
```
tomprince, line 325: It looks like this fixture isn't requested here or below. I *think* this is actually passing the fixture function, rather than the resolved fixture.

integration/test_list.py
```diff
@@ -28,14 +28,15 @@
...
     'magic-folder list' happy-path works
     """
 
     with start_action(action_type=u"integration:test_list:zelda", include_args=[], include_result=False):
         zelda = yield util.MagicFolderEnabledNode.create(
             reactor,
+            tahoe_venv,
             request,
             temp_dir,
             introducer_furl,
```
tomprince, line 42: Could this use one of our fixtures, instead of open-coding it?

integration/conftest.py
```diff
@@ -101,10 +108,48 @@ def cleanup():
...
+# NB: conceptually, it kind of makes sense to parametrize this fixture
+# on "Tahoe version" .. a quick attempt at that lead to all but the
+# first version failing; I think something doesn't get cleaned-up
+# properly between runs "or something". Instead, we use a command-line
+# option to specify tahoe-version and have multiple separate runs
+# (which looks better in CI anyway).
+@pytest.fixture(
+    scope='session',
+)
+def tahoe_venv(request, reactor, temp_dir):
```
tomprince, line 120: This should use https://docs.pytest.org/en/latest/reference/reference.html#std-fixture-tmp_path_factory instead. Requesting `tmpdir` (or `tmp_path`) forces that fixture to be session scoped as-well, so every test would be using the same temporary directory.

I think this is probably why trying to run multiple versions in the same process was failing.

integration/conftest.py
```diff
@@ -53,6 +54,11 @@ def pytest_addoption(parser):
         "--coverage", action="store_true", dest="coverage",
         help="Collect coverage statistics",
     )
+    parser.addoption(
+        "--tahoe-version", dest="tahoe_version",
+        help="Tahoe version to install (used for storage-servers and client)",
+        default="tahoe-lafs==1.15.1",
+    )
```
tomprince, line 61: I think we have this point to a `requirements/tahoe-integration-1.15.text` file (in the root, #410 will probably put more requirements files there) or something, instead of a version string. (At least for 1.14, I think we might need to bound cryptography/pyopenssl versions; at least I did when testing against that version.) I guess `--tahoe-requirements-file` might thus make a better name?

integration/conftest.py
```diff
@@ -101,10 +108,48 @@ def cleanup():
...
+    tahoe_version = request.config.getoption("tahoe_version")
+    print("creating venv", venv_dir, sys.executable)
+    out_protocol = _DumpOutputProtocol(None)
+    reactor.spawnProcess(
+        out_protocol,
+        sys.executable,
+        ("python", "-m", "virtualenv", venv_dir),
+        env=environ,
+    )
+    pytest_twisted.blockon(out_protocol.done)
```
tomprince, line 135: Any reason we can't just use `subprocess.check_call` here?

tomprince (changes requested):

integration/util.py
```diff
@@ -326,7 +329,11 @@ def _magic_folder_runner(proto, reactor, request, other_args):
     )
 
 
-def _tahoe_runner(proto, reactor, request, other_args):
+# XXX have to get 'tahoe_venv/bin/python' or whatever all the way down
+# here to _tahoe_runner .. from I guess MagicFolderEnabledNode?
+
+
+def _tahoe_runner(proto, reactor, tahoe_venv, request, other_args):
```
tomprince, line 336: This function needs to be updated to actually use `tahoe_venv`.

integration/test_list.py
meejah: You mean "zelda" or ...? I suspect the point here was to avoid polluting the "alice" / "bob" clients with extra folders (but also, not really relevant to these changes, right?) Or did you mean something specifically about the `tahoe_venv`..?

integration/conftest.py
meejah: okay .. that could be interesting to fix (separately) if that's indeed the problem. (It almost seems better to have a command-line option anyway, so that we see 3x top-level CI runs instead of one with 3x tahoe versions "inside" it ...)

integration/test_list.py
tomprince: Oh. I guess because those are session scoped? It isn't clear why this needs its own node, but other tests don't.

integration/conftest.py
tomprince: Yeah, I think having the command line option is reasonable, I'm not suggesting removing it. We definitely want to fix this fixture, so it isn't turning the function-scoped fixture `temp_dir` into a session-scoped one.

integration/conftest.py
meejah: ...and then `pip install -r <whatever-was-in-that-option>` right? Yeah if we have to have more stuff in there makes sense

tomprince: Exactly.

integration/conftest.py
meejah: Probably not? I guess "consistency with other calls"?

integration/test_list.py
meejah: This PR is about Tahoe versions. That's not immediately clear to me either.

I'm sure there are many improvements we can make to the integration-tests (esp. since they're pretty "tahoe"-centric still, having come from there) but we can ignore that here right?

tomprince (approved):

PR #424, opened 2021-06-14
fix 423

### LINKED ISSUE #423: integration breakage
Something recent has broken integration tests, seeing this after merging `main` to downloader branch:

```
    def write_config(tahoe_cfg, config):                                                                  
        """                                                                                              
        Write a configuration to a file.                                                                  
                                                                                                          
        :param FilePath tahoe_cfg: The path to which to write the config.                                
                                                                                                         
        :param ConfigParser config: The configuration to write.                                     
                                                                                                          
        :return: ``None``                                                                           
        """                                                                                               
>       tmp = tahoe_cfg.temporarySibling()                                                                
E       AttributeError: 'unicode' object has no attribute 'temporarySibling'    
```

meejah:
Fixes #423

### DIFF
--- a/integration/util.py
+++ b/integration/util.py
@@ -24,6 +24,9 @@ from twisted.internet.defer import (
 from twisted.internet.task import (
     deferLater,
 )
+from twisted.python.filepath import (
+    FilePath,
+)
 from twisted.internet.protocol import (
     ProcessProtocol,
 )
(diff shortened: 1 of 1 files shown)


tomprince (changes requested):
I think this is only an issue against tahoe-lafs main.

Issue #210, opened 2020-07-23
Add an HTTP API to create a new magic folder


exarkun:
Accept a *POST* request to /v1/magic-folder with a JSON-encoded body describing a new magic folder.  Create that folder and return a success response.

eg:

```
POST /v1/magic-folder HTTP/1.1
Content-Type: application/json

{"name": "folder name", "local-path": "absolute local path"}

200 OK

```

meejah:
This will also need an `author-name` as well (and should optionally accept `poll-interval` to be in line with the CLI).

meejah:
hey, I assigned @tomprince because I believe you're already working on this; if not please unassign yourself :)

PR #429, opened 2021-06-16
Use in memory config in cli tests

tomprince:
This is preparation for writing the CLI test for #210.

### DIFF
--- a/src/magic_folder/list.py
+++ b/src/magic_folder/list.py
@@ -17,20 +17,15 @@ from twisted.internet.defer import (
     inlineCallbacks,
 )
 
-from .client import (
-    create_magic_folder_client,
-)
-
-
 @inlineCallbacks
-def magic_folder_list(reactor, config, http_client, output, as_json=False, include_secret_information=False):
+def magic_folder_list(reactor, config, client, output, as_json=False, include_secret_information=False):
     """
     List folders associated with a node.
 
     :param GlobalConfigDatabase config: our configuration
 
-    :param HTTPClient http_client: a treq.HTTPClient instance we can
-        use to make API requests.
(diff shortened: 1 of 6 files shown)


src/magic_folder/list.py
```diff
@@ -17,19 +17,14 @@
...
 @inlineCallbacks
-def magic_folder_list(reactor, config, http_client, output, as_json=False, include_secret_information=False):
+def magic_folder_list(reactor, config, client, output, as_json=False, include_secret_information=False):
     """
     List folders associated with a node.
 
     :param GlobalConfigDatabase config: our configuration
 
-    :param HTTPClient http_client: a treq.HTTPClient instance we can
+    :param MagicFolderClient client: a treq.HTTPClient instance we can
```
meejah, line 27: comment seems wrong now?

src/magic_folder/test/cli/test_magic_folder.py
```diff
@@ -533,22 +526,65 @@ def test_config_directory_is_file(self):
...
+                api_endpoint=Equals(u"tcp:5555"),
+                tahoe_node_directory=Equals(nodedir.path),
+            )
+        )
+
+    @inline_callbacks
+    def test_default_config_directory(self):
+        """
+        Not passing a --config loads the configuration from the default directory.
+        """
```
meejah, line 571: This test doesn't seem to match what it says here; how is this testing the default config directory?

meejah (approved):
two things inline to address

src/magic_folder/test/cli/test_magic_folder.py
tomprince: I've added a comment explaining how we are testing the default.

PR #430, opened 2021-06-16
Add a way for business logic to throw an expcetion that is propagated to the API.

tomprince:
This is a followup to #405, which allows code to throw errors that are passed to users.

### DIFF
--- a/newsfragments/430.minor
+++ b/newsfragments/430.minor
@@ -1,0 +1,1 @@
+Add your info here
--- a/src/magic_folder/common.py
+++ b/src/magic_folder/common.py
@@ -14,6 +14,8 @@ from __future__ import (
 from contextlib import contextmanager
 import unicodedata
 
+import attr
+
 
 class BadResponseCode(Exception):
     """
@@ -32,6 +34,43 @@ class BadMetadataResponse(Exception):
     An attempt to load some metadata about something received a response that
     cannot be interpreted as that metadata.
     """
+
+@attr.s(auto_exc=True, frozen=True)
+class APIError(Exception):
+    """
+    An error to be reported from the API.
+
+    :ivar reason unicode: The message to be returned as the ``reason`` key of
+       the error response.
+    :ivar code int: The HTTP status code to use for this error.
+    """
+
+    reason = attr.ib(validator=attr.validators.instance_of(unicode))
+    code = attr.ib(
+        default=None,
+        validator=attr.validators.optional(attr.validators.instance_of(int)),
+    )
+
+    @classmethod
+    def from_exception(cls, code, exception, prefix=None):
+        """
+        Return an exception with the given code, and reason from the given exception.
+
+        :param code int: The HTTP status code to use for this error.
+        :param exception Exception: The exception to get the error message from.
+        :param prefix unicode: A prefix to add to the error message.
+
+        :returns APIError: An error with the given error code and a message that is
+            ``"{prefix}: {exception message}"``.
+        """
+        if prefix is not None:
+            reason = u"{}: {}".format(prefix, exception)
+        else:
+            reason = u"{}".format(exception)
+        return cls(code=code, reason=reason)
+
+    def __str__(self):
+        return self.reason
 
 
 @contextmanager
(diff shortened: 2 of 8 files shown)


newsfragments/430.minor
```diff
@@ -0,0 +1 @@
+Add your info here
```
meejah, line 1: We should be following the dev-guide and filing tickets, _then_ doing PRs for those tickets. If you want to propose changes, please propose a change to the dev guide.

https://github.com/LeastAuthority/magic-folder/blob/main/docs/development.rst

meejah:
I see a bunch of Nix changes; are those a mistake?

newsfragments/430.minor
tomprince: This is part of #210, but split out into a separate PR to be easier to review.

tomprince:
> I see a bunch of Nix changes; are those mistakenly included in this PR?

`@attr.s(auto_exc=...)` needs a newer attrs than is in the nix snapshot we are using.

newsfragments/430.minor
meejah: that's fine, but having PRs point newsfragments at themselves isn't really the point. i get that it's annoying sometimes, but that's what we've got ...

src/magic_folder/common.py
```diff
@@ -33,6 +35,29 @@ class BadMetadataResponse(Exception):
...
+    """
+
+    reason = attr.ib(validator=attr.validators.instance_of(unicode))
+    code = attr.ib(
+        default=None,
+        validator=attr.validators.optional(attr.validators.instance_of(int)),
+    )
+
+    @classmethod
+    def from_exception(cls, code, exception, prefix=None):
```
meejah, line 51: should document the params. I assume "code" is an HTTP status code, but would be good to say that (if it's true).

src/magic_folder/test/test_web.py
```diff
@@ -986,6 +987,66 @@ def test_add_participant(self, folder_name, personal_dmd):
...
+        local_path = FilePath(self.mktemp())
+        local_path.makedirs()
+        folder_config = magic_folder_config(
+            create_local_author(author),
+            local_path,
+        )
+
+        root = create_fake_tahoe_root()
+        # put our Collective DMD into the fake root
+        root._uri.data[folder_config["collective-dircap"]] = dumps([
```
meejah, line 1007: For future reference, there's `root.add_data` and `root.add_mutable_data` that don't use internals

src/magic_folder/web.py
```diff
@@ -264,41 +280,38 @@ def add_participant(self, request, folder_name):
...
-            if not IDirnodeURI.providedBy(dmd):
-                raise _InputError("personal_dmd must be a directory-capability")
-            if not dmd.is_readonly():
-                raise _InputError("personal_dmd must be read-only")
-            personal_dmd_cap = participant["personal_dmd"]
-        except _InputError as e:
-            request.setResponseCode(http.BAD_REQUEST)
-            returnValue(json.dumps({"reason": str(e)}))
+        dmd = tahoe_uri_from_string(participant["personal_dmd"])
+        if not IDirnodeURI.providedBy(dmd):
```
meejah, line 310: I know this is just re-indented existing code, but there's `util.capabilitiies.*` methods (for future reference).

meejah (commented):
Code looks fine, but @exarkun will have to review the Nix changes (I guess maybe more Nix changes are required still though?)

src/magic_folder/test/test_web.py
tomprince: Hmm. I went and looked a the other tests, and it looks like they aren't using those either. I think there may be a mismatch between the API the fake provides, and what is useful for us. I suspect we should add some wrappers or extend it to have a more friendly API. I filed #436 to do that.

tomprince:
> will have to review the Nix changes (I guess maybe more Nix changes are required still though?

I think the nix changes are fine without review (once #432 lands so they work), since it is just bumping the versions being used to something closer to what is used on other platforms.

PR #432, opened 2021-06-16
Use cachix.

tomprince:


nix/default.nix
```diff
@@ -1,7 +1,11 @@
 # This is the main entrypoint for the magic-folder derivation.
 { pkgs ? import <nixpkgs> { } }:
 let
-  pkgs' = pkgs.extend (import ./overlays.nix);
+  tahoe-repo = (import ./tahoe-lafs-repo.nix);
+  pkgs' = pkgs.appendOverlays [
+    (import (tahoe-repo + "/nix/overlays.nix"))
+    (import ./overlays.nix)
+  ];
   callPackage = pkgs'.python27Packages.callPackage;
```
exarkun, line 9: It seems like this is beyond just "use cachix".  What's the motivation here?

tomprince: See https://github.com/LeastAuthority/magic-folder/pull/432/commits/f60d4a254c08adaefaff278bc234f20bf3b8ac0f

tomprince: I was running into trouble using a new version of attrs (for a separate pull request), because the tahoe dependencies were not picking up changes from the overlay added in this repo.

exarkun: Okay.  Please do file tickets describing the scope/goals of PRs (pre-review) in the future.  It saves reviewer time.

nix/default.nix
```diff
@@ -1,7 +1,11 @@
 # This is the main entrypoint for the magic-folder derivation.
 { pkgs ? import <nixpkgs> { } }:
 let
-  pkgs' = pkgs.extend (import ./overlays.nix);
+  tahoe-repo = (import ./tahoe-lafs-repo.nix);
+  pkgs' = pkgs.appendOverlays [
```
exarkun, line 5: I didn't actually know the difference between pkgs.extend and pkgs.appendOverlays.  I didn't find any documentation about them, either.  The implementation does not suggest there is much difference at all:

```
    # Extend the package set with zero or more overlays. This preserves                                                                                                                                            
    # preexisting overlays. Prefer to initialize with the right overlays                                                                                                                                           
    # in one go when calling Nixpkgs, for performance and simplicity.                                                                                                                                              
    appendOverlays = extraOverlays:
      if extraOverlays == []
      then self
      else import ./stage.nix (args // { overlays = args.overlays ++ extraOverlays; });

    # Extend the package set with a single overlay. This preserves                                                                                                                                                 
    # preexisting overlays. Prefer to initialize with the right overlays                                                                                                                                           
    # in one go when calling Nixpkgs, for performance and simplicity.                                                                                                                                              
    # Prefer appendOverlays if used repeatedly.                                                                                                                                                                    
    extend = f: self.appendOverlays [f];
```

My experience with overlays so far has been that they're very fragile and extremely difficult to compose so I would be completely unsurprised that despite this seeming equivalence, there is still some reason `appendOverlays` is needed here rather than `extend`.

Maybe it has to do with the order?  Here we pass [tahoe's, ours] but if use tahoe's default.nix then I think it ends up as [ours, tahoe's].  Any dependencies between what we're overriding and what tahoe is overriding could end up with different derivations as a result of this different order...

nix/tahoe-lafs.nix
```diff
@@ -1,7 +1,7 @@
 { python2Packages }:
 let
   repo = ((import ./tahoe-lafs-repo.nix) + "/nix");
-  tahoe-lafs-drv = repo + "/default.nix";
+  tahoe-lafs-drv = repo + "/tahoe-lafs.nix";
```
exarkun, line 4: I was mildly concerned about expanding the public interface of the Tahoe-LAFS Nix packaging from `default.nix` to `tahoe-lafs.nix` and `overlays.nix` but since everything is pinned I guess at worst we have to deal with upgrade challenges at some point in the future but not any surprise breakage.

nix/default.nix
tomprince: The difference is that `appendOverlays` takes a list, whereas `extend` takes a single overlay.

> Maybe it has to do with the order?

Yeah, there are two things:
- The changes to our overlay to make it compose well aren't in the tahoe overlay, so we need to apply our after the tahoe overlay.
- Some the packages we depend on directly are ones that are modified in the tahoe overlay, so we want to define our packages in the context of the tahoe overlay (I don't *think* this depends on the order of the overlays, but it depends on tahoe's overlay being applied at the top-level, rather than just for the tahoe-lafs pacakge).

exarkun (approved):

PR #439, opened 2021-06-23
438.deprecations

### LINKED ISSUE #438: "deprecations" job in CI
The "deprecations" job is producing limited value currently: 
- the "inline_callbacks" warning is bogus
- there is a warning about `readBody` which seems related to testing APIs
- we are planning to pin all dependencies anyway (so could look at deprecations once, before upgrading any deps)

meejah:
Fixes #438

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,8 +53,6 @@ workflows:
       # Other assorted tasks and configurations
       - "lint":
           <<: *DOCKERHUB_CONTEXT
-      - "deprecations":
-          <<: *DOCKERHUB_CONTEXT
       - "c-locale":
           <<: *DOCKERHUB_CONTEXT
       # Any locale other than C or UTF-8.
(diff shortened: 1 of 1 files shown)


tomprince:
If we are removing the jobs, it probably makes sense to remove the support code for them as well (in tox and misc).

meejah:
> If we are removing the jobs, it probably makes sense to remove the support code for them as well (in tox and misc).

It's probably still useful to run them by hand (e.g. when upgrading pinned dependencies)

tomprince (approved):
> > If we are removing the jobs, it probably makes sense to remove the support code for them as well (in tox and misc).
> 
> It's probably still useful to run them by hand (e.g. when upgrading pinned dependencies)

I ran into some `DeprecationsWarning`s in pytest, when switching to use `inline_callbacks` there. This prompted me to dig into this code. It turns out that the our wrapper isn't great at parsing, and  was treating a failure printed in `magic_folder.test.test_upload.RemoteSnapshotCreatorTests.test_write_snapshot_to_tahoe_fails` as a warning.

If that gets fixed, then the job reports no errors. However, it turns out that is because in CI, we use the subunit reporter. But the warnings from tests get printed by `TreeReporter`, so aren't printed in CI. Thus, *in CI*, these jobs are useless.

Thus, I think you are correct, it makes sense to keep them around for local use, and we can address fixing them in CI later.

PR #431, opened 2021-06-16
Add an HTTP API to create a new magic folder.

tomprince:
Fixes #210.

PR #441, opened 2021-06-23
Improve logging in CLI tests.

tomprince:
This adds some minor improvements to the logging in `magic_folder.test.cli`, that I wrote while working on #430 and #431.


src/magic_folder/test/cli/test_magic_folder.py
```diff
@@ -220,21 +217,6 @@ def test_list_some_json(self):
         )
 
 
-def addOutcomeDetails(testcase, outcome):
```
tomprince, line 223: This is an unused bit of code that does a similar thing via testtools, instead of eliot.

meejah (approved):

Issue #138, opened 2020-06-19
Observe File addition


vu3rdd:
Observe for File creation/modification events and push the candidate files into a work queue. Another process would empty the queue and convert these items into an in-memory representation of the local snapshot.

meejah:
There is a design: https://github.com/LeastAuthority/magic-folder/blob/main/docs/proposed/scanning-for-changes.rst

Issue #112, opened 2020-04-28
Snapshot metadata includes a creation timestamp


exarkun:
Snapshots should include a timestamp reflecting their creation time.

A suggested format for this is UTC time represented as POSIX time (seconds since epoch) but an alternate format to consider is ISO8601 datetime.

meejah:
will be fixed by #442

meejah:
Snapshots include metadata for mtime and ctime (as epoch-seconds timestamps).

PR #442, opened 2021-06-24
Scanner

### LINKED ISSUE #109: Create and upload a snapshot when a file is added to a magic-folder

meejah:
Automatic scanner for local changes.

Fixes #109 and #138 and #112

Separately, it would good to have an HTTP API to trigger a scan immediately (but not part of this PR).


src/magic_folder/scanner.py
```diff
@@ -0,0 +1,147 @@
...
+    if existing_mtime is None:
+        try:
+            existing_mtime = folder_config.get_remotesnapshot_mtime(name)
+        except KeyError:
+            existing_mtime = None
+
+    if existing_mtime is None:
+        # we have no record of this file; it must be new.
+        return True
+    return local_mtime > existing_mtime
```
tomprince, line 111: We want to compare equality here, not if the local time is newer. If they are not equal, then we should upload, even if the on-disk is older than we recorded.

src/magic_folder/config.py
```diff
@@ -211,16 +211,19 @@
         CREATE TABLE remote_snapshots
         (
             name          TEXT PRIMARY KEY, -- mangled name in UTF-8
-            snapshot_cap  TEXT              -- Tahoe-LAFS URI that represents the remote snapshot
+            snapshot_cap  TEXT,             -- Tahoe-LAFS URI that represents the remote snapshot
+            timestamp     INTEGER           -- Last-modified timestamp
```
tomprince, line 215: We also want to store at least the size, for use when scanning. Looking at old magic-folders, it stored (size, mtime, ctime), and compared for equality when seeing if it should upload a new file. Without a good reason for diverging, we should probably do the same thing.

---

However, I wonder if we want this information in the `remote_snapshots` table, which then requires some logic in the scanner to figure out if it should look at that or local snapshots. Or, instead we could have a new table, that has the state of the on-disk file as-of the last snapshot, whether that snapshot is local or remote.

---

Thinking ahead, that might save us the need to always load the entire tree of local snapshots, when we just want to look at the current one, as well.

tomprince: We may also want to store a checksum, to avoid uploading the same contents, even if a user `touch(1)`s a file. (We don't have a guarantee that the convergence secrets will match). This is something that could  be done as a followup, but it may be worth being aware of, as a thing we want to do in the future.

meejah: Filed https://github.com/LeastAuthority/magic-folder/issues/446 for "store size, probably".

No use-case for _avoiding_ an upload -- maybe the mtime is important to the user (all Tahoe will actually upload is the new Snapshot data anyway, not content). So: possible future enhancement.

New table: localsnapshot data is local, so we can just ask the file (if we even need that data).

src/magic_folder/scanner.py
meejah: it's not clear to me which one is right, so okay .. changed

src/magic_folder/config.py
tomprince: > No use-case for avoiding an upload -- maybe the mtime is important to the user (all Tahoe will actually upload is the new Snapshot data anyway, not content). So: possible future enhancement.

Only if convergence secrets match, which, we have no guarantee of. This will particularly be true once we have multi-user folders, as a user may very well be in multiple folders with disjoint participants, that don't want to coordinate convergence secrets. Also, gridsync doesn't (https://github.com/gridsync/gridsync/issues/347) preserve the convergence sequence, so touch a file in a recovered folder will cause an upload without some check like this. (But as I said in my original comment, this is something we can do as a followup.

> New table: localsnapshot data is local, so we can just ask the file (if we even need that data).

I'm not sure what you mean by this, which probably means that I explained myself poorly. 

The idea is, that the scanner only cares about the file info from the most recent snapshot of the file, whether that snapshot is:
- a local snapshot of the file that we haven't yet uploaded
- a remote snapshot of the file that this device has upload (so was previously a local snapshot here
- a remote snapshot that we downloaded from a peer and wrote to the file


My proposal is to have a table that record this data, rather than having that information spread over several tables. The idea is that this table would be analogous to git's index or hg's dirstate. This data would be updated:
- when we create a new local snapshot
- when we overwrite a file with a snapshot from a peer

and we wouldn't need to update it when we convert a local snapshot to a remote snapshot, as it would still reflect the filesystem info from when we took the snapshot.

Having thought about this a bit more, this doesn't necessarily need to be a separate table, but it would change the meaning (and perhaps the table should then have a different name). As I alluded to, this could also hold the most recent local snapshot, which would let us avoid needing to process every local snapshot for a file, everytime we are interested in the leaf one. Then this would represent the current state of the file as we know it, including on the filesystem and local and remote snapshots.

tomprince:
This is superseded by #465 and followups.

PR #443, opened 2021-06-24
per-folder status information

meejah:
Per-folder status information.

It's clear we need some sort of per-folder information. It's not clear precisely _what_ but this puts in some infrastructure for that and a couple ideas (derived from the use-cases in the existing -- incomplete -- lightweight-design).

Currently this will tell listeners about ephemeral uploader and downloader information. (Clients _should not_ be trying to remember that themselves -- but this doesn't yet tackle a "last 30 updates" type thing as also implied by the existing use-cases).


meejah:
oops, I filed #447 too

PR #444, opened 2021-06-25
Maybe fix deprecation job

tomprince:
It turns out that `misc/build_helpers/run-deprecations.py` is  very bad at parsing, and it was picking up a Failure that we were printing as a deprecation warning.


src/magic_folder/uploader.py
```diff
@@ -297,7 +297,7 @@ def upload_local_snapshots(self):
                 # Unable to reach Tahoe storage nodes because of network
                 # errors or because the tahoe storage nodes are
                 # offline. Retry?
-                print(Failure())
```
meejah, line 300: Dang, does this really fix the deprecation builder??!

(Anyway, the reason I put that there is because I lost some time on a silly syntax error that was eating...and `Exception` is of course way too broad for "tahoe did a thing wrong" anyway)

tomprince: It makes the builder green, but as I explain in https://github.com/LeastAuthority/magic-folder/pull/439#pullrequestreview-692421764, it just makes the builder useless, as it won't catch any errors.

tomprince: Yeah, we definitely need better error handling here, but this will at least log the failure, and also cause unit tests that trigger it to fail.

exarkun:
This seems low priority.  I'm putting it at the bottom of my magic-folder stack for the moment.

exarkun (approved):
Thanks.  Generally looks good.  Please file a ticket and add a news fragment (I know this PR has been hanging around for a little while so maybe I don't need to mention this).

Also, there's a failing integration test on CI.  Please check to see if that failure is related to this effort (it seems unlikely it would be) and file a ticket about it if there is not already one.

Then merge.  Thanks.

src/magic_folder/tahoe_client.py
```diff
@@ -167,8 +164,7 @@ def get_welcome(self):
         resp = yield self.http_client.get(
             self.url.add(u"t", u"json").to_uri().to_text().encode("ascii"),
         )
-        js = yield resp.content()
-        returnValue(json.loads(js.decode("utf8")))
+        returnValue((yield resp.json()))
```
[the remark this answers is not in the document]
tomprince: I was changing everything to use `.json` from `readBody` + `json.loads`; this had `.content` + `json.loads`, which I also changed for consistency.

Issue #367, opened 2021-05-27
`magic-folder run` should generate eliot logging on stdout (ant not a `magic-folder-cli.<n>.eliot` file).


tomprince:

meejah:
My understanding is that eliot is targeted at developers (perhaps that's just "our current use of eliot" though?). How will a user be expected to understand the eliot logs? Would these be intermingled with the other logging?

tomprince:
They are aimed at anybody trying to understand the system. Developers, or users, or a developer looking at logs provided by a user. Most of the interesting logging is done via eliot, so if we don't capture it somehow, then we won't have meaningful logging.

meejah:
We do capture it.

meejah:
Also, after a quick look, they're logging security-sensitive secret information so absolutely should not be shared by users with anyone else .. at least not as they sit currently.

tomprince:
>     * they should not go to stdout

Certainly on Linux they want to go to stdout, I'm not sure about OSs. Whatever starts them (which presumably usually won't be a user directly) can redirect them wherever appropriate (e.g. a systemd user unit will capture the output to the journal, or gridsync can redirect output to a file). If a user is running it in a terminal manually, I think it is likely they are trying to debug something, and so outputting to the terminal makes sense.

meejah:
I think the only change we might need immediately is to specify _where_ eliot logs should go (instead of the magic-file-name that happens now). Anything that runs magic-folder can then easily do whatever it likes with those logs. The only use-case we currently have is GridSync.

There is a lot to consider here, and I don't think we _need_ to do anything at all with Eliot logs. Their output/shape / emitted messages haven't really been considered (beyond what we inherited from tahoe, I think?) and many of the existing magic-folder commands would be no longer useful if stdout starts printing Eliot logs as well (i.e. anything _except_ `magic-folder run`). There's also the existing security problems with eliot logs as they sit.

As the only use-case for the Eliot logging is "debug / understand the system" I think we should just wait to see what (if anything) GridSync might want to do with these. For development right now does it make it significantly easier for CI / CD if the eliot log name is specified by the runner?

Or, I guess, what is the motivation / use-case for changing this stuff at all?

tomprince:
> many of the existing magic-folder commands would be no longer useful if stdout starts printing Eliot logs as well (i.e. anything except magic-folder run)

Which is why this issue is specifically only about `magic-folder run`.

[...]

PR #445, opened 2021-06-25
Capture eliot logging in integration tests.

### LINKED ISSUE #448: Collect eliot logs from magic-folder service in integration tests.
Currently, the logs are left in a bunch of ` magic-folder-cli.<pid>.eliot ` files in the current directory. This is somewhat awkward to deal with. In particular, to ensure that you are only look at the logs from the most recent run, you need to ensure you remove the old logs before starting the tests.

This would be much improved, if the eliot logs from the services were integrated into the eliot logs from the integration tests.

tomprince:
Fixes #448. See #367.

Right now, the `magic-folder` CLI command always generate eliot logs in a `magic-folder-cli.{pid}.eliot` file in the current directory. This isn't great for a couple of reason:
- it is inconsiderate to pollute the user's working directory with a bunch of log files
- it makes hard to capture the logs of service in a way that is easy to access later

This PR implements enough control over eliot logging to be able capture the logs in the integration tests logs[1], and adds support for capturing them.
- Remove the existing automatically generated log files.
  - For non-service CLI commands, we don't currently generate any logs, so we are creating a bunch of empty files.
  - For service invocations, the new options provide a better means of capturing the logs.
- Add an `--eliot-fd` option to pass a file-descriptor to write logs to. The integration tests have the service write to FD 3 to capture logs.
- Add an `--eliot-task-fields` option which causes magic-folder to create a new top-level action (with fields provided by a JSON dict provided by this option) that is the ancestor of all logs from the service. This is used by the integration tests to make navigating the logs from each node and each run of the service easier (the node is a field of the root action, and each run has its own root action containing all the logs from the run).
- The integration tests are updated to pass the new options when running the service, and to collect the output from FD 3 of the service, and feed the logs back into its own eliot logger.

[1] Currently, to inspect the logs, what I've been doing to read them is:
1. Before running the tests, delete all the existing `magic-folder-cli.*.eliot` files (or often, after running the tests and realizing I need access to the logs, deleting them and running the tests again.
2. Run the tests.
3. Look through the logs by passing a glob to eliot-tree or, if I need to differentiate between different instances of the service, using `wc -l *.eliot` to identify which logs are from services (as opposed to other CLI commands, which don't currently generate any eliot logs), and then try to find the service instance I'm interested in.

### DIFF
--- a/integration/util.py
+++ b/integration/util.py
@@ -406,6 +406,27 @@ class _DumpOutputProtocol(ProcessProtocol):
         self._out.write(data)
 
 
+@attr.s
+class EliotLogStream(object):
+    """
+    Capture a stream of eliot logs and feed it to the eliot logger.
+
+    This is intended to capture eliot log output from a subprocess, and include
+    them in the logs for this process.
+    """
+    _eliot_buffer = attr.ib(init=False, default=b"")
+
+    def data_received(self, data):
+        # We write directly to the logger, as we don't want
+        # eliot.Message to add its default fields.
+        from eliot._output import _DEFAULT_LOGGER as logger
+
+        lines = (self._eliot_buffer + data).split(b'\n')
+        self._eliot_buffer = lines.pop(-1)
+        for line in lines:
+            logger.write(json.loads(line))
+
+
 def run_service(
     reactor,
     request,
@@ -421,6 +442,8 @@ def run_service(
 
     This will start the service, and the returned deferred will fire with
     the process, once the given magic text is seeen.
+
+    This will capture eliot logs from file descriptor 3 of the process.
 
     :param reactor: The reactor to use to launch the process.
     :param request: The pytest request object to use for cleanup.
@@ -439,7 +462,8 @@ def run_service(
             protocol,
             executable,
             args,
-            path=cwd
+            path=cwd,
+            childFDs={1: 'r', 2: 'r', 3: 'r'},
         )
         request.addfinalizer(partial(_cleanup_service_process, process, protocol.exited))
         return protocol.magic_seen.addCallback(lambda ignored: process)
@@ -487,6 +511,8 @@ class _MagicTextProtocol(ProcessProtocol):
     """
     Internal helper. Monitors all stdout looking for a magic string,
     and then .callback()s on self.done and .errback's if the process exits
+
+    Also capture eliot logs from file descriptor 3, and logs them.
     """
 
     def __init__(self, magic_text):
@@ -494,6 +520,7 @@ class _MagicTextProtocol(ProcessProtocol):
         self.exited = Deferred()
         self._magic_text = magic_text
         self._output = StringIO()
+        self._eliot_stream = EliotLogStream()
         self._action = current_action()
         assert self._action is not None
 
@@ -501,6 +528,11 @@ class _MagicTextProtocol(ProcessProtocol):
         with self._action:
             Message.log(message_type=u"process-ended")
         self.exited.callback(None)
+
+    def childDataReceived(self, childFD, data):
+        ProcessProtocol.childDataReceived(self, childFD, data)
+        if childFD == 3:
+            self._eliot_stream.data_received(data)
 
     def outReceived(self, data):
         with self._action.context():
@@ -969,6 +1001,14 @@ def _run_magic_folder(reactor, request, temp_dir, name):
     ]) + [
         "--config",
         config_dir,
+        "--eliot-fd",
+        "3",
+        "--debug",
+        "--eliot-task-fields",
+        json.dumps({
+            "action_type": "magic-folder:service",
+            "node": name,
+        }),
         "run",
     ]
     action_fields = {
--- a/src/magic_folder/cli.py
+++ b/src/magic_folder/cli.py
@@ -19,6 +19,7 @@ from appdirs import (
 )
 
 
+from twisted.internet import defer
 from twisted.internet.task import (
     react,
 )
@@ -38,9 +39,6 @@ from twisted.python.filepath import (
     FilePath,
 )
 from twisted.python import usage
-from twisted.internet.defer import (
-    maybeDeferred,
-)
 
 from treq.client import (
     HTTPClient,
@@ -87,6 +85,7 @@ from .join import (
 from .service import (
     MagicFolderService,
 )
+from .util.eliotutil import maybe_enable_eliot_logging, with_eliot_options
 
 if six.PY2:
     def to_unicode(s):
@@ -514,8 +513,7 @@ def run(options):
     return service.run()
 
 
-
-
+@with_eliot_options
 class BaseOptions(usage.Options):
     optFlags = [
         ["version", "V", "Display version numbers."],
@@ -640,7 +638,6 @@ subDispatch = {
 }
 
 
-@inline_callbacks
 def dispatch_magic_folder_command(args):
     """
     Run a magic-folder command with the given args
@@ -660,10 +657,15 @@ def dispatch_magic_folder_command(args):
             print(options)
         raise SystemExit(1)
 
-    yield run_magic_folder_options(options)
+    return run_magic_folder_options(options)
 
 
-@inline_callbacks
+# If `--eliot-task-fields` is passed, then `maybe_enable_eliot_logging` will
+# start an action that is meant to be a parent of *all* logs this process
+# generatres. Since we call that function in this generator, if we used
+# `eliot.inline_callbacks` here, eliot would remove that action context from
+# it stack when when we yield to reactor.
+@defer.inlineCallbacks
 def run_magic_folder_options(options):
     """
     Runs a magic-folder subcommand with the provided options.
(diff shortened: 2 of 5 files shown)


integration/util.py
```diff
@@ -945,7 +974,8 @@ def _run_magic_folder(reactor, request, temp_dir, name):
 
     :return Deferred[IProcessTransport]: The started process.
     """
-    config_dir = join(temp_dir, "magic-daemon-{}".format(name))
+    with start_action(action_type=u"integration:run-magic-folder").context():
```
exarkun, line 977: I don't understand.  Start a never-finishing action?

Indenting `config_dir` into this block seems like it does nothing except satisfy the syntactic requirement that there be something in the block.  I think `pass` is a slightly less confusing thing to put in the block.  But using `with` at all here seems misleading.  The point of the context manager is to automatically finish the action.  This action never finishes...

If this action has a parent then its parent might finish.  That seems like it would be confusing?  If this action has no parent then will the fact that it never finishes confuse tools that interpret the logs?

Yet ... nothing interesting is ever executed in the context of this action, so all of this is irrelevant.  Can this ever have anything other than a single start message?  If not, why isn't it just a Message?

src/magic_folder/cli.py
```diff
@@ -660,10 +657,13 @@ def dispatch_magic_folder_command(args):
...
 
-    yield run_magic_folder_options(options)
+    return run_magic_folder_options(options)
 
 
-@inline_callbacks
+# We potenitally start an eliot task in this function, that we want to be
+# global in scope, the we explicitly don't use `eliot.twisted.inline_callbacks`
+# here, as the task would only be active while the function was running.
+@defer.inlineCallbacks
```
exarkun, line 666: `inline_callbacks` prevents the Eliot context stack from being corrupted by generators.  I don't understand the explanation the comment is giving for such corruption being okay.

exarkun (changes requested):
Thanks.  I left some comments inline.  Some of them are just questions.  In general, I had a lot of difficulty understanding many of the changes here.  It's not clear how they support the stated goal.  It would be nice to have a bit more explanation that ties the implementation changes together.

integration/util.py
tomprince: I think this was probably obsolete or a mis-merge from #449.

src/magic_folder/cli.py
tomprince: The idea is that `maybe_enable_eliot_logging` starts a task that we want *every* message logged by a child of. `inline_callbacks` prevents that, since, as soon as we yield to the reactor, it removes that action from the stack.

Alternatively (ideally?), we could start the task before calling the reactor, except that the way the code is structured, we don't have access to the parsed options at that stage. It isn't straightforward to reorganize the code to do that either, as various tests depend on being able to call into the CLI function without it starting the reactor.

integration/util.py
```diff
@@ -406,6 +406,24 @@ def errReceived(self, data):
         self._out.write(data)
 
 
+@attr.s
+class EliotLogStream(object):
+    """
+    Capture an eliot log stream, and feed it to the eliot logger.
```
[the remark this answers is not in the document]
tomprince: I've reworded this a bit. Hopefully it is clearer now?

src/magic_folder/util/eliotutil.py
```diff
@@ -117,31 +123,101 @@ def validator(v):
...
+
+    def to_fd(reactor):
+        return FileDestination(os.fdopen(fd, "w"))
+
+    self.setdefault("eliot-destinations", []).append(to_fd)
+
+
+def opt_eliot_task_fields(self, task_fields):
+    """
+    Wrap all logs in a task with given (JSON) fields. (for testing)
```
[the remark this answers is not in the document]
tomprince: I included a more detailed comment below, explaining what this is trying to do. Originally I had included it in the docstring, but that ended up with having a ~7 line description in the `--help` output (for something that I don't expect users will use). I am open to suggestions of better wordings (I agree the current one isn't clear), but I'm not sure how to condense it down to fit on one or two lines of usage.

src/magic_folder/cli.py
tomprince: If updated the comment here, maybe that is a little clearer?

src/magic_folder/util/eliotutil.py
```diff
@@ -117,31 +123,101 @@ def validator(v):
     return validator
 
 
+def opt_eliot_fd(self, fd):
```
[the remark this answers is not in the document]
tomprince: Yes, I did see it. I had considered using it, but decided not to. I don't recall the exact reason, but I suspect it was partly because I was unsure about the file syntax it uses, and didn't want to spend to much time thinking about it.

tomprince:
> It would be nice to have a bit more explanation that ties the implementation changes together.

I've added some more words to the PR description.

integration/util.py
exarkun: Okay.  It's still there.  Can we fix it?

src/magic_folder/cli.py
```diff
@@ -660,10 +657,15 @@ def dispatch_magic_folder_command(args):
...
         raise SystemExit(1)
 
-    yield run_magic_folder_options(options)
+    return run_magic_folder_options(options)
 
 
-@inline_callbacks
+# If `--eliot-task-fields` is passed, then `maybe_enable_eliot_logging` will
+# start an action that is meant to be a parent of *all* logs this process
+# generatres. Since we call that function in this generator, if we used
```
exarkun, line 665: s/generatres/generates/

src/magic_folder/cli.py
exarkun: The comment does a better job of explaining what's going.  What's going on seems like a hack exploiting a bug that comes from an unintentional interaction between two complex systems ... So maybe not a great thing to be doing.

The motivation makes sense.  This makes me think of `eliot.add_global_fields` which is a *bit* similar.  I'm not sure if it could be bent to accomplish the same goal here or not.

~At minimum it seems like we should file an Eliot ticket describing this use-case and asking for a stable interface implementing the behavior, then link to that ticket from here (and maybe vice versa).~  See below.

src/magic_folder/util/eliotutil.py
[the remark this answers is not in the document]
exarkun: Okay.  Please file a ticket for unifying these two related things.

src/magic_folder/util/eliotutil.py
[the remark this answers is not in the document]
exarkun: That comment helps a lot, thanks.  It makes me understand the top-level task being carefully managed by `run_magic_folder_options` and `_EliotLogging`, too.  With that in mind, here's a suggestion for what to do with this code in the future:

The parent process starts an action and either:

* passes the task_uuid to the child where the child passes it to add_global_fields.  Or
* injects it into the child's Eliot messages as they pass through `EliotLogStream`.

Either way, all of the child process's Eliot messages now appear to be children of that first action the parent process started, independent of the state of the Eliot Action context state in either process (as long as the parent takes care to not let that action finish too early, of course, but at a guess `_MagicTextProtocol.processEnded` might be a reasonable place to end that action).

This can all be follow-up work though.  I think maybe *this* is the ticket should be filed, rather than the one I suggested earlier.

exarkun (approved):
Thanks.  I think I managed a better understanding this time, the extra comments definitely helped.  A few more suggestions inline but overall looks good.  Please address those and then merge.

integration/util.py
tomprince: Hmm. I though I had.

src/magic_folder/util/eliotutil.py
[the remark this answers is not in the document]
tomprince: > What's going on seems like a hack exploiting a bug that comes from an unintentional interaction between two complex systems ... So maybe not a great thing to be doing.

I presume the motivation for these suggestions are to avoid depending on this interaction. I think a better approach to avoiding it would be to restructure our code to avoid depending on it, rather than trying to do things that have a complex interaction with eliot's tree structure. (In particular, what we are doing may be slightly fragile, but it doesn't involve maintaining any code dealing with complex data structures).

One possibility is to pass in `task.react` to `run_magic_folder_options`, and have it call it. That would allow tests to pass in something that just runs the function instead, and allow `maybe_enable_eliot_logging` to start it's task before the reactor starts. (I guess we'd still need some sort of trampoline around `run_magic_folder_options`, so it can block calling `task.react` but yield in tests).

> That comment helps a lot, thanks. It makes me understand the top-level task being carefully managed by `run_magic_folder_options` and `_EliotLogging`, too. With that in mind, here's a suggestion for what to do with this code in the future:
> 
> The parent process starts an action and either:
> 
>     * passes the task_uuid to the child where the child passes it to add_global_fields.  Or

I don't think this would do what we want; either it would collapse the tree structure in the child, or add an extra field that is ignored by the tree structure.

>     * injects it into the child's Eliot messages as they pass through `EliotLogStream`.

I think if we were to do this, we'd need to rewrite the entire tree structure as we do this. If the child doesn't know
about the action, then it will create new UUIDs for each top-level action it has. The parent task will need to track all those and generate corresponding task-level sequence number, and then rewrite it to have it's uuid, and prepend onto the task level.

src/magic_folder/util/eliotutil.py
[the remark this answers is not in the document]
tomprince: #466

src/magic_folder/util/eliotutil.py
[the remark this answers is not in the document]
tomprince: #467

Issue #440, opened 2021-06-23
/v1/status improvements


meejah:
Several upgrades to the `/v1/status` information are desired to help UIs produce better information for users.

PR #447, opened 2021-06-25
per-folder status

meejah:
Some additional status information. See also #440

We probably want even more (and/or different) information but definitely we need per-folder information. This surfaces current uploads, downloads and the 30 most-recently updated Snapshots. (And makes a small fix to downloader).

*Note:* This is based off #442 so that needs to be merged first.


meejah:
This will have to be re-done since the scanner got re-written last week w/o #442

PR #449, opened 2021-06-26
Improve eliot logging in integration tests.

tomprince:
Cleanup all the existing eliot logging in the integration tests, in preparation for capturing the logs from the services. (see #445)

The main goal here is to ensure that all eliot logs are captured in an appropriate action (in particular one that hasn't been finished, as eliot won't associate messages from after action was finished, when parsing).

This add a bunch of wrappers to the code for running services and commands, to ensure that logs for each process are consistently captured in a given action, and  that that action doesn't end until the process does.

This also puts foolscap log capturing behind a pytest option, as I think that is unlikely to be useful to magic-folder developers most of the time.

### DIFF
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -66,7 +66,7 @@ def pytest_addoption(parser):
 
 @pytest.fixture(autouse=True, scope='session')
 def eliot_logging():
-    with open("integration.eliot.json", "w") as f:
+    with open("eliot.log", "w") as f:
         to_file(f)
         yield
 
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -11,7 +11,7 @@ from pathlib2 import Path
 from time import sleep
 from os import mkdir, listdir, environ
 from os.path import join, exists
-from tempfile import mkdtemp, mktemp
+from tempfile import mkdtemp
 from functools import partial
 
 from configparser import ConfigParser
@@ -202,7 +202,7 @@ def flog_gatherer(reactor, temp_dir, flog_binary, request):
         def cleanup():
             _cleanup_tahoe_process(twistd_process, twistd_protocol.exited)
 
-            flog_file = mktemp('.flog_dump')
+            flog_file = 'integration.flog_dump'
             flog_protocol = _DumpOutputProtocol(open(flog_file, 'w'))
             flog_dir = join(temp_dir, 'flog_gather')
             flogs = [x for x in listdir(flog_dir) if x.endswith('.flog')]
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -64,6 +64,11 @@ def pytest_addoption(parser):
         help="A tox env to run tahoe from.",
         default="tahoe1_15",
     )
+    parser.addoption(
+        "--gather-foolscap-logs", action="store_true", dest="gather_foolscap_logs",
+        help="Gather foolscap logs from tahoe processes.",
+        default=False,
+    )
 
 @pytest.fixture(autouse=True, scope='session')
 def eliot_logging():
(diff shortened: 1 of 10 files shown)


tomprince:
Note that there are a lot of changes in this PR, but they are split into relatively small commits, that can be reviewed on their own.

tomprince:
@exarkun I'm asking you for a review on this, since I think meejah is out next week.

tomprince:
The macOS integration failure is an intermittent issue that hopefully this and #445 will help debug.

integration/conftest.py
```diff
@@ -151,78 +159,74 @@ def tahoe_venv(request, reactor):
...
 
 @pytest.fixture(scope='session')
-@log_call(
-    action_type=u"integration:introducer",
-    include_args=["temp_dir", "flog_gatherer"],
-    include_result=False,
-)
 def introducer(reactor, tahoe_venv, temp_dir, flog_gatherer, request):
-    config = '''
+    with start_task(action_type=u"integration:introducer").context():
```
exarkun, line 228: I can't recall a time when I've used `start_task` instead of `start_action`.  I understand that the difference is that `start_task` will not search out a parent from the Eliot context: it will always be a new top-level action?  So "running the introducer" is a top-level action regardless of the Eliot context in which this fixture is invoked?

I'm not sure how to reason about when it makes sense to start a new top-level action (apart from there being no other active action) and when it makes sense to respect the current Eliot context.  Do you have a set of rules for this you can share?  Or maybe just a comment about this specific case.

integration/conftest.py
```diff
@@ -231,52 +235,43 @@ def introducer(reactor, tahoe_venv, temp_dir, flog_gatherer, request):
...
+            )
+            pytest_twisted.blockon(done_proto.done)
+
+        # over-write the config file with our stuff
+        with open(join(intro_dir, 'tahoe.cfg'), 'w') as f:
+            f.write(config)
+
+        # on windows, "tahoe start" means: run forever in the foreground,
+        # but on linux it means daemonize. "tahoe run" is consistent
+        # between platforms.
```
exarkun, line 264: the code that is directly responsible for choosing a tahoe subcommand seems to be somewhere else now.  also "tahoe start" has been removed.  Maybe this comment has outlived its usefulness.

integration/conftest.py
```diff
@@ -231,52 +235,43 @@ def introducer(reactor, tahoe_venv, temp_dir, flog_gatherer, request):
...
-        ),
-    )
-    request.addfinalizer(partial(_cleanup_tahoe_process, transport, protocol.exited))
-
-    pytest_twisted.blockon(protocol.magic_seen)
-    return TahoeProcess(transport, intro_dir)
 
 
 @pytest.fixture(scope='session')
-@log_call(action_type=u"integration:introducer:furl", include_args=["temp_dir"])
```
exarkun, line 279: Is the net of this PR that the Eliot log no longer contains the introducer furl?  Previously `log_call` would log the return value of this function now it doesn't.  Does something else?

integration/test_add.py
```diff
@@ -4,46 +4,22 @@
...
-            "--config", alice.magic_config_directory,
-            "add",
-            "--name", "test",
-            "--author", "laptop",
-            alice.magic_directory,
-        ],
+    yield alice.add(
+        "test",
+        alice.magic_directory,
+        author="laptop",
```
exarkun, line 19: This is so much nicer.  Thanks.

newsfragments/449.minor
```diff
@@ -0,0 +1 @@
+Add your info here
```
exarkun, line 1: Ticket please.

integration/util.py
```diff
@@ -80,6 +79,8 @@ class MagicFolderEnabledNode(object):
     temp_dir = attr.ib()
     name = attr.ib()
 
+    action = attr.ib()
```
exarkun, line 82: This could use some docs.  It took me a couple tries to understand the `request.addfinalizer(action.finish)` bits below but now I see it is related to this attribute.

exarkun (approved):
Thanks.  Seems like a good step overall.  Some comments inline, nothing major I think.  Please address to your satisfaction and then merge.

integration/conftest.py
tomprince: Mostly the reason it is this way is that I played around with various options, looking at the eliot-tree output of test runs with them.
- Partly what is going on is that fixtures are run outside the context of the thing requesting them, particularly session fixtures. One iteration did have a session-scope auto-use fixture that did have a task, that all other session scoped fixtures were child actions of, but that didn't make it easier to read the logs. I *do* have that for tests, which does work better (though I'm not sure there are fest (i.e. function) scoped fixtures that would capture logs for.
- If you have multiple, mostly unrelated fixtures, that each have an action wrapping them, then each action is going to be a child of the previous one, but the nesting is going to be somewhat arbitrary, and makes the logs harder to read. (One particular example of confusing nesting, in a slightly different context, is that the magic-folder (stdout) logs from restarting the service where in the test that restarted them, even if it kept running into a later test).

integration/conftest.py
tomprince: Yep. The comment was moved there, but this instance wasn't removed (there were originally two copies of this comment).

integration/conftest.py
tomprince: I think so.

integration/util.py
tomprince: Not exactly, but close enough. I've added some docs for this attribute and that call.

PR #451, opened 2021-06-28
Store the stat info of the current snapshot.

### LINKED ISSUE #450: Store the current file state of the current snapshot.
In order to detect changes to local files (see #138), we need to know what the most recent state of the file is.

This is typically done by recording the mtime, ctime and file size of each file, and then checking if the file on disk has different value than we recorded.

To that end, we add those values to the database, for the current snapshot, and add code to update those values when we write a new local snapshot, or overwrite a file with remote snapshot.

tomprince:
Fixes #450.

### DIFF
--- a/src/magic_folder/api_cli.py
+++ b/src/magic_folder/api_cli.py
@@ -116,9 +116,14 @@ def dump_state(options):
             q.extend(s.parents_local)
         print("    {}: {}".format(snap_name, parents[:-4]), file=options.stdout)
     print("  remote snapshots:", file=options.stdout)
-    for snap_name in config.get_all_remotesnapshot_paths():
-        snap = config.get_remotesnapshot(snap_name)
-        print("    {}: {}".format(snap_name, snap), file=options.stdout)
+    for snap_name in config.get_all_snapshot_paths():
+        try:
+            cap = config.get_remotesnapshot(snap_name)
+        except KeyError:
+            cap = None
+        path_state = config.get_currentsnapshot_pathstate(snap_name)
+        print("    {}: {}".format(snap_name, cap), file=options.stdout)
+        print("    {}  {}".format(" "*len(snap_name), path_state), file=options.stdout)
 
 
 class AddParticipantOptions(usage.Options):
--- a/src/magic_folder/downloader.py
+++ b/src/magic_folder/downloader.py
@@ -465,8 +465,24 @@ class LocalMagicFolderFilesystem(object):
             local_path.moveTo(tmp)
         mtime = remote_snapshot.metadata["modification_time"]
         os.utime(staged_content.path, (mtime, mtime))
-        path_state = get_pathinfo(staged_content).state
+
+        # We want to get the path state of the file we are about to write.
+        # Ideally, we would get this from the staged file. However, depending
+        # on the operating system, the ctime can change when we rename.
+        # Instead, we get the path state before and after the move, and use
+        # the later if the mtime and size match.
+        staged_path_state = get_pathinfo(staged_content).state
         staged_content.moveTo(local_path)
+        path_state = get_pathinfo(local_path).state
+        if (
+            staged_path_state.mtime_ns != path_state.mtime_ns
+            or staged_path_state.size != path_state.size
+            or staged_path_state.ctime_ns > path_state.ctime_ns
+        ):
+            # The path got changed after we moved it, return the staged
+            # path_state so it is detected as a change.
+            path_state = staged_path_state
+
         if tmp is not None:
             # FIXME: We should verify that this moved file has
             # the same (except ctime) state as the the previous snapshot
--- a/src/magic_folder/downloader.py
+++ b/src/magic_folder/downloader.py
@@ -9,6 +9,7 @@ from __future__ import (
     unicode_literals,
 )
 
+import os
 from collections import deque
 from functools import wraps
 
@@ -41,6 +42,7 @@ from twisted.python.filepath import (
 from twisted.internet.defer import (
     DeferredLock,
     returnValue,
+    succeed,
 )
 from twisted.web.error import (
     SchemeNotSupported,
@@ -55,6 +57,7 @@ from .magicpath import (
 from .snapshot import (
     create_snapshot_from_capability,
 )
+from .util.file import PathState, get_pathinfo, seconds_to_ns
 
 def _exclusively(f):
     @wraps(f)
@@ -221,6 +224,8 @@ class IMagicFolderFilesystem(Interface):
 
         :param FilePath staged_content: a local path to the downloaded
             content.
+
+        :return PathState: The path state of the file that was written.
         """
 
     def mark_conflict(remote_snapshot, staged_content):
@@ -388,7 +393,9 @@ class MagicFolderUpdater(object):
                     self._magic_fs.mark_conflict(snapshot, staged)
                     # FIXME note conflict internally
                 else:
-                    self._magic_fs.mark_overwrite(snapshot, staged)
+                    # FIXME: This should take the path and mtime as arguments,
+                    # instead of the snapshot argument
+                    path_state = self._magic_fs.mark_overwrite(snapshot, staged)
 
                     # Note, if we crash here (after moving the file into place
                     # but before noting that in our database) then we could
@@ -398,7 +405,9 @@ class MagicFolderUpdater(object):
 
 
                     # remember the last remote we've downloaded
-                    self._config.store_remotesnapshot(snapshot.name, snapshot)
+                    self._config.store_downloaded_snapshot(
+                        snapshot.name, snapshot, path_state
+                    )
 
                     # XXX careful here, we still need something that makes
                     # sure mismatches between remotesnapshots in our db and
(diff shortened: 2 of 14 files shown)


src/magic_folder/downloader.py
```diff
@@ -442,15 +446,41 @@ def mark_overwrite(self, remote_snapshot, staged_content):
...
         local_path = self.magic_path.preauthChild(magic2path(remote_snapshot.name))
         tmp = None
         if local_path.exists():
-            tmp = local_path.temporarySibling(b".snap")
+            # Note that the scanner won't discover this brief temp-file
+            # because Twisted only runs one thing at a time (and we don't
+            # yield between creation and deletion)
+            tmp = local_path.temporarySibling(b".snaptmp")
             local_path.moveTo(tmp)
+        # FIXME: pass this (and snapshot name?) in
```
exarkun, line 460: pass it in to what?

src/magic_folder/snapshot.py
```diff
@@ -525,7 +528,7 @@ def create_snapshot(name, author, data_producer, snapshot_stash_dir, parents=Non
     finally:
         os.close(temp_file_fd)
 
-    now = time.time()
+    now = modified_time or int(time.time())
```
exarkun, line 531: Contrary to documentation (and surprising) if `modified_time` is 0.

src/magic_folder/test/strategies.py
```diff
@@ -439,3 +442,14 @@ def local_snapshots():
...
         identifier=uuids(),
     )
+
+def path_states():
+    """
+    Build ``PathState`` instances.
+    """
+    return builds(
+        PathState,
+        mtime_ns=integers(min_value=0, max_value=2 ** 31 - 1),
```
exarkun, line 452: `mtime_ns` can probably be larger than 2 ** 31 - 1, can't it?  By a factor of 1,000,000,000?

src/magic_folder/test/test_config.py
```diff
@@ -547,20 +548,51 @@ def setup_example(self):
...
         """
         The capability for a ``RemoteSnapshot`` added with
         ``MagicFolderConfig.store_remotesnapshot`` can be read back with
         ``MagicFolderConfig.get_remotesnapshot``.
         """
-        self.db.store_remotesnapshot(snapshot.name, snapshot)
-        loaded = self.db.get_remotesnapshot(snapshot.name)
+        self.db.store_remotesnapshot(snapshot.name, snapshot, path_state)
+        capability = self.db.get_remotesnapshot(snapshot.name)
+        db_path_state = self.db.get_currentsnapshot_pathstate(snapshot.name)
```
exarkun, line 561: Seems a bit unfortunate that the symmetry between `store_remotesnapshot` and `get_remotesnapshot` has been broken here and now it takes two Python calls and two SQL statements to get back out what was put in with just one of each.  Maybe this is worth considering a follow-up refactoring to deal with.

src/magic_folder/test/test_config.py
```diff
@@ -547,20 +548,51 @@ def setup_example(self):
...
+            Equals((snapshot.capability, path_state))
         )
 
+    @given(
+        remote_snapshots(),
+        path_states(),
+    )
+    def test_remotesnapshot_with_existing_state(self, snapshot, path_state):
+        """
+        A ``RemoveSnapshot`` can be add without path state, if existing path
```
exarkun, line 573: s/add/added/

src/magic_folder/test/test_config.py
```diff
@@ -547,20 +548,51 @@ def setup_example(self):
...
+
+    @given(
+        remote_snapshots(),
+    )
+    def test_store_remote_without_state(self, snapshot):
+        """
+        Calling :py:`MagicFolderConfig.store_remotesnapshot` without a path
+        state, when there isn't already corresponding path state fails.
+        """
+        with ExpectedException(Exception):
```
exarkun, line 593: Maybe `store_remotesnapshot` can document this exceptional case.  It might also be nice to raise a distinctive exception type for it rather than using the generic Exception.

src/magic_folder/test/test_config.py
```diff
@@ -571,6 +603,21 @@ def test_remotesnapshot_not_found(self, path):
...
+            self.db.get_currentsnapshot_pathstate(path)
+
+    @given(
+        path_segments(),
+        path_states(),
+    )
+    def test_remotesnapshot_not_found_with_state(self, path, path_state):
+        """
+        ``MagicFolderConfig.get_remotesnapshot`` raises ``KeyError`` if there is
+        no known remote snapshot for the given path, but path state for it.
```
exarkun, line 616: This docstring doesn't seem grammatical to me.

src/magic_folder/api_cli.py
```diff
@@ -116,9 +116,14 @@ def dump_state(options):
...
-        snap = config.get_remotesnapshot(snap_name)
-        print("    {}: {}".format(snap_name, snap), file=options.stdout)
+    for snap_name in config.get_all_snapshot_paths():
+        try:
+            cap = config.get_remotesnapshot(snap_name)
+        except KeyError:
+            cap = None
+        path_state = config.get_currentsnapshot_pathstate(snap_name)
+        print("    {}: {}".format(snap_name, cap), file=options.stdout)
+        print("    {}  {}".format(" "*len(snap_name), path_state), file=options.stdout)
```
exarkun, line 126: I suspect the primary output format should be something machine-parseable.  JSON, presumably.  Clearly that's not the prevailing convention with the existing CLI implementation so nothing actionable here for now.

Also, I thought the CLI was going to be a thin wrapper around an HTTP API.  It seems like most of the CLI is.  Maybe that's in future plans for `dump_state`?

exarkun (approved):
Thanks.  Various comments inline.  Please address and merge.

src/magic_folder/api_cli.py
tomprince: I'm not entirely sure what the plan for `dump-state` is, I'm just adding the additional data we are storing.

I suspect what we really want to do is change it to be by path, rather than remote and current snapshots as separate things.

src/magic_folder/downloader.py
```diff
@@ -442,15 +446,41 @@ def mark_overwrite(self, remote_snapshot, staged_content):
...
         """
         local_path = self.magic_path.preauthChild(magic2path(remote_snapshot.name))
         tmp = None
         if local_path.exists():
-            tmp = local_path.temporarySibling(b".snap")
+            # Note that the scanner won't discover this brief temp-file
+            # because Twisted only runs one thing at a time (and we don't
+            # yield between creation and deletion)
+            tmp = local_path.temporarySibling(b".snaptmp")
             local_path.moveTo(tmp)
```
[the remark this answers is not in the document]
tomprince: I haven't thought through this code in detail, other than to realize that we should do https://github.com/LeastAuthority/magic-folder/issues/454. I got push-back in the review when I suggested improving how we write files, and it landed as-is.

src/magic_folder/downloader.py
```diff
@@ -388,28 +391,29 @@ def _process(self, snapshot):
                     self._magic_fs.mark_conflict(snapshot, staged)
                     # FIXME note conflict internally
                 else:
-                    self._magic_fs.mark_overwrite(snapshot, staged)
```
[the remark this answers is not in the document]
tomprince: I'm not sure the history of still having this function (or why it was the cause for a rewrite), but #454 covers improving the situation here.

src/magic_folder/test/test_config.py
tomprince: You may be right. I've linked this comment to https://github.com/LeastAuthority/magic-folder/issues/454#issuecomment-870886295, mentioning that we may want to get these values back atomically.

I didn't implement it here, since there were no non-test consumers that cared about having both, but we probably will want to when change downloading to take this new information into account.

src/magic_folder/downloader.py
```diff
@@ -442,15 +446,41 @@ def mark_overwrite(self, remote_snapshot, staged_content):
...
+
+        :return PathState:
         """
         local_path = self.magic_path.preauthChild(magic2path(remote_snapshot.name))
         tmp = None
         if local_path.exists():
-            tmp = local_path.temporarySibling(b".snap")
+            # Note that the scanner won't discover this brief temp-file
+            # because Twisted only runs one thing at a time (and we don't
+            # yield between creation and deletion)
```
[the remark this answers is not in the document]
tomprince: I've updated the comment to point here.

src/magic_folder/downloader.py
```diff
@@ -489,18 +529,29 @@ class InMemoryMagicFolderFilesystem(object):
 
     def __init__(self):
         self.actions = []
+        self._staged_content = {}
```
meejah, line 532: What is this for? Nothing ever deletes entries in it

tomprince: This is to ensure that the object passed to `mark_overwrite` matches the object returned from the corresponding `download_content_to_staging`. It is a mapping from `id(remote_snapshot)` (since `RemoteSnapshot`'s are not hashable) to marker objects.

meejah: Nothing ever deletes from it though.

tomprince: I'm not sure why you think it is obvious that something should. After thinking about if for a while, I guess it would make sense to remove the entry in `mark_overwrite`/`mark_conflict`, to allow verifying that we don't pass the same file twice. That would catch errors that the current code doesn't. However, as a test-double , the lack of that doesn't break anything, and it is checking more than previously.

I think I originally added this because I at one point, I had `download_content_to_staging` returning a tuple, and wanted to ensure that the right bit of it got passed to `mark_overwrite`, other ways of misusing it were not on my mind. 

In any case, I think the shape of this abstraction may need to be changed for #454 (I think the code that touches the filesystem needs to be more involved with deciding whether there is a conflict), so I think we can live with not catching the additional error in the double.

src/magic_folder/test/strategies.py
meejah: What's the motivation to store nanoseconds (instead of e.g. seconds?)

tomprince: The idea is to notice when the file has changed, by storing its metadata. To detect as many changes as possible, we shouldn't throw away any data provided by the OS. I picked ns based on what old magic folders did, but revisiting this, it looks like unix tends to have ns resolution, and windows has 100 ns resolution, so that choice seems reasonable.

Storing less accuracy than is available to us means that there are changes that we won't notice, because the metadata is the same after our rounding.

Issue #320, opened 2021-05-04
Add/remove services on the fly


meejah:
Once #317 and #318/#319 are done, the `IService` that "runs" each magic-folder should be added / removed on the fly when the HTTP API is engaged.

The net result being that a re-start is _not_ required to see immediate effects of adding or removing a folder.

PR #452, opened 2021-06-29
Don't require a restart to add a new magic-folder.

tomprince:
This is a followup to #210, that implements the add portion of #320.


src/magic_folder/service.py
```diff
@@ -195,3 +196,58 @@ def stopService(self):
...
+        if name in self.config.list_magic_folders():
+            raise APIError(
+                code=http.CONFLICT,
+                reason="Already have a magic-folder named '{}'".format(name),
+            )
+
+        # create our author
+        author = create_local_author(author_name)
+
+        # create an unlinked directory and get the dmd write-cap
```
exarkun, line 227: s/dmd/collective/

exarkun (approved):
Thanks.  Nice unification of creation w/ the rest of the service-based config management code.  Please fix lint CI job and one minor inline comment and then merge.

PR #463, opened 2021-07-01
Add an endpoint to request a scan of a magic folder.

### LINKED ISSUE #333: test_web should use `create_testing_configuration` instead of `create_global_configuration`
That way, and in-memory database is used for the global configuration.

(Unfortunately, `ListMagicFolderTests` uses `load_global_configuration` to access the created configuration, otherwise this could be a trivial replacement)

tomprince:
See #138. 

This is a followup to #465 that adds an API endpoint and corresponding CLI command to request a scan of a folder. It also parameterizes the existing synchronization integration tests, to run them with `add-snapshot` and `scan-folder`.

This also incidentally fixes #333 as part of refactoring some code for the tests.

All the commits are self-contained, and so could be reviewed individually.

### DIFF
--- a/integration/test_synchronize.py
+++ b/integration/test_synchronize.py
@@ -35,9 +35,19 @@ def add_snapshot(node, folder_name, path):
     """
     Take a snapshot of the given path in the given magic folder.
 
-    :param MagigFolderEnabledNode node: The node on which to take the snapshot.
+    :param MagicFolderEnabledNode node: The node on which to take the snapshot.
     """
     return node.add_snapshot(folder_name, path)
+
+
+def scan_folder(node, folder_name, path):
+    """
+    Scan the given magic folder. This should cause the given path to be
+    snapshotted.
+
+    :param MagicFolderEnabledNode node: The node on which to do the scan.
+    """
+    return node.scan_folder(folder_name)
 
 
 def periodic_scan(node, folder_name, path):
@@ -46,7 +56,7 @@ def periodic_scan(node, folder_name, path):
     This should cause the given path to be
     snapshotted.
 
-    :param MagigFolderEnabledNode node: The node on which to do the scan.
+    :param MagicFolderEnabledNode node: The node on which to do the scan.
     """
     Message.log(message_type="integration:wait_for_scan", node=node.name, folder=folder_name)
     time.sleep(1)
@@ -66,6 +76,7 @@ def enable_periodic_scans(magic_folder_nodes, monkeypatch):
 @pytest.fixture(
     params=[
         add_snapshot,
+        scan_folder,
         pytest.lazy_fixture('periodic_scan'),
     ]
 )
--- a/src/magic_folder/api_cli.py
+++ b/src/magic_folder/api_cli.py
@@ -18,9 +18,6 @@ from autobahn.twisted.websocket import (
     create_client_agent,
 )
 
-from twisted.python.filepath import (
-    FilePath,
-)
 from twisted.python import usage
 from twisted.internet.defer import (
     maybeDeferred,
@@ -28,15 +25,12 @@ from twisted.internet.defer import (
 )
 
 from .cli import (
-    _default_config_path,
-    load_global_configuration,
+    BaseOptions,
     to_unicode,
 )
 from .client import (
     CannotAccessAPIError,
     MagicFolderApiError,
-    create_http_client,
-    create_magic_folder_client,
 )
 from .util.eliotutil import maybe_enable_eliot_logging, with_eliot_options
 
@@ -61,7 +55,7 @@ def add_snapshot(options):
     Add one new Snapshot of a particular file in a particular
     magic-folder.
     """
-    res = yield options.parent.get_client().add_snapshot(
+    res = yield options.parent.client.add_snapshot(
         options['folder'].decode("utf8"),
         options['file'].decode("utf8"),
     )
@@ -147,7 +141,7 @@ def add_participant(options):
     """
     Add one new participant to an existing magic-folder
     """
-    res = yield options.parent.get_client().add_participant(
+    res = yield options.parent.client.add_participant(
         options['folder'].decode("utf8"),
         options['author-name'].decode("utf8"),
         options['personal-dmd'].decode("utf8"),
@@ -174,7 +168,7 @@ def list_participants(options):
     """
     List all participants in a magic-folder
     """
-    res = yield options.parent.get_client().list_participants(
+    res = yield options.parent.client.list_participants(
         options['folder'].decode("utf8"),
     )
     print("{}".format(json.dumps(res, indent=4)), file=options.stdout)
@@ -229,57 +223,11 @@ def monitor(options):
 
 
 @with_eliot_options
-class MagicFolderApiCommand(usage.Options):
+class MagicFolderApiCommand(BaseOptions):
     """
     top-level command (entry-point is "magic-folder-api")
     """
-    stdin = sys.stdin
-    stdout = sys.stdout
-    stderr = sys.stderr
-    _client = None  # initialized (at most once) in get_client()
     _websocket_agent = None  # initialized (at most once) in get_websocket_agent()
-
-    optFlags = [
-        ["version", "V", "Display version numbers."],
-    ]
-    optParameters = [
-        ("config", "c", _default_config_path,
-         "The directory containing configuration"),
-    ]
-
-    _config = None  # lazy-instantiated by .config @property
-
-    @property
-    def _config_path(self):
-        """
-        The FilePath where our config is located
-        """
-        return FilePath(self['config'])
-
-    @property
-    def config(self):
-        """
-        a GlobalConfigDatabase instance representing the current
-        configuration location.
-        """
-        if self._config is None:
-            try:
-                self._config = load_global_configuration(self._config_path)
-            except Exception as e:
-                raise usage.UsageError(
-                    u"Unable to load '{}': {}".format(self._config_path.path, e)
-                )
-        return self._config
-
-    def get_client(self):
-        if self._client is None:
-            from twisted.internet import reactor
-            self._client = create_magic_folder_client(
-                reactor,
-                self.config,
-                create_http_client(reactor, self.config.api_client_endpoint),
-            )
-        return self._client
 
     def get_websocket_agent(self):
         if self._websocket_agent is None:
--- a/src/magic_folder/api_cli.py
+++ b/src/magic_folder/api_cli.py
@@ -174,6 +174,25 @@ def list_participants(options):
     print("{}".format(json.dumps(res, indent=4)), file=options.stdout)
 
 
+class ScanFolderOptions(usage.Options):
+    optParameters = [
+        ("folder", "n", None, "Name of the magic-folder participants to scan", to_unicode),
+    ]
+
+    def postOptions(self):
+        required_args = [
+            ("folder", "--folder / -n is required"),
+        ]
+        for (arg, error) in required_args:
+            if self[arg] is None:
+                raise usage.UsageError(error)
+
+def scan_folder(options):
+    return options.parent.client.scan_folder(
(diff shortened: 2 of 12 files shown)


src/magic_folder/test/fixtures.py
```diff
@@ -179,3 +186,95 @@ def _setUp(self):
...
+    @classmethod
+    def create(
+        cls,
+        reactor,
+        basedir,
+        auth_token=None,
+        folders=None,
+        start_folder_services=False,
+        tahoe_client=None,
+    ):
```
tomprince, line 205: This function is taken from `magic_folder.test.test_web.treq_for_folders`, but returns an object that has more than just the HTTP client as attributes.

src/magic_folder/api_cli.py
```diff
@@ -220,62 +234,17 @@ def monitor(options):
     return Deferred()
 
 
-class MagicFolderApiCommand(usage.Options):
+class MagicFolderApiCommand(BaseOptions):
```
tomprince, line 237: The logic I removed here was almost identical to the main cli command one (with slightly different names). By sharing the logic here, it makes it easier to share the logic in `magic_folder.test.cli.common` for running commands in process. (Particularly the logic around setting config and http_clients)

src/magic_folder/cli.py
```diff
@@ -553,11 +557,13 @@ def config(self):
...
-        if self._http_client is None:
-            from twisted.internet import reactor
-            self._http_client = create_http_client(reactor, self.config.api_client_endpoint)
         if self._client is None:
             from twisted.internet import reactor
+
+            if self._http_client is None:
+                self._http_client = create_http_client(
+                    reactor, self.config.api_client_endpoint
+                )
```
tomprince, line 566: Since this is only used if we don't already have a client, there is no need to create one if we do have a client.

I had made this change, because I thought I had seen a test in `magic_folder.test.test_api_cli` that set it, but it appears I was mistaken. (Those tests newly hit this class, since I changed the options class there to inherit from this one.

src/magic_folder/test/test_web.py
```diff
@@ -560,15 +486,13 @@ def test_list_folders(self, folders):
             False,
         )
 
-        # note that treq_for_folders() will end up creating a configuration here
-        config = load_global_configuration(basedir)
```
tomprince, line 564: This `load_global_configuration` call was the only reason we weren't using `create_testing_configuration` in these tests. By providing a version of `treq_for_folders` that also returns the config, we can get rid of this call, and switch to using `create_testing_configuration` in (the refactored) `treq_for_folders`, which is #333.

I made this change here, since I wanted to write another tests that uses the config to make assertions about the effect of calling the API. It also turns out that the new helper was useful in writing the CLI tests as well.

src/magic_folder/test/cli/common.py
```diff
@@ -26,6 +29,66 @@
...
+@attr.s
+class Command(object):
+    """
+    A magic-folder command entry point.
+
+    :ivar Type[usage.Options] options: The options class to use.
+    :ivar Callable[[usage.Options], Deferred[None]] entry_point:
+        The function to pass the options to.
+    """
+    options = attr.ib(validator=_subclass_of(usage.Options))
```
meejah, line 89: What's wrong with the built-in "instance_of"?

tomprince: This is not taking an object, but a class. `issubclass` is not `isinstance`, which is what these validators map to.

src/magic_folder/test/test_web.py
```diff
@@ -701,6 +626,204 @@ def test_werkzeug_issue_2157_fix(self):
...
+            )
+        )
+
+    @given(
+        local_authors(),
+        folder_names(),
+        relative_paths(),
+        binary(),
+    )
+    def test_snapshot_invalid_json(self, author, folder_name, path_in_folder, some_content):
```
meejah, line 790: name doesn't really match test

src/magic_folder/web.py
```diff
@@ -357,6 +358,23 @@ def list_all_sanpshots(self, request):
         _application_json(request)
         return json.dumps(dict(_list_all_snapshots(self._global_config)))
 
+    @app.route("/magic-folder/<string:folder_name>/scan", methods=['PUT'])
+    @inline_callbacks
+    def scan_folder(self, request, folder_name):
+        try:
+            folder_service = self._global_service.get_folder_service(folder_name)
+        except KeyError:
+            returnValue(NoResource(b"{}"))
```
meejah, line 367: this ends up doing the wrong thing (i believe you said earlier) (i.e. HTML instead of json)

src/magic_folder/web.py
```diff
@@ -357,6 +358,23 @@ def list_all_sanpshots(self, request):
...
+
+        body = _load_json(request.content.read())
+        if body != {"wait-for-snapshots": True}:
```
meejah, line 370: Seems weird to have an option if we only ever do one thing...

src/magic_folder/web.py
tomprince: Yes, good catch. I guess I must have written this before I discovered that issue.

src/magic_folder/web.py
tomprince: My feeling is that we may not always want this behaviour, and if that is the case, there is a reasonable chance to the default may not be the current behaviour. This way, we can add an option in the future, and have the default be whatever we want, and old clients will still get the current behaviour (that is, we can backwards compatibly add this option with whatever default we want).

meejah: There's no use-case for it. If we only have one behavior, it doesn't need an option. (We can always add any _new_ option we want in the future and give it an appropriate default, whether this or anything else).

tomprince: We *can't* pick whatever default we want, without changing the behavior. The only way to add a new option without causing a change of behavior, is for the default is for to be the existing behavior. We can also *remove* this option in the future, if we decide the current behavior is the only one we want to support, as long as we continue to accept (and ignore) the option.

meejah: So you're saying you want to keep this option now, even though it's essentially hard-coded, because you believe we might in the future want to change the endpoint to by default not wait for snapshots...?

tomprince: Exactly.

tomprince: The use case I have in mind is when the caller wants fire-and-forget behavior, they want to trigger a scan, but don't want to wait for the results. That feels like it is probably the most common use case. I needed/wanted a endpoint that waited for snapshots in the tests, so that is what I implemented; it was also easier to reason about not having the possibility of queuing the same file to be snapshoted twice, before the first was complete

meejah: Sure, waiting is fine. Why can't we just add an option _if_ we want that in future ("fire and forget") instead of cluttering it now?

tomprince: We definitely could. I'd slightly prefer that to be the default behavior though. If the clutter wasn't entirely localized to the function (and its callers), or we couldn't just drop the option, if we don't want to keep it, I wouldn't suggest the option. As it is, it feels like keeping the option of having sensible defaults in the future has minimal maintenance burden (both immediate and ongoing).

src/magic_folder/client.py
```diff
@@ -160,6 +160,12 @@ def add_folder(self, magic_folder, author_name, local_path, poll_interval, scan_
             'scan_interval': scan_interval,
         }, ensure_ascii=False).encode('utf-8'))
 
+    def scan_folder(self, magic_folder):
```
exarkun, line 163: Please add a docstring here.

integration/test_synchronize.py
```diff
@@ -40,6 +40,16 @@ def add_snapshot(node, folder_name, path):
     return node.add_snapshot(folder_name, path)
 
 
+def scan_folder(node, folder_name, path):
+    """
+    Scan the given magic folder. This should cause the given path to be
+    snapshotted.
+
+    :param MagigFolderEnabledNode node: The node on which to do the scan.
```
exarkun, line 48: typo - `Magig`

src/magic_folder/web.py
exarkun: I probably would have left this option out and added something later when we found we needed it.  Since it's already here though, I don't see a lot of value in removing it.

src/magic_folder/test/cli/common.py
```diff
@@ -26,6 +29,66 @@
...
+            http_client._agent.flush()
+            yield
+
+    coop = Cooperator()
+    task = coop.cooperate(pump())
+    try:
+        yield
+    finally:
+        task.stop()
+        coop.stop()
```
exarkun, line 63: Ugh.  It is quite dismaying to see this kind of general infrastructure _still_ being written and rewritten in application-specific test suite helper libraries...

I guess this comment is not actionable.

src/magic_folder/test/cli/common.py
```diff
@@ -26,6 +29,66 @@
...
+
+def _subclass_of(base):
```
exarkun, line 65: Please add a docstring here.

src/magic_folder/test/cli/common.py
```diff
@@ -38,10 +101,11 @@ def succeeded(self):
         return self.code == 0
 
 @inline_callbacks
-def cli(argv, global_config=None, http_client=None):
+def _run_cli(command, argv, global_config=None, http_client=None):
     """
-    Perform an in-process equivalent to the given magic-folder command.
+    Perform an in-process equivalent to the given command.
```
exarkun, line 106: is it worth saying something about the special `RequestTraversalAgent`-handling behavior you also get from this function?

src/magic_folder/test/cli/test_api_cli.py
```diff
@@ -0,0 +1,122 @@
...
+            self.folder_name
+        )
+
+        self.folder_service.local_snapshot_service.startService()
+        self.addCleanup(self.folder_service.local_snapshot_service.stopService)
+
+    @inline_callbacks
+    def test_scan_magic_folder(self):
+        """
+        Scaning a magic folder creates a snaphot of new files.
```
exarkun, line 76: Typo - `Scaning`

src/magic_folder/test/test_web.py
```diff
@@ -629,6 +630,197 @@ def test_werkzeug_issue_2157_fix(self):
...
+    @given(
+        author_names(),
+        folder_names(),
+        relative_paths(),
+        binary(),
+    )
+    def test_scan_folder(self, author_name, folder_name, path_in_folder, some_content):
+        """
+        A **PUT** to **/v1/magic-folder/:folder-name/scan** creates a new local
+        snapshot for a file at in the named folder.
```
exarkun, line 707: Some jumbled words here - `for a file at in the`

src/magic_folder/test/test_web.py
```diff
@@ -629,6 +630,197 @@ def test_werkzeug_issue_2157_fix(self):
...
+        folder_config = node.global_config.get_magic_folder(folder_name)
+        snapshot_paths = folder_config.get_all_localsnapshot_paths()
+        self.assertThat(
+            snapshot_paths,
+            Equals({path2magic(path_in_folder)}),
+        )
+
+    def test_snapshot_no_folder(self):
+        """
+        An error results using /v1/magic-folder/<folder-name>/scan API on non-existent
```
exarkun, line 754: maybe "results from"?

exarkun (approved):
Thanks.  Some minor comments inline.  Otherwise this looks very good.  There are some conflicts with main but they look pretty simple.  Please resolve those, the inline comments, get a completely green CI run, and then merge.  Thanks.

src/magic_folder/web.py
meejah: The "clutter" here is in the public HTTP API not just internal .. so now GridSync has to hard-code use of this option too (for example).

PR #465, opened 2021-07-01
Initial scanner service

tomprince:
See #138.

This is the initial implementation of the scanner itself. It isn't currently hooked up to anything. There will be followups to:
- add an endpoint that will request an immediate scan ( #463)
- add configuration to scan periodically

The code for both of those is mostly written, but needs polish and tests.

I filed  #464 for adding coverage to the `@exclusively` helper.

### DIFF
--- a/src/magic_folder/scanner.py
+++ b/src/magic_folder/scanner.py
@@ -1,0 +1,131 @@
+# Copyright (c) Least Authority TFA GmbH.
+
+"""
+Scan a Magic Folder for changes.
+"""
+
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import attr
+from eliot import start_action, write_failure
+from eliot.twisted import inline_callbacks
+from twisted.application.service import Service
+from twisted.internet.defer import DeferredLock, gatherResults
+from twisted.internet.task import Cooperator
+
+from .magicpath import path2magic
+from .util.file import get_pathinfo
+from .util.twisted import exclusively
+
+
+def _create_cooperator(clock):
+    """
+    Create a cooperator
+    """
+
+    def schedule(f):
+        return clock.callLater(0, f)
+
+    # NOTE: We don't use CooperatorSevice here, since:
+    # - There is not a way to set the reactor it uses
+    # - Once we enable periodic scans, we want to wait to stop the cooperator
+    #   stop it until after the TimerService for periodic scans has stoppped,
+    #   so that the Cooperator will have no pending work.
+    return Cooperator(
+        scheduler=schedule,
+    )
+
+
+@attr.s
+class ScannerService(Service):
+    """
+    Periodically scan a local Magic Folder for new or updated files, and request
+    the local snapshot service snapshot them.
+    """
+
+    _config = attr.ib()
+    _local_snapshot_service = attr.ib()
+    _status = attr.ib()
+    _cooperator = attr.ib()
+    _lock = attr.ib(init=False, factory=DeferredLock)
+
+    @classmethod
+    def from_config(cls, clock, folder_config, local_snapshot_service, status):
+        return cls(
+            config=folder_config,
+            local_snapshot_service=local_snapshot_service,
+            status=status,
+            cooperator=_create_cooperator(clock),
+        )
+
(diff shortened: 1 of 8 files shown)


src/magic_folder/scanner.py
```diff
@@ -0,0 +1,129 @@
...
+    Create a cooperator
+    """
+
+    def schedule(f):
+        return clock.callLater(0, f)
+
+    # NOTE: We don't use CooperatorSevice here, since:
+    # - There is not a way to set the reactor it uses
+    # - Once we enable periodic scans, we want to wait to
+    #   stop it until after it has We want to stop it after the TimerService has
```
exarkun, line 32: Something weird going on in this comment

src/magic_folder/scanner.py
```diff
@@ -0,0 +1,129 @@
...
+    #   stoppped, so that it will have no pending work.
+    return Cooperator(
+        scheduler=schedule,
+    )
+
+
+@attr.s
+class ScannerService(Service):
+    """
+    Periodically scan a local Magic Folder for new or updated files
```
exarkun, line 42: Maybe say something about what happens when the scan finds such a thing.

src/magic_folder/scanner.py
```diff
@@ -0,0 +1,129 @@
...
+    :param Cooperator cooperator: The cooperator to use to control yielding to
+        the reactor.
+    :param MagicFolderConfig folder_config: the folder for which we
+        are scanning
+
+    :param callable on_new_file: a 1-argument callable. This function
+        will be invoked for each updated / new file we find. The
+        argument will be a FilePath of the updated/new file.
+
+    :returns: the scan duration, in seconds
```
exarkun, line 109: I think it returns a `Deferred` that fires with ... `TaskDone()` maybe?  Not sure exactly what happens with the version of Twisted we're using, actually.  Probably not a scan duration in seconds though.

src/magic_folder/scanner.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+    # XXX we don't handle deletes
+    def _process():
+        for path in folder_config.magic_path.walk():
+            if path.isdir():
+                continue
+            relpath = "/".join(path.segmentsFrom(folder_config.magic_path))
+            name = path2magic(relpath)
+            try:
+                snapshot_state = folder_config.get_currentsnapshot_pathstate(name)
```
exarkun, line 119: `get_currentsnapshot_pathstate` (and `get_pathinfo`) are both what r0ml might call "potato" interfaces.  I don't think there's much to do about it right now but when we start worrying about poor performance I suspect a fruitful area of investigation will be to vectorize these so we can ask the database for snapshot pathstat for _all_ (or at least N) of the interesting paths at once (dunno if there is something similar we can do for `get_pathinfo` or not ... maybe depends on the platform).

exarkun (approved):
Thanks.  Generally looks good.  It didn't even feel like a 400 line review. :)  Some minor comments inline: address those to your satisfaction.

You mentioned something earlier today about unicode and I see that some CI jobs are failing with what look like some unicode/path issues.  So address those before merging, please.  I dunno if those are deep problems or just annoying ones.  If you want more input on them just let me know.

I was hoping to see a coverage report here to help understand how much test coverage there is for the implementation changes in this PR but I see codecov is apparently not making those reports anymore?  I did manage to find one on the codecov website - https://app.codecov.io/gh/leastauthority/magic-folder/compare/465 - but (somewhat as usual) it's really hard to interpret.  For example, it tells me downloader.py is gaining an uncovered line - but that line is `from functools import wraps` which is itself being deleted.  I don't have the energy right now to dig through the whole report to see how much of it is valid and how much is nonsense. :/  Perhaps you can do a local coverage run and convince yourself things look good (or perhaps you already have done so).

So, all in all, nice, thank you.  Please address the mentioned points and then merge.

src/magic_folder/scanner.py
tomprince: If and when we do, I think we'll need to be careful that we get both bits of info we are going to compare without yielding. That would probably mean unrolling this loop in the cooperator, or something.

PR #468, opened 2021-07-02
Enable periodic scans.

### LINKED ISSUE #138: Observe File addition

tomprince:
Fixes #138.


src/magic_folder/config.py
```diff
@@ -1443,9 +1464,31 @@ def create_magic_folder(self, name, magic_path, author,
...
         valid_magic_folder_name(name)
+        if not isinstance(poll_interval, int) or poll_interval <= 0:
+            raise APIError(
+                code=http.BAD_REQUEST,
+                reason="Poll interval must be a positive integer (not '{!r}').".format(
+                    poll_interval
+                ),
+            )
+        if scan_interval is not None and (
+            not isinstance(scan_interval, int) or scan_interval <= 0
```
hacklschorsch, line 1484: [In the DB above, 0 is an allowed value](https://github.com/LeastAuthority/magic-folder/pull/468/files#diff-db422d0a72330e38baed0b613e6f69f9f154b6e67e0db758516627d0492fb634R147), but here 0 is forbidden.

src/magic_folder/config.py
```diff
@@ -1443,9 +1464,31 @@ def create_magic_folder(self, name, magic_path, author,
...
+        ):
+            raise APIError(
+                code=http.BAD_REQUEST,
+                reason="Scan interval must be a positive integer or null (not '{!r}').".format(
```
hacklschorsch, line 1488: `null` or `0` ? Now my confusion is complete 😸

src/magic_folder/migrate.py
```diff
@@ -81,7 +81,8 @@ def magic_folder_migrate(config_dir, listen_endpoint_str, tahoe_node_directory,
             author,
             mf_config[u'collective_dircap'],
             mf_config[u'upload_dircap'],
-            mf_config[u'poll_interval'],  # is this always available?
+            int(mf_config[u'poll_interval']),  # is this always available?
+            int(mf_config[u'poll_interval']),
```
hacklschorsch, line 85: should this maybe be `scan_interval` ?

src/magic_folder/service.py
```diff
@@ -221,6 +224,13 @@ def create_folder(self, name, author_name, local_dir, poll_interval):
                 reason="Already have a magic-folder named '{}'".format(name),
             )
 
+        if scan_interval is not None and scan_interval < 0:
+            raise APIError(
+                code=http.BAD_REQUEST,
+                reason="scan_interval must be positive integer or null",
```
hacklschorsch, line 230: ...or `0`?

src/magic_folder/test/cli/test_magic_folder.py
```diff
@@ -592,6 +652,61 @@ def test_poll_interval(self):
...
+            )
+        self.assertEqual(str(ctx.exception), "--poll-interval must be a positive integer")
+
+    def test_poll_interval_zero(self):
+        with self.assertRaises(usage.UsageError) as ctx:
+            parse_cli(
+                "add",
+                "--name", "test",
+                "--author", "test",
+                "--poll-interval=-0",
```
hacklschorsch, line 672: And what happens with plain `0` here?  I expect a `UsageError` to be thrown?

src/magic_folder/test/cli/test_magic_folder.py
```diff
@@ -592,6 +652,61 @@ def test_poll_interval(self):
...
+            )
+        self.assertEqual(str(ctx.exception), "--scan-interval must be a positive integer")
+
+    def test_scan_interval_zero(self):
+        with self.assertRaises(usage.UsageError) as ctx:
+            parse_cli(
+                "add",
+                "--name", "test",
+                "--author", "test",
+                "--scan-interval=-0",
```
hacklschorsch, line 705: And what happens with plain `0` here?  I expect it to work and to turn off local scanning.

src/magic_folder/test/test_web.py
```diff
@@ -365,6 +366,7 @@ def magic_folder_config(author, local_directory):
         u"collective-dircap": u"URI:DIR2-RO:{}:{}".format(b2a("\0" * 16), b2a("\1" * 32)),
         u"upload-dircap": u"URI:DIR2:{}:{}".format(b2a("\2" * 16), b2a("\3" * 32)),
         u"poll-interval": 60,
+        u"scan-interval": None,
```
hacklschorsch, line 369: Is `None` different from `null` and `0`?

src/magic_folder/test/test_web.py
```diff
@@ -522,6 +526,36 @@ def test_add_folder_invalid_json(self):
...
+                })
+            ),
+            succeeded(
+                matches_response(
+                    code_matcher=Equals(BAD_REQUEST),
+                    body_matcher=AfterPreprocessing(
+                        loads,
+                        MatchesDict(
+                            {
+                                "reason": StartsWith("scan_interval must be positive integer or null"),
```
hacklschorsch, line 552: So I begin to believe `null` equals `0`, while `None` equals `not set`. Correct?

If this is so in the rest of the application, `null` as the word (as opposed to `positive integer`) is probably better English than `0` here... the digit `0` might be less confusing still

hacklschorsch (commented):
Sweet!

I found one typo between `scan` and `poll` that IMHO should be fixed; the rest is musings about the (in)equality between `0`, `null` and `None` and how `scan_interval` seems (to me) to not be fully consistent regarding this.

src/magic_folder/test/test_web.py
meejah: `null` is JS terminology and is basically the same as `None` (i.e. a `null` in some JSON will become a `None` on the python side and vice-versa). Python doesn't have "`null`".

hacklschorsch: Thanks meejah!

My comment was about our using the three terms seemingly interchangeably.  I was a bit uncompromising maybe with having the word `null` used in docs meaning the number `0`. Users will probably figure that one out.

From my reading the patch I believe currently sometimes `None` and sometimes `0` means _do not scan at all_.
It should be one of those two; I personally like `0` here better since it's of the same type as the other valid values.

meejah: Yeah, definitely "choose one" I'd say.
I also prefer `0` to mean "disable scanning" (which is also somewhat common for these sorts of options IME).

meejah:
@tomprince I think my notifications say "requested review" .. do you mean on top of the review @hacklschorsch already provided? Also CI is red currently too (so either Flo or I can take another pass after the changes Flo suggests and CI is green).

src/magic_folder/test/cli/test_magic_folder.py
tomprince: I think this was a bad copy-paste from the `-1` (where I changed the `1` and didn't remove the `0`. Fixed.

src/magic_folder/migrate.py
tomprince: tahoe-lafs didn't have that (which is where this is migrating from. I've added a comment to clarify.

src/magic_folder/config.py
```diff
@@ -1443,9 +1464,31 @@ def create_magic_folder(self, name, magic_path, author,
         :param unicode upload_dircap: the write-capability of the
             directory we upload data into.
 
+        :param int poll_interval: how often to scan for remote changes
+            (in seconds).
+
+        :param Optional[int] scan_interval: how often to scan for local changes
+            (in seconds).
```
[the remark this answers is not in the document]
tomprince: No, I intended for `0` to be disallowed, and `None` to indicate turning it off. I've fixed the help in the cli usage.

src/magic_folder/config.py
tomprince: `null`

src/magic_folder/config.py
tomprince: I've added a constraint to the DB.

src/magic_folder/test/test_web.py
tomprince: I think the only place where `0` was mentioned was in the `--help` output (and it isn't supported. I've removed that mention of it. `0` shouldn't be supported, and `null` (in the API/db)/ `None` in python means no scanning.

tomprince:
> @tomprince I think my notifications say "requested review" .. do you mean on top of the review @hacklschorsch already provided? Also CI is red currently too (so either Flo or I can take another pass after the changes Flo suggests and CI is green + conflicts resolved).

I requested that before the review @hacklschorsch did. In any case, I think I've addressed the issues.

hacklschorsch:
I believe @tomprince requested a review by @meejah, but I did one anyway - thought I might be able to help out :) Seems I added confusion. Sorry. Anyway, thanks for addressing my points @tomprince ; Great work as always!

meejah:
okay, if things are fixed per Flo's review once conflicts are resolved and CI are green please merge.

Issue #461, opened 2021-07-01
Don't upload multiple local snapshots at once.


tomprince:
Currently, when we upload local snapshots, we load all the existing snapshots for a given file from our database, requests that *all* of them be uploaded to tahoe, and then publish the leaf snapshot in our personal DMD.

This causes a number of issues:
- we can lose snapshots taken while uploading (#459, #385)
- there can be a unbounded delay in publishing any snaphost of the file (#460)

Instead, we should publish snapshots as we upload them (and update our database of local snapshots to replace the uploaded snapshot with the remote version).

Issue #173, opened 2020-07-02
Missing tests for with_cursor decorator


vu3rdd:
PR #168 adds a new decorator, `with_cursor` adopted from ZKAP repository. However, there are no explicit tests for the `with_cursor` itself.

PR #470, opened 2021-07-05
Fix SQL re-entrancy bugs in `MagicFolderConfig`.

### LINKED ISSUE #382: Fix `MagicFolderConfig.store_local_snapshot` to use transactions correctly.
It uses the `with_cursor` decorator, but calls it self recursively. If I am reading the sqlite docs and sqlite module docs correctly, this won't do the right thing, as the transaction type used is not recursive.

It looks like this is the only function that runs into this issue, but it would be good to make `with_cursor` resilient to this kind of error as well.

### LINKED ISSUE #457: Magic folder stop uploading snapshot for a file, if it crashes at the wrong point while taking a local snaphot of it.
If magic folder crashes at the wrong point while recording a local snapshot, it will stop uploading snapshots or taking new local snapshots of that file. It may also cause other parts of the service to stop working.

When we take a snapshot, we write the snapshot to our database, including writing out pointer to its parents. There code that does that uses transactions incorrectly (#382), when a snapshot has parents, that can cause it to leave the database in an inconsistent state if the service crashes. When the service starts up again, it will not be able to retrieve any local snapshots for that file from the database, preventing the service from uploading them, or creating new child snapshots.

tomprince:
Fixes #382. Fixes #457.

This updates `with_cursor` to ensure that it can't be used recursively, and then fixes the two issues that discovered:
- `.get_local_snapshot` called `.author`; added a entrypoint which can be called with a cursor
- `.store_local_snapshot` called itself recursively with its parents; changed to check that its parents exist instead of trying to add them unconditionally. Since we only ever add one snapshot at a time, the existing callers don't need adjusting.

### DIFF
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -1498,7 +1498,7 @@ class GlobalConfigDatabase(object):
         # Close the per-folder state database. Since `get_magic_folder` caches
         # its return value, this should be the only instance. We need to close
         # it explicitly since otherwise we can't delete it below on windows.
-        folder_config.database.close()
+        folder_config._database.close()
         # we remove things from the database first and then give
         # best-effort attempt to remove stuff from the
         # filesystem. First confirm we have this folder and its
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -36,7 +36,6 @@ from hyperlink import (
 
 from functools import (
     partial,
-    wraps,
 )
 from itertools import (
     chain,
@@ -106,6 +105,7 @@ from eliot import (
 )
 
 from .util.capabilities import is_readonly_directory_cap, is_directory_cap
+from .util.database import with_cursor, LockableDatabase
 from .util.eliotutil import (
     RELPATH,
     validateSetMembership,
@@ -443,24 +443,6 @@ def load_global_configuration(basedir):
     )
 
 
-# XXX: with_cursor lacks unit tests, see:
-#      https://github.com/LeastAuthority/magic-folder/issues/173
-def with_cursor(f):
-    """
-    Decorate a function so it is automatically passed a cursor with an active
-    transaction as the first positional argument.  If the function returns
-    normally then the transaction will be committed.  Otherwise, the
-    transaction will be rolled back.
-    """
-    @wraps(f)
-    def with_cursor(self, *a, **kw):
-        with self.database:
-            cursor = self.database.cursor()
-            cursor.execute("BEGIN IMMEDIATE TRANSACTION")
-            return f(self, cursor, *a, **kw)
-    return with_cursor
-
-
 def _upgraded(schema, connection):
     """
     Return ``connection`` fully upgraded according to ``schema``.
@@ -736,7 +718,7 @@ class MagicFolderConfig(object):
     Low-level access to a single magic-folder's configuration
     """
     name = attr.ib()
-    database = attr.ib()  # sqlite3 Connection
+    _database = attr.ib(converter=LockableDatabase)  # sqlite3 Connection
     _get_current_timestamp = attr.ib(default=time.time)
 
     @classmethod
@@ -1369,13 +1351,14 @@ class GlobalConfigDatabase(object):
         return self._token_provider.rotate()
 
     @property
-    @with_cursor
-    def api_endpoint(self, cursor):
+    def api_endpoint(self):
         """
         The twisted server-string describing our API listener
         """
-        cursor.execute("SELECT api_endpoint FROM config")
-        return cursor.fetchone()[0].encode("utf8")
+        with self.database:
+            cursor = self.database.cursor()
+            cursor.execute("SELECT api_endpoint FROM config")
+            return cursor.fetchone()[0].encode("utf8")
 
     @api_endpoint.setter
     def api_endpoint(self, ep_string):
@@ -1385,13 +1368,14 @@ class GlobalConfigDatabase(object):
             cursor.execute("UPDATE config SET api_endpoint=?", (ep_string, ))
 
     @property
-    @with_cursor
-    def api_client_endpoint(self, cursor):
+    def api_client_endpoint(self):
         """
         The twisted client-string describing our API listener
         """
-        cursor.execute("SELECT api_client_endpoint FROM config")
-        return cursor.fetchone()[0].encode("utf8")
+        with self.database:
+            cursor = self.database.cursor()
+            cursor.execute("SELECT api_client_endpoint FROM config")
+            return cursor.fetchone()[0].encode("utf8")
 
     @api_client_endpoint.setter
     def api_client_endpoint(self, ep_string):
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -273,6 +273,14 @@ class LocalSnapshotCollision(Exception):
     database.
     """
 
+@attr.s(auto_exc=True, frozen=True)
+class LocalSnapshotMissingParent(Exception):
+    """
+    An attempt was made to store a local snapshot whose parents aren't in
+    the local database.
+    """
+    parent_identifier = attr.ib(validator=attr.validators.instance_of(UUID))
+
 
 @attr.s(auto_exc=True, frozen=True)
 class RemoteSnapshotWithoutPathState(Exception):
@@ -858,6 +866,21 @@ class MagicFolderConfig(object):
 
         :param LocalSnapshot snapshot: The snapshot to store.
         """
+        # Ensure that the local parent snapshots are already in the database.
+        for parent in snapshot.parents_local:
+            cursor.execute(
+                """
+                SELECT
+                    count(*) from [local_snapshots]
+                WHERE
+                    identifier= ?
+                """,
+                (unicode(parent.identifier),),
+            )
+            count = cursor.fetchone()
+            if count[0] != 1:
+                raise LocalSnapshotMissingParent(parent.identifier)
+
         try:
             # Create the primary row.
             cursor.execute(
@@ -906,41 +929,23 @@ class MagicFolderConfig(object):
                     [parent_identifier]
                 )
             VALUES
-                (?, ?, 0, ?)
+                (?, ?, ?, ?)
             """,
-            list(
-                (unicode(snapshot.identifier), index, parent_identifier)
-                for (index, parent_identifier)
-                in enumerate(snapshot.parents_remote)
-            ),
-        )
-
-        # Create any implied local parents that don't exist already.
-        for local_parent in snapshot.parents_local:
-            try:
-                self.store_local_snapshot(local_parent)
-            except LocalSnapshotCollision:
-                # If it exists already, fine.
-                pass
-
-        # Now insert references to them.
-        cursor.executemany(
-            """
-            INSERT INTO
-                [local_snapshot_parent] (
-                    [snapshot_identifier],
-                    [index],
-                    [local_only],
-                    [parent_identifier]
+            [
+                (unicode(snapshot.identifier), index, local_only, parent_identifier)
+                for (index, (local_only, parent_identifier)) in enumerate(
+                    chain(
+                        (
+                            (False, parent_identifier)
+                            for parent_identifier in snapshot.parents_remote
+                        ),
+                        (
+                            (True, unicode(parent.identifier))
+                            for parent in snapshot.parents_local
+                        ),
+                    )
                 )
-            VALUES
-                (?, ?, 1, ?)
-            """,
-            list(
-                (unicode(snapshot.identifier), index, unicode(parent.identifier))
-                for (index, parent)
-                in enumerate(snapshot.parents_local, len(snapshot.parents_remote))
-            ),
+            ],
         )
 
     @with_cursor
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -813,8 +813,11 @@ class MagicFolderConfig(object):
         return cls(name, connection)
 
     @property
+    def author(self):
+        return self._get_author()
+
     @with_cursor
-    def author(self, cursor):
+    def _get_author(self, cursor):
         cursor.execute("SELECT author_name, author_private_key FROM config")
         name, keydata = cursor.fetchone()
         return LocalAuthor(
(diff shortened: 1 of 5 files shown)


src/magic_folder/config.py
```diff
@@ -862,6 +855,20 @@ def store_local_snapshot(self, cursor, snapshot):
...
+                SELECT
+                    count(*) from [local_snapshots]
+                WHERE
+                    identifier= ?
+                """,
+                (unicode(parent.identifier),),
+            )
+            count = cursor.fetchone()
+            if count[0] != 1:
+                raise LocalSnapshotMissingParent(parent.identifier)
```
tomprince, line 870: I think this check may end up being temporary, as I think it may make sense to move the logic from [here](https://github.com/LeastAuthority/magic-folder/blob/a53f0bab97dfa897ad535ae9f23c56b78cb52429/src/magic_folder/uploader.py#L121-L138) into the call to `store_local_snapshots` below (as part of #461).

meejah: ^ would probably be better as a comment in the code describing why the check is there (and under what conditions you think it can go away).

tomprince: I don't think it makes sense to have a comment in the code about removing the check. I think it is information that is useful to a reviewer, as I am adding a check in one PR, but already have a WIP PR (#471) where I then turn around and review it.

I don't think I think it makes sense to add a comment to code, about how a potential interface change to this code, would remove the need for the check. The check makes sense as the code is, but obviously doesn't with the anticipated change. In particular, as it is *potential* change[1], it seems like a comment is something that we easily get out of date.

[1] In writing this comment, I realized that plan I had doesn't handle one particular case, and so the plan I had for changing this may not actually make sense, which strengthens my feeling that it doesn't make sense to mention it in the code.

---

Here is a bit of an explanation of the changes I was intending:

The WIP code I have changes this function to take a local snapshot with unspecified parents, so I replace this check with one that checks that that both `.parents_local` and `.parents_raw` are `None` (for unspecified) as opposed to a (possibly empty) list. I do this so we can find the current snapshot (which will be the parent) in the same transaction we write the data for it. I wanted to do this so that if a local parent snapshot is uploaded and becomes remote as we are writing the stash file, we aren't left with a dangling snapshot identifier).

However, I realized that (in some cases), we could also race with the downloader in a way that would cause the calculated parent to become incoherent, so I made need to take a different approach. I added some further discussion of my current thoughts on this race [here](https://github.com/tp-la/magic-folder/commit/5d69e7c4def99d3024b67eef6c5a81090bacdf48), which is currently part of #471.

src/magic_folder/util/database.py
```diff
@@ -0,0 +1,132 @@
...
+        default=False, init=False, validator=attr.validators.instance_of(bool)
+    )
+
+    @contextlib.contextmanager
+    def transaction(self, function_name=None):
+        """
+        Context manager that starts a transaction and rollsback
+        if an exception is raised.
+
+        :param Optional[function_name]: The name of the calling function, used
```
exarkun, line 78: `:param Optional[str] function_name:` perhaps?  However, there seems to be exactly one caller and it supplies a value.  So maybe it needn't be optional?  Unless `f.__name__` is allowed to be `None`...

src/magic_folder/util/database.py
```diff
@@ -0,0 +1,132 @@
...
+
+    _database = attr.ib(validator=attr.validators.instance_of(sqlite3.Connection))
+    _in_transaction = attr.ib(
+        default=False, init=False, validator=attr.validators.instance_of(bool)
+    )
+
+    @contextlib.contextmanager
+    def transaction(self, function_name=None):
+        """
+        Context manager that starts a transaction and rollsback
```
exarkun, line 75: I think "rolls back" must be two words. :slightly_smiling_face:

src/magic_folder/util/database.py
```diff
@@ -0,0 +1,132 @@
...
+            with self._database:
+                cursor = self._database.cursor()
+                cursor.execute("BEGIN IMMEDIATE TRANSACTION")
+                yield cursor
+        finally:
+            self._in_transaction = False
+
+
+# XXX: with_cursor lacks unit tests, see:
+#      https://github.com/LeastAuthority/magic-folder/issues/173
```
exarkun, line 102: So close to being fully tested now!  But I guess that just means #173 will be easily resolved with a follow-up PR.

src/magic_folder/util/database.py
```diff
@@ -0,0 +1,132 @@
...
+        :raises RecusiveTransaction: if used recusively.
+
+        :return sqlite3.Cursor:
+        """
+        if self._in_transaction:
+            frame = currentframe(2)
+            if frame.f_code == with_cursor._wrapped_code:
+                frame = frame.f_back
+            caller = frame.f_code.co_name
+            raise RecusiveTransaction(function=function_name, caller=caller)
```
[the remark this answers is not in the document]
exarkun: And I guess I shouldn't say anything about `SAVEPOINT` here... maybe some other time.

src/magic_folder/config.py
```diff
@@ -849,7 +842,7 @@ def get_local_snapshot(self, cursor, name):
         return _construct_local_snapshot(
             leaf_identifier,
             name,
-            self.author,
+            self._get_author.__wrapped__(self, cursor),
```
exarkun, line 845: `__wrapped__` doesn't appear to be a documented feature of `functools.wraps`.  I'm only guessing that's what `six.wraps` actually resolves to...

I suspect you could move the `with_cursor` decorator back on to `author` and then have `_get_author` available as the function you want to have here, but I'm not sure.  If not, I guess it's fine ... if someone messes with `__wrapped__` then I guess this code will break and the tests will tell us, not a big deal...

src/magic_folder/config.py
```diff
@@ -862,6 +855,21 @@ def store_local_snapshot(self, cursor, snapshot):
...
+        for parent in snapshot.parents_local:
+            cursor.execute(
+                """
+                SELECT
+                    count(*) from [local_snapshots]
+                WHERE
+                    identifier= ?
+                """,
+                (unicode(parent.identifier),),
+            )
```
exarkun, line 868: I suppose that if we wanted this to be one query instead of N we could do dynamic SQL query string generation to make the `IN ...` expression to check all the parents at once.  But since N = 1 right now I guess there's no point.

src/magic_folder/config.py
exarkun: > I don't think it makes sense to have a comment in the code about ...

I guess different things might make different amounts of sense to different people.  In the case of this kind of documentation request, maybe we should err on the side of explaining more things in more places, hoping that anyone who goes looking for an explanation will find at least _something_ that helps them out?

> In particular, as it is potential change[1], it seems like a comment is something that we easily get out of date.

I'm not sure what it would get out of date with.  Is the plan written down somewhere else?  If so, we don't have to be redundant - adding a link in the code to wherever else that is still does a better job of maximizing the chances of everyone understanding (as compared to not adding any explanation).

src/magic_folder/config.py
```diff
@@ -1269,13 +1259,14 @@ def rotate_api_token(self):
         return self._token_provider.rotate()
 
     @property
-    @with_cursor
```
exarkun, line 1272: It seems a little bit unfortunate that `with_cursor` is now incompatible with `GlobalConfigDatabase`?  Maybe mainly because this is a surprising outcome - only *some* of our database code can use the helper.  But maybe also eventually practically, as we have to duplicate more transaction management code?  Or forget to duplicate it and just get it wrong.

Is it already wrong?  `remove_magic_folder` looks up the location of folders separately from deleting them.  And little of the code in `GlobalConfigDatabase` explicitly starts a transaction at all.  There's a lot of `with self.database` but I don't remember what that means for a pysqlite Connection object.

exarkun (commented):
Thanks.  Some comments inline.

src/magic_folder/config.py
tomprince: `__wrapped__` was added in python 3.2, as something set by `update_wrapper` (which `wraps` is documented as calling).

src/magic_folder/config.py
tomprince: > In the case of this kind of documentation request, maybe we should err on the side of explaining more things in more places

While I generally agree with this, I don't think it applies in this case. It was added to forestall a question from a reviewer about why one PR added a check, and then the next removed it. I think it is only relevant in the context of reviewing this and #471 in reasonably close succesion, which, in hindsight hasn't happened. 

In particular, I think adding a comment about this to the code makes it *less* clear.

src/magic_folder/config.py
tomprince: I agree. The reason I made it incompatible with the current `GlobalConfigDatabase` code is because I wanted to ensure that any code that touches the same database as `with_cursor` goes through the same code, to ensure that we have consistent transaction handling.

There was enough code in `GlobalConfigDatabase` that didn't use `with_cursor` that I didn't want to rewrite entirely in this PR, that I thought it made more sense to remove the two uses of it here, for the moment.

>  There's a lot of with self.database but I don't remember what that means for a pysqlite Connection object.

All it does is call either `.commit` or `.rollback` when the block exits, depending on whether the block exits with an exception or not. However, [by default](https://docs.python.org/2.7/library/sqlite3.html#controlling-transactions), pysqlite automatically starts a transaction for any statement that it thinks modifies data.

> Is it already wrong? [...]

Skimming the code, I don't *think* so, assuming that we are the only process writing to the database (which we should be, assuming there isn't multiple copies of the service running, which is an assumption we make elsewhere too).

src/magic_folder/util/database.py
tomprince: I do want to support direct usage of this function as well. I think, for example, that `GlobalConfigDatabase` may want to use this in some places.

src/magic_folder/util/database.py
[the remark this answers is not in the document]
tomprince: > And I guess I shouldn't say anything about SAVEPOINT here... maybe some other time.

I did spend a bit of time looking at that. I think we'd need to switch to managing transactions manually, rather than using pysqlite's partially automatic handling (see [here](https://docs.python.org/2.7/library/sqlite3.html#controlling-transactions)). In particular, we'd need to manually handle commit vs rollback as `with database` doesn't understand `SAVEPOINT` (and I had thought that it didn't work with regular transaction in manual mode, but reading the py2.7 code, it does still seem to work for that).

It may be worth doing that in the future (at the least stopping using the automatic transaction management).

src/magic_folder/test/test_upload.py
```diff
@@ -218,10 +218,10 @@ def test_write_snapshot_to_tahoe_fails(self, name, contents):
                 d,
                 succeeded(Always()),
             )
+            config.store_local_snapshot(snapshots[-1])
             parents = [snapshots[-1]]
 
         local_snapshot = snapshots[-1]
-        config.store_local_snapshot(snapshots[-1])
```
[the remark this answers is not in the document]
tomprince: With `MAGIC_FOLDER_HYPOTHESIS_PROFILE=magic-folder-ci`, the test fails by raising `LocalSnapshotMissingParent` from this line, as soon as `contents` is longer than 1.

> The visible impact of this change is that the test goes from passing and quiet to passing and dumping a hundred lines of errors to stdout.

Are you sure that the test had no output before? At least for me, I get ouput in either case. #444 fixes the extraneous output.

src/magic_folder/util/database.py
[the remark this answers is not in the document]
tomprince: > Also, the full stack is necessary (or at least desirable) to debug the problem in the general case. It is just as much an error to attempt to begin a nested transaction after going through a function not decorated with with_cursor - so currentframe(2) doesn't necessarily get you to anything particularly interesting, database-wise.

The `2` is about the `contextlib.contextmanager` wrapper (specifically `GeneratorContextManager.__enter__` I think). So 
`currentframe(2)` should always point at the caller of `.transaction()`. I've got a commit with a comment explaining that I had written awhile ago adding a `.close` for https://github.com/LeastAuthority/magic-folder/pull/475/commits/f4bbe95402dbddb2b052e77ab29042d2ad4b85ca (which I only discovered I had needed after opening this PR). I'll push that here, since otherwise this will fail once its merged.

> Eh ... I am not excited to see stack and frame inspection code here.

I don't feel strongly about having this, though I do think it is valuable to include the *called* function name for the `with_cursor` case, since that won't be in the stack trace. I don't recall exactly why I added it; it may have been for convenience when figuring out where we were triggering this, or it may been so the message would have symmetry of "from <caller>" vs the "when calling <callee>".

I suspect part of my thinking was that this message is probably enough to diagnose the issue *without* needing the full stack trace, so spending a bit of effort here on a good error message will save time down the road (particularly given our current usage of transactions in `@property` everywhere).

tomprince:
I think I've addressed or responded to all the comments. I've also pushed an extra commit that adds `.close` to the wrapper, which is needed after #475 (which I added after I had pushed this PR).

src/magic_folder/config.py
tomprince: In fact, it looks like I have a local commit switching it over, but it was causing test failures, due to `list_magic_folders` being a generator, which is used in a context where `get_magic_folder` is called while iterating the generator. That is certainly fixable, but I didn't want spend time on it, particularly given that both of those functions are read-only.

src/magic_folder/util/database.py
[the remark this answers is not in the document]
exarkun: > I did spend a bit of time looking at that. I think we'd need to switch to managing transactions manually, rather than using pysqlite's partially automatic handling (see here).

For what it's worth, I don't think there was any particular *intent* to use pysqlite's partially automatic transaction handling - it's just what you get if you don't read a bunch of extra documentation and carefully choose your connect parameters.  So, yes, at some point it probably makes sense to move away from that.

src/magic_folder/test/test_upload.py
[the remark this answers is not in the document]
exarkun: Okay, I see the output on main now too.

exarkun (approved):
Thanks.  Looks good to merge.

PR #474, opened 2021-07-06
Use pytest's temporary directory management.

### LINKED ISSUE #473: Use pytest's native temporary directory handling.
We currently have some custom code for keeping around (some) temporary directories, but also create a bunch that aren't managed at all. We should just delegate to pytest.

tomprince:
Fixes #473.

### DIFF
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -5,13 +5,11 @@ from __future__ import (
 )
 
 import sys
-import shutil
 import subprocess
 from pathlib2 import Path
 from time import sleep
 from os import mkdir, listdir, environ
 from os.path import join, exists
-from tempfile import mkdtemp
 from configparser import ConfigParser
 
 from foolscap.furl import (
@@ -47,10 +45,6 @@ from .util import (
 # pytest customization hooks
 
 def pytest_addoption(parser):
-    parser.addoption(
-        "--keep-tempdir", action="store_true", dest="keep",
-        help="Keep the tmpdir with the client directories (introducer, etc)",
-    )
     parser.addoption(
         "--coverage", action="store_true", dest="coverage",
         help="Collect coverage statistics",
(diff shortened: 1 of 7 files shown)


pytest.ini
```diff
@@ -1,4 +1,5 @@
 [pytest]
+testpaths = integration
```
tomprince, line 2: This is unrelated, but means `pytest` without arguments doesn't try to run trial tests. I ran into this when testing this branch. In particular, our trial tests have incompatible assumptions about temporary paths, and will write to the current directory when run via pytest.

integration/conftest.py
```diff
@@ -96,28 +91,22 @@ def reactor():
...
-        if request.config.getoption('keep'):
-            print("Keeping tempdir '{}'".format(tmp))
-        else:
-            try:
-                shutil.rmtree(tmp, ignore_errors=True)
-            except Exception as e:
-                print("Failed to remove tmpdir: {}".format(e))
-    request.addfinalizer(cleanup)
+    base = tmp_path_factory.mktemp("magic-folder")
+    return str(base)
```
exarkun, line 100: Should this really be `str` and not `bytes` or `unicode`?

integration/conftest.py
```diff
@@ -96,28 +91,22 @@ def reactor():
...
 
-    return tmp
+
+@pytest.fixture(scope='function')
+def temp_filepath(tmp_path):
+    """
+    Fixture that returns a per test function temporary directory represented
+    as a :py:`FilePath`.
+    """
+    return FilePath(str(tmp_path))
```
exarkun, line 109: Ditto comment about `str` here

integration/conftest.py
```diff
@@ -96,28 +91,22 @@ def reactor():
 
 
 @pytest.fixture(scope='session')
-@log_call(action_type=u"integration:temp_dir", include_args=[])
-def temp_dir(request):
+@log_call(action_type=u"integration:base_dir", include_args=[])
+def base_dir(request, tmp_path_factory):
```
exarkun, line 95: It might be worth documenting that `tmp_path_factory` (and `request`) are magic pytest fixtures since it may not be obvious to all readers all the time.  A link to the docs might be nice too, although I didn't have any trouble looking up `tmp_path_factory` just based on the name.

integration/conftest.py
```diff
@@ -96,28 +91,22 @@ def reactor():
...
-            except Exception as e:
-                print("Failed to remove tmpdir: {}".format(e))
-    request.addfinalizer(cleanup)
+    base = tmp_path_factory.mktemp("magic-folder")
+    return str(base)
 
-    return tmp
+
+@pytest.fixture(scope='function')
+def temp_filepath(tmp_path):
```
exarkun, line 104: Some words in the docstring about `tmp_path` would be nice, too, I think.

exarkun (approved):
Thanks.  This looks pretty good to me.  I left a few minor comments inline.  Also I see there are some conflicts with main.  They're not all trivial but neither do they look terrible complicated.

Please address inline comments, resolve conflicts, and then merge.  Thanks.

integration/conftest.py
tomprince: I *think* so yes. I'm specifically wanting to call `pathlib2.Path._str__` here. The docstring for `__bytes__` suggests it should only be used on unix. However, on python2, unicode-mode `FilePath` doesn't allow filesystem operations[1].

[1] This may just be because it forwards the values to the underlying `os` etc function, which tries to encode to ascii.

PR #475, opened 2021-07-06
Add an HTTP API to leave a magic folder.

### LINKED ISSUE #318: HTTP API for "magic-folder leave"
We need an HTTP API that does the work of removing a magic-folder.

### LINKED ISSUE #319: "magic-folder leave" uses HTTP API
Once #318 is done, the `magic-folder leave` command should use the HTTP API.

### LINKED ISSUE #320: Add/remove services on the fly

tomprince:
Fixes #320. Fixes #318. Fixes #319.

### DIFF
--- a/src/magic_folder/api_cli.py
+++ b/src/magic_folder/api_cli.py
@@ -426,7 +426,7 @@ def run_magic_folder_api_options(options):
 
         except MagicFolderApiError as e:
             # these kinds of errors should report via JSON from the endpoints
-            print(u"{}".format(e.body), file=options.stderr)
+            print(json.dumps(e.body), file=options.stderr)
             raise SystemExit(2)
 
         except Exception as e:
--- a/src/magic_folder/api_cli.py
+++ b/src/magic_folder/api_cli.py
@@ -30,6 +30,7 @@ from twisted.internet.defer import (
 from .cli import (
     _default_config_path,
     load_global_configuration,
+    to_unicode,
 )
 from .client import (
     CannotAccessAPIError,
@@ -69,7 +70,7 @@ def add_snapshot(options):
 
 class DumpStateOptions(usage.Options):
     optParameters = [
-        ("folder", "n", None, "Name of the magic-folder whose state to dump"),
+        ("folder", "n", None, "Name of the magic-folder whose state to dump", to_unicode),
     ]
 
     def postOptions(self):
--- a/src/magic_folder/client.py
+++ b/src/magic_folder/client.py
@@ -168,6 +168,20 @@ class MagicFolderClient(object):
             'scan_interval': scan_interval,
         }, ensure_ascii=False).encode('utf-8'))
 
+    def leave_folder(self, magic_folder, really_delete_write_capability):
+        # type: (unicode, bool) -> dict
+        api_url = self.base_url.child(u"v1").child(u"magic-folder").child(magic_folder)
+        return self._authorized_request(
+            "DELETE",
+            api_url,
+            body=json.dumps(
+                {
+                    "really-delete-write-capability": really_delete_write_capability,
+                },
+                ensure_ascii=False,
+            ).encode("utf-8"),
+        )
+
     @inlineCallbacks
     def _authorized_request(self, method, url, body=b""):
         """
--- a/src/magic_folder/client.py
+++ b/src/magic_folder/client.py
@@ -75,33 +75,41 @@ class MagicFolderApiError(ClientError):
     A Magic Folder HTTP API returned a failure code.
     """
     code = attr.ib()
+    reason = attr.ib()
     body = attr.ib()
 
+    @property
+    def reason(self):
+        return self.body.get("reason")
+
     def __repr__(self):
-        return "<MagicFolderApiError code={} body={!r}>".format(
+        return "<MagicFolderApiError code={} reason={!r} body={!r}>".format(
             self.code,
+            self.reason,
             self.body,
         )
 
     def __str__(self):
-        return u"Magic Folder HTTP API reported error {}: {}".format(
+        extra_fields = {k: v for (k, v) in self.body.items() if k != "reason"}
+        return u"Magic Folder HTTP API reported error {}: {}{}".format(
             self.code,
-            self.body,
+            self.reason,
+            " ({})".format(extra_fields) if extra_fields else "",
         )
 
 
 @inlineCallbacks
-def _get_content_check_code(acceptable_codes, res):
+def _get_json_check_code(acceptable_codes, res):
     """
     Check that the given response's code is acceptable and read the response
     body.
 
     :raise MagicFolderApiError: If the response code is not acceptable.
 
-    :return Deferred[bytes]: If the response code is acceptable, a Deferred
-        which fires with the response body.
+    :return Deferred[Any]: If the response code is acceptable, a Deferred
+        which fires with the parsed response body.
     """
-    body = yield res.content()
+    body = yield res.json()
     if res.code not in acceptable_codes:
         raise MagicFolderApiError(res.code, body)
     returnValue(body)
@@ -181,9 +189,8 @@ class MagicFolderClient(object):
                 "Can't reach the magic folder daemon at all"
             )
 
-        body = yield _get_content_check_code([http.OK, http.CREATED], response)
-        # all responses should contain JSON
-        returnValue(json.loads(body))
+        body = yield _get_json_check_code([http.OK, http.CREATED], response)
+        returnValue(body)
 
 
 @implementer(IAgentEndpointFactory)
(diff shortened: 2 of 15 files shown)


src/magic_folder/client.py
```diff
@@ -181,8 +199,7 @@ def _authorized_request(self, method, url, body=b""):
             )
 
         body = yield _get_content_check_code([http.OK, http.CREATED], response)
-        # all responses should contain JSON
-        returnValue(json.loads(body))
+        returnValue(body)
```
meejah, line 202: now we won't validate that it's actually-json?

src/magic_folder/api_cli.py
```diff
@@ -401,7 +402,7 @@ def run_magic_folder_api_options(options):
 
         except MagicFolderApiError as e:
             # these kinds of errors should report via JSON from the endpoints
-            print(u"{}".format(e.body), file=options.stderr)
+            print(e, file=options.stderr)
```
meejah, line 405: before this change, all output from magic-folder-api was json (idea is that's easier for other programs to integrate with).

src/magic_folder/test/test_web.py
```diff
@@ -598,6 +528,281 @@ def test_list_folders(self, folders):
...
+                ),
+            ),
+        )
+
+    @given(
+        folder_names(),
+    )
+    def test_leave_folder_fail(self, folder_name):
+        """
+        A request for **DELETE /v1/magic-folder/<folder-name>** succeeds.
```
meejah, line 622: mismatches with test name

src/magic_folder/test/test_web.py
```diff
@@ -598,6 +528,281 @@ def test_list_folders(self, folders):
...
+            authorized_request(
+                node.http_client,
+                AUTH_TOKEN,
+                "DELETE",
+                self.url.child(folder_name),
+                b"{}",
+            ),
+            succeeded(
+                matches_response(
+                    code_matcher=MatchesAny(
```
meejah, line 653: doesn't need matchesany (just one thing)

meejah (commented):
i think it's best if the output of `magic-folder-api` remains json (so at least other tools can rely on parsing it).

src/magic_folder/api_cli.py
tomprince: Not always. The next exception handler doesn't nor does `dump-state`. Also, before this PR, several of the commands would return HTML if given a folder that doesn't exist (it turns out `NoResource` returns HTML with the message passed to it).

That said, I do think having an option for JSON output may make sense, but I'm not I'm not sure that that it should be the default. That said, I don't feel strongly about this change (though something needs to be adjusted, as `MagicFolderApiError` has a parsed version, with `reason` broken out into it's own attribute, at the moment.

src/magic_folder/client.py
tomprince: I've changed `_get_content_check_code` above to return JSON as parsed by treq (largely because I wanted to pass the parsed version to `MagicFolderApiError` in the error case.

tomprince: I've updated the name of the function to make that clear.

tomprince:
It looks like the windows error is us falling afoul of the inability to delete open files on windows; specifically, the folder state database. The obvious solution of calling `.close()` on the database doesn't work, at least in the unit tests. I'm adding some extra logging in a branch to try an track down where other connections to it might be being opened.

src/magic_folder/api_cli.py
meejah: I already proposed a change, just merge my pr-to-this-pr.

meejah: Sure, `dump-state` doesn't do json by default (maybe it should?). But the commands that _do_ will now _sometimes_ return not-json, which is bad for integration.

meejah: Bigger-picture here is that `magic-folder-api` is _supposed_ to be a fairly thin wrapper over the HTTP API and is _supposed_ to produce JSON output (e.g. https://github.com/LeastAuthority/magic-folder/issues/376 ). This change takes us further than that.

tomprince: I've reverted this to output json.

tomprince:
> It looks like the windows error is us falling afoul of the inability to delete open files on windows; specifically, the folder state database. The obvious solution of calling `.close()` on the database doesn't work, at least in the unit tests. I'm adding some extra logging in a branch to try an track down where other connections to it might be being opened.

I've added a cache of `MagicFolderConfig` objects, so we only ever have a single connection, which we close before deleting the file.

meejah:
please don't force-push to branches that already have comments; then GitHub loses their spot.

tomprince:
> please don't force-push to branches that already have comments; then GitHub loses their spot.

As I have said before, I find it hard to work on a branch, without keeping a clean commit history, which requries force-pushing. Looking at the comments at least on the discussion page (as opposed to the diff page), it looks like the comments are still have the context they had.

meejah:
Hmm, maybe they fixed something (and it was just the "resolved" that caused them to disappear)

meejah:
You can still rebase "stuff that's yet to push" without messing with already-pushed commits...

tomprince:
> You can still rebase "stuff that's yet to push" without messing with already-pushed commits...

That doesn't give me an easy way to see https://github.com/LeastAuthority/magic-folder/pull/475/commits/8ec259e371cd283ad073ba1500de5f34f553683b with the changes that you requested on its own, since it would then be split up into one commit where I made the original change, and then a different commit where I then change all the code in it. I split the code up into small commits, because I find it easier to reason about. Not changing those commits to reflect changes loses the ability to reason about those small chunks.

meejah:
> That doesn't give me an easy way to see 8ec259e with the changes that you requested on its own, since it would then be split up into one commit where I made the original change, and then a different commit where I then change all the code in it

On the flip side, not squashing those gives reviewers an easier way to see what changed as a result of a review.

tomprince:
It looks like the integration tests are failing due to the windows fix. Investigating.

tomprince:
@meejah  This is ready for review again.

src/magic_folder/test/test_web.py
```diff
@@ -528,6 +532,282 @@ def test_list_folders(self, folders):
...
```
meejah, line 611: Only one thing, just `Equals()`?

meejah (approved):
some inline fixups, then looks fine

src/magic_folder/service.py
```diff
@@ -262,3 +263,31 @@ def create_folder(self, name, author_name, local_dir, poll_interval, scan_interv
...
+            if folder.config.is_admin():
+                if not really_delete_write_capability:
+                    raise APIError(
+                        code=http.CONFLICT,  # FIXME?
+                        reason="magic folder '{}' holds a write capability"
+                        ", not deleting.".format(name),
+                    )
+
+            yield folder.disownServiceParent()
+            fails = self.config.remove_magic_folder(name)
```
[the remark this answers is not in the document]
tomprince: This is existing behavior. I don't have strong feelings about it, but I'd rather not change it here.

src/magic_folder/test/test_web.py
```diff
@@ -270,14 +272,16 @@ def treq_for_folders(
         reactor, basedir, auth_token, folders, start_folder_services, tahoe_client
     ).http_client
 
-def magic_folder_config(author_name, local_directory):
+
+def magic_folder_config(author_name, local_directory, admin=True):
```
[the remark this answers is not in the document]
tomprince: This is to support testing the existing `--really-delete-write-capability` CLI option, or rather, testing not specifying that option.

src/magic_folder/config.py
```diff
@@ -1199,6 +1201,11 @@ def collective_dircap(self, cursor):
         cursor.execute("SELECT collective_dircap FROM config")
         return cursor.fetchone()[0].encode("utf8")
 
+    @collective_dircap.setter
+    @with_cursor
+    def collective_dircap(self, cursor, dircap):
```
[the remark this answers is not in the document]
tomprince: This is to support testing the existing `--really-delete-write-capability` CLI option, or rather, testing not specifying that option.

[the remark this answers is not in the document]
meejah: The setter should still probably check that it _is_ a dircap .. and probably a comment regarding "this is for testing"?
(We should never "set" these, really, is the point: once a thing is created that's it .. be just haven't put in the "invite/join" stuff yet)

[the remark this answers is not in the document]
tomprince: I did add a check. See #504 for a comment.

PR #477, opened 2021-07-07
Additional per-folder information

meejah:
Progress on #440, adding more per-folder information including some history items.

### DIFF
--- a/src/magic_folder/downloader.py
+++ b/src/magic_folder/downloader.py
@@ -40,6 +40,7 @@ from twisted.python.filepath import (
 )
 from twisted.internet.defer import (
     DeferredLock,
+    DeferredList,
     returnValue,
     succeed,
 )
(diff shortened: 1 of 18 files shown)


src/magic_folder/downloader.py
```diff
@@ -56,8 +56,17 @@
...
-from .util.file import PathState, get_pathinfo, seconds_to_ns
-from .util.twisted import exclusively
+from .util.file import (
+    PathState,
+    get_pathinfo,
+    seconds_to_ns,
+)
+from .util.twisted import (
+    exclusively,
+)
```
tomprince, line 66: I'd slightly prefer to leave these as single lines. We've talked about (when we aren't under a deadline) using `black` and `isort` on the codebase. I've been trying (as much as practical), to at least format new code that way, to minimize the diff when we do. The original is how black would format these lines.

meejah: okay, but all the rest is formatted like that. consistency is good.

PR #478, opened 2021-07-07
Reveal size, mtime of all files in a folder

### LINKED ISSUE #476: Add HTTP API to retrieve all file statuses for a given folder
GridSync needs, at minimum, the size of every file in a folder in order to be able to ask `ZKAPAuthorizer` how many storage-time credits will be used to store those files.

Add an endpoint returning the status of every file in a folder. Currently, that status only needs to include the size.

Something like `GET /v1/magic-folder/<folder-name>/file-status` or similar. Returning data like:

```
{
    "rel/path": {
        "size": 1234
    }
}
````

### PR DESCRIPTION
**Pull request by meejah**
fixes #476

### DIFF
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -1144,11 +1144,13 @@ class MagicFolderConfig(object):
             """,
         )
 
-        for row in cursor.fetchall():
-            yield (
+        return [
+            (
                 row[0],  # name
                 PathState(mtime_ns=row[1], ctime_ns=row[2], size=row[3]),
             )
+            for row in cursor.fetchall()
+        ]
 
     @property
     @with_cursor
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -1132,6 +1132,7 @@ class MagicFolderConfig(object):
 
         :returns Iterable[(unicode, PathState)]: an iterable of
             2-tuples of (name, PathState instance), one for each file
+            (ordered by timestamp)
         """
         cursor.execute(
             """
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -723,6 +723,7 @@ def _construct_local_snapshot(identifier, name, author, content_paths, metadata,
         ),
     )
 
+
 @attr.s
 class MagicFolderConfig(object):
     """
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/magic_folder/config.py [hunk 1]
```diff
@@ -1123,6 +1124,32 @@ def get_currentsnapshot_pathstate(self, cursor, name):
...
+                name, mtime_ns, ctime_ns, size
+            FROM
+                current_snapshots
+            ORDER BY
+                mtime_ns DESC
+            """,
+        )
+
+        for row in cursor.fetchall():
+            yield (
```
tomprince, line 1148: Don't `yield` in a function decorated by `with_cursor`, it doesn't do what you expect (the transaction is closed before any code in the function runs).

In any case, I don't think we should be holding open a transaction while the generator runs. In this case, it isn't an issue, since it consumed synchronously without other calls to `MagicFolderConfig`, but it seems like the kind of thing that will cause unexpected surprises down the line.

src/magic_folder/config.py
```diff
@@ -1123,6 +1124,32 @@ def get_currentsnapshot_pathstate(self, cursor, name):
...
+            2-tuples of (name, PathState instance), one for each file
+        """
+        cursor.execute(
+            """
+            SELECT
+                name, mtime_ns, ctime_ns, size
+            FROM
+                current_snapshots
+            ORDER BY
+                mtime_ns DESC
```
tomprince, line 1143: It is probably worth a mention in the docstring about this ordering. My guess of ordering would be by name, rather than mtime.

src/magic_folder/web.py
```diff
@@ -428,6 +432,24 @@ def all_folder_configs():
...
+        """
+        Render status information for every file in a given folder
+        """
+        _application_json(request)  # set reply headers
+        folder_config = self._global_config.get_magic_folder(folder_name)
+
+        return json.dumps([
+            {
+                "name": name,
+                "mtime": ns_to_seconds(ps.mtime_ns),
```
tomprince, line 446: I was considering suggesting that this should be a method on `PathState`, but I guess maybe it would be to similarly named to `.mtime_ns`, and apt to cause confusion or error. (Certainly it shouldn't be a property for that reason)

tomprince (approved):
This looks good once get_all_current_snapshot_pathstates is changed to return a list or dict. I agree that it would be better to avoid allocating one, but I can't think of an easy

src/magic_folder/config.py
tomprince: I added a check to #470 that will detect and error in this case.

PR #485, opened 2021-07-13
Add more logging to help debug test_create_then_recover integration failures

### LINKED ISSUE #491: There is little visibility into what the downloader is doing w.r.t. updating local files
As reported in #484 there is an integration tests that fails intermittently.  It fails with a timeout while waiting to see updated content on the filesystem.  It expects it to be written there by the downloader after the downloader finds the content on the grid.

When the test fails, it's hard to make sense of what has gone wrong.  Did the downloader write the content to the wrong place?  Did the downloader break and fail to download it at all?  Is the downloader making progress towards that goal but needs more time (if so, what is taking it so long)?  Or is the downloader actually working fine and the test has incorrect assumptions or requirements?

It would be great to augment the downloader with some more logging so that these questions could be answered by examining a log after the fact.

exarkun:
Fixes #491

### DIFF
--- a/src/magic_folder/downloader.py
+++ b/src/magic_folder/downloader.py
@@ -366,7 +366,12 @@ class MagicFolderUpdater(object):
 
             self._status.download_started(self._config.name, relpath)
             try:
-                staged = yield self._magic_fs.download_content_to_staging(snapshot, self.tahoe_client)
+                with start_action(
+                    action_type=u"downloader:updater:content-to-staging",
+                    name=snapshot.name,
+                    capability=snapshot.capability,
+                ):
+                    staged = yield self._magic_fs.download_content_to_staging(snapshot, self.tahoe_client)
             finally:
                 self._status.download_finished(self._config.name, relpath)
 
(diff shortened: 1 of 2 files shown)


src/magic_folder/downloader.py
```diff
@@ -460,6 +467,15 @@ def mark_overwrite(self, remote_snapshot, staged_content):
...
         """
+        # Could be items associated with the action but then we have to define
+        # an action type to teach Eliot how to serialize the remote snapshot
+        # and the FilePath.  Probably *should* do that but just doing this for
+        # now...  Maybe it should be /easier/ to do that?
+        Message.log(
+            message_type=u"downloader:filesystem:mark-overwrite",
+            content_relpath=magic2path(remote_snapshot.name),
+            staged_content_path=staged_content.path,
+        )
```
tomprince, line 478: I added support for `FilePath` in #474 (following the suggestion in https://github.com/itamarst/eliot/issues/374#issuecomment-467999103 to use a custom JSON encoder). That doesn't help for the mangled path, though.

For that, we should do one or both of:
- stop using mangled paths except at the point of writing to a tahoe-dir node
- provide a type for relative paths, rather than just using strings everywhere.

exarkun: A type for relative paths sounds like a really good idea.

exarkun:
I retargetted this at #491 and I think it's plausibly done now.

meejah (approved):

Issue #420, opened 2021-06-14
Add a version indicator to the various DMDs we create.


tomprince:
I can imagine that at some point in the future, we'll want to change the structure of DMDs, or add extra non-snapshot data to them. We should add version metadata to them (perhaps a `@version` file, which won't conflict with any of the snapshot entries, see #395).

We can then check that version before doing anything with a remote DMD.

It probably also makes sense to include the version in the snapshot metadata, for similar reasons.

meejah:
We already have a version in the snapshot metadata.

meejah:
I don't think we need an immediate version now. If we do need a second version in the future, we can add one using an "impossible" / invalid-snapshot-name like you suggest and that can be "version 2" essentially.

..or, we could just add additional information that way and "never" need a version for the directories themselves. Or we could use the Tahoe metadata for this. 

So closing this in favour of doing #395 and nothing else.

tomprince:
We'll need to ensure that we notice a file with an "invalid" name, before doing any processing on the folder, so that we notice that we have an incompatible folder before doing anything with it.

tomprince:
To expand on that, if we depend on just #395, we need to somehow ensure that the entry we use gets processed before anything else. Looking at the current code, the downloader appears to use a dictionary for files, which at least on py2.7 will be in a random order, and I don't think the uploader ever looks at the DMD, so we have no way to reliably notice a newer version of the DMD in either of the processes.

meejah:
> We'll need to ensure that we notice a file with an "invalid" name,

That is addressed by #395 

> To expand on that, if we depend on just #395, we need to somehow ensure that the entry we use gets processed before anything else. 

Yes, _if_ we introduce a second version or anything else, we'd need to add code (for that plus probably other things). We don't need that code now (nor do we need a version marker now).

tomprince:
> That is addressed by #395 

We might only notice it after we made changes to either the DMD or the file system, which might cause issues.

meejah:
Are you saying it's _impossible_ to add `@version` or any other "special" thing later?

tomprince:
> Are you saying it's _impossible_ to add `@version` or any other "special" thing later?

I'm saying it is impossible[1] to add it in such a way that the current version will notice that it is dealing with a later version early enough to avoid doing something with the DMD that is inappropriate because we've changed the format in some incompatible way.

[...]

PR #487, opened 2021-07-13
Move tahoe-using logic out of the web service.

### LINKED ISSUE #486: Move tahoe using logic out of the web service.
As part of #420, I need to add some check when accessing the collective and individual DMDs we have. Because the webservice reaches into the config, and calls `participants_from_collective` directly, I'd need to add duplicate logic there to handle the failure.

Instead, I think it makes sense to keep the business logic on the folder service, and call it from the web service (so that the web service just deals with serialization).

There is other business logic in the web service that would make sense to move out as well, I think, but, as it deals with serializing somewhat complex data, that we don't have a good in-process abstraction currently, the interface between the folder/global service and the web service isn't obvious, so I don't want to block on that here.

tomprince:
Fixes #486.

This moves all the business logic involving tahoe out from the web service. In particular, it makes the web service avoid creating an `IParticipants` itself, so that error handling for #420 doesn't need to be duplicated there.

### DIFF
--- a/src/magic_folder/magic_folder.py
+++ b/src/magic_folder/magic_folder.py
@@ -7,9 +7,11 @@ from __future__ import (
 
 import six
 
-from twisted.python.filepath import FilePath
+from twisted.python.filepath import FilePath, InsecurePath
 from twisted.internet import defer
+from twisted.internet.defer import Deferred
 from twisted.application import service
+from twisted.web import http
 
 from eliot import (
     Field,
@@ -17,6 +19,7 @@ from eliot import (
     MessageType,
 )
 
+from .common import APIError
 from .util.eliotutil import (
     RELPATH,
     validateSetMembership,
@@ -36,6 +39,7 @@ from .downloader import (
     LocalMagicFolderFilesystem,
 )
 from .participants import (
+    IParticipant,
     participants_from_collective,
 )
 from .scanner import (
@@ -185,6 +189,36 @@ class MagicFolder(service.MultiService):
             been snapshotted.
         """
         return self.scanner_service.scan_once()
+
+    def participants(self):
+        # type: () -> Deferred[list[IParticipant]]
+        """
+        List all participants of this folder
+        """
+        return self._participants.list()
+
+    def add_participant(self, author, participant_directory):
+        return self._participants.add(author, participant_directory)
+
+    def add_snapshot(self, relative_path):
+        # type: (unicode) -> Deferred[None]
+        """
+        Create a new snapshot of the given file.
+        """
+
+        # preauthChild allows path-separators in the "path" (i.e. not
+        # just a single path-segment). That is precisely what we want
+        # here, though. It sill does not allow the path to "jump out"
+        # of the base magic_path -- that is, an InsecurePath error
+        # will result if you pass an absolute path outside the folder
+        # or a relative path that reaches up too far.
+        try:
+            path = self.config.magic_path.preauthChild(relative_path)
+        except InsecurePath as e:
+            return defer.fail(
+                APIError.from_exception(http.NOT_ACCEPTABLE, e)
+            )
+        return self.local_snapshot_service.add_file(path)
 
 
 _NICKNAME = Field.for_types(
--- a/src/magic_folder/uploader.py
+++ b/src/magic_folder/uploader.py
@@ -21,6 +21,7 @@ from twisted.internet.defer import (
 from twisted.internet.task import (
     LoopingCall,
 )
+from twisted.web import http
 from zope.interface import (
     Interface,
     implementer,
@@ -34,6 +35,7 @@ from eliot import (
 from eliot.twisted import (
     inline_callbacks,
 )
+from .common import APIError
 from .util.eliotutil import (
     RELPATH,
     log_call_deferred,
@@ -228,19 +230,21 @@ class LocalSnapshotService(service.Service):
             self._status.upload_queued(relpath)
         except ValueError:
             ADD_FILE_FAILURE.log(relpath=path.path) #FIXME relpath
-            raise ValueError(
-                "The path being added '{!r}' is not within '{!r}'".format(
+            raise APIError(
+                reason=u"The path being added '{!r}' is not within '{!r}'".format(
                     path.path,
                     self._config.magic_path.path,
-                )
+                ),
+                code=http.NOT_ACCEPTABLE,
             )
 
         # isdir() can fail and can raise an appropriate exception like
         # FileNotFoundError or PermissionError or other filesystem
         # exceptions
         if path.asBytesMode('utf-8').isdir():
-            raise ValueError(
-                "expected a regular file, {!r} is a directory".format(path.path),
+            raise APIError(
+                reason=u"expected a regular file, {!r} is a directory".format(path.path),
+                code=http.NOT_ACCEPTABLE,
             )
 
         # add file into the queue
--- a/src/magic_folder/client.py
+++ b/src/magic_folder/client.py
@@ -250,7 +250,7 @@ def create_http_client(reactor, api_client_endpoint_str):
 
 # See https://github.com/LeastAuthority/magic-folder/issues/280
 # global_service should expect/demand an Interface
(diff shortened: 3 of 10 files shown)


exarkun:
This seems to contain some of the same changes as I've seen elsewhere (eg the `NoSuchMagicFolder` exception).  I'm not sure what the right diff to look at to avoid seeing those duplications is.

tomprince:
> This seems to contain some of the same changes as I've seen elsewhere (eg the `NoSuchMagicFolder` exception). I'm not sure what the right diff to look at to avoid seeing those duplications is.

I've rebased this on `main` so the duplicated code diff should be gone.

src/magic_folder/magic_folder.py
```diff
@@ -175,6 +179,36 @@ def ready(self):
...
+        """
+        List all participants of this folder
+        """
+        return self._participants.list()
+
+    def add_participant(self, author, participant_directory):
+        return self._participants.add(author, participant_directory)
+
+    def add_snapshot(self, relative_path):
+        # type: (unicode) -> Deferred[None]
```
meejah, line 193: probably better in the docstring, to document the params

src/magic_folder/magic_folder.py
```diff
@@ -175,6 +179,36 @@ def ready(self):
...
+        # just a single path-segment). That is precisely what we want
+        # here, though. It sill does not allow the path to "jump out"
+        # of the base magic_path -- that is, an InsecurePath error
+        # will result if you pass an absolute path outside the folder
+        # or a relative path that reaches up too far.
+        try:
+            path = self.config.magic_path.preauthChild(relative_path)
+        except InsecurePath as e:
+            return defer.fail(
+                APIError.from_exception(http.NOT_ACCEPTABLE, e)
```
meejah, line 208: The ticket says you want to remove tahoe-code from the Web code .. but now there's Web code mixed into the magic-folder code; this should raise something other than a web-specific exception.

src/magic_folder/testing/web.py
```diff
@@ -413,9 +413,14 @@ def render_GET(self, request):
         capability = None
         for arg, value in uri.query:
             if arg == u"uri":
+                if capability is not None:
+                    raise Exception("Can't sepecify uri twice in query args.")
```
meejah, line 417: This isn't what Tahoe will do, I don't think -- and anyway isn't used (according to codecov) so just delete.

src/magic_folder/testing/web.py
```diff
@@ -413,9 +413,14 @@ def render_GET(self, request):
...
                 capability = value
+
         # it's legal to use the form "/uri/<capability>"
-        if capability is None and request.postpath and request.postpath[0]:
+        if request.postpath and request.postpath[0]:
+            if capability is not None:
+                raise Exception("Can't specify uri in path and query.")
```
meejah, line 423: ditto

src/magic_folder/uploader.py
```diff
@@ -230,19 +232,21 @@ def add_file(self, path):
             self._status.upload_queued(self._config.name, relpath)
         except ValueError:
             ADD_FILE_FAILURE.log(relpath=path.path) #FIXME relpath
-            raise ValueError(
-                "The path being added '{!r}' is not within '{!r}'".format(
+            raise APIError(
```
meejah, line 235: Same comment per spreading Web code .. a non-Web-specific error should be raised here.

meejah (changes requested):
This moves several pieces of Web-specific code outside of `web.py` (which I think is bad).

`participants_from_collective()` seems like the right way to get the abstraction that represents participants .. so where else other than there do you think a check needs to be inserted? There isn't really enough context in #487 to learn what problem this is really solving (where are the other places a check might be needed?)

tomprince (commented):
> This moves several pieces of Web-specific code outside of web.py (which I think is bad).

As I expand on below, I disagree that that is what is going on.

> participants_from_collective() seems like the right way to get the abstraction that represents participants .. so where else other than there do you think a check needs to be inserted?

I agree that `participants_from_collective` is where the checks need to be. But having that call multiple places means we need to do the check multiple times (currently, it would be every time any of these endpoints are called). The motivation for this change is (partly) so that we re-use the existing participant we have, rather getting a new object each call. 

The other aspect is to move both logic, and knowledge of how `MagicFolder` is broken down into individual services.

src/magic_folder/testing/web.py
tomprince: I'd be find moving them to a separate PR so that it doesn't block this one.

src/magic_folder/magic_folder.py
meejah: Yes, it's the HTTP error-code that seems web-specific. I guess if we're _already_ scattering HTTP codes throughout other code then this is fine.

src/magic_folder/testing/web.py
meejah: well, it's un-tested code and arguably this stuff _should_ represent "whatever Tahoe does" (even if you think Tahoe is wrong) .. right?

src/magic_folder/magic_folder.py
tomprince: Yeah, we've already done that. I'd prefer to not have that, but I think (at least until the code base is more settled), having that at the location of the error makes more sense. The other option would be to have a bunch of hyper-specific exception classes, and map them to error codes either in the exception (which is *really* all that different than what we have now) or have some sort of table in the web code. The latter feels like unnecessary maintenance overhead at this point.

We can do things like https://github.com/LeastAuthority/magic-folder/blob/039a99443b4f0404bf27d0c786e62448579e1203/src/magic_folder/common.py#L109-L127 if there are common errors, but I think most of the explicit uses of `APIError` are all more-or-less one-offs.

src/magic_folder/testing/web.py
tomprince: > should represent "whatever Tahoe does"

I would rather say, it should ensure that we are using the API correctly. That is very often done by making the test double behave the same as the real thing. But I think making the test double more pedantic is also a reasonable thing to do.

meejah: Fine, if you think that's required it at least needs a test.

meejah:
so the big-picture here is that this refactoring is because of performance concerns with doing the check too many times..?

tomprince:
> so the big-picture here is that this refactoring is because of performance concerns with doing the check too many times..?

Partly that (there are network round trips involved in doing the checks), but also that the API calls I am changing aren't the right place to be *reporting* those errors. We should report those errors on startup once, and not every time somebody makes a call to an API that happens to need access to an `IParticipants`.

That said, that is why I am making the change *now*. But I think it is a good change regardless. I think the web code should know as little as possible about how magic folder works; that is, each method should be as close to deserialize, call a method, serialize results. Creating, leaving and scanning a folder are already implemented that way; this moves a few more methods to that style. I'd like eventually to move every method to that.

The methods I didn't move here (that existed when I wrote this patch), all returned complex data that wasn't immediately obvious how to split between business logic and presentation, so I deferred them; otherwise I'd've converted everything.

src/magic_folder/testing/web.py
tomprince: I've removed these changes from the PR.

meejah:
> We should report those errors on startup once, and not every time somebody makes a call to an API that happens to need access to an IParticipants.

That seems unlikely to be correct, since the device with the writecap could have written new stuff to the participants collective at any time .. and it could have received updated software meantime and written a new format (which, it seems, is the entire point of putting a version check in the DMD-accessing code).

tomprince:
> > We should report those errors on startup once, and not every time somebody makes a call to an API that happens to need access to an IParticipants.
> 
> That seems unlikely to be correct, since the device with the writecap could have written new stuff to the participants collective at any time .. and it could have received updated software meantime and written a new format (which, it seems, is the entire point of putting a version check in the DMD-accessing code).

Sure, we should check collective-DMD version every time we access it (at least if we don't hold the write-cap). Similarly for all the peers DMDs. I don't think we need to that for our personal-DMD every time, and perhaps not for the collective-DMD if we hold the write-cap for it.

PR #488, opened 2021-07-14
Refactor `treq_for_folders` to allow access to the database as well.

tomprince:
This moves the body of that function to `MagicFolderNode.create`, which creates an object with attributes for treq client, service, and global config.

This is such a helpful fixture, that I've used in in #463, #475, and #487. It would be nice to get at least this landed.

### DIFF
--- a/src/magic_folder/test/common.py
+++ b/src/magic_folder/test/common.py
@@ -13,6 +13,7 @@ __all__ = [
     "flush_logged_errors",
     "skip",
     "skipIf",
+    "success_result_of",
 ]
 
 import os
@@ -56,6 +57,7 @@ from twisted.internet.interfaces import (
 )
 from twisted.internet.endpoints import AdoptedStreamServerEndpoint
 from twisted.python import log
+from twisted.trial.unittest import SynchronousTestCase as _TwistedSynchronousTestCase
 
 from allmydata import uri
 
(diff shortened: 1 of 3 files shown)


src/magic_folder/test/test_web.py
```diff
@@ -376,7 +284,7 @@ class MagicFolderTests(SyncTestCase):
 
     def setUp(self):
         super(MagicFolderTests, self).setUp()
-        self.author = create_local_author(u"alice")
+        self.author = u"alice"
```
meejah, line 287: maybe `.author_name` since it's just a string now?

src/magic_folder/test/common.py
```diff
@@ -365,3 +366,6 @@ class AsyncBrokenTestCase(_TestCaseMixin, TestCase):
     run_tests_with = EliotLoggedRunTest.make_factory(
         AsynchronousDeferredRunTestForBrokenTwisted.make_factory(timeout=60.0),
     )
+
+_TWISTED_TEST_CASE = _TwistedSynchronousTestCase()
+success_result_of = _TWISTED_TEST_CASE.successResultOf
```
meejah, line 371: this smells a little weird; at least worth a comment explaining why it's required. I also don't see `_TWISTED_TEST_CASE` used anywhere else, so can it be dropped?

meejah (approved):
yeah not having access to the config/db is annoying for sure.
LGTM, some inline comments

src/magic_folder/test/common.py
tomprince: > I also don't see `_TWISTED_TEST_CASE` used anywhere else, so can it be dropped?

I was vaguely thinking that `failureResultOf` and `assertNoResult` might also be useful, and would re-use the dummy object. The use case isn't as clear for those though (and isn't needed for this PR), so I didn't included them.

PR #489, opened 2021-07-14
Use `magic_folder.util.capabilities` everywhere.

tomprince:
While working on #420, I went updated all the direct uses of `allmydata.uri` to use our wrappers instead.

I left the use in `magic_folders.testing.web` (and it's tests), as the code is currently shared with tahoe-lafs. I also left some dead uses in `magic_folder.test.common`, which I'll remove, with a bunch of other dead code in a separate PR.

### DIFF
--- a/src/magic_folder/test/test_upload.py
+++ b/src/magic_folder/test/test_upload.py
@@ -57,6 +57,7 @@ from ..snapshot import (
 from ..magicpath import (
     path2magic,
 )
+from ..util.capabilities import is_immutable_directory_cap, to_verify_capability
 from ..util.file import PathState
 from twisted.internet import task
 
@@ -77,10 +78,6 @@ from magic_folder.tahoe_client import (
     TahoeAPIError,
 )
 
-from allmydata.uri import (
-    is_uri,
-    from_string as uri_from_string,
-)
 
 class RemoteSnapshotCreatorTests(SyncTestCase):
     """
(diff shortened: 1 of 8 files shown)


src/magic_folder/test/test_upload.py
```diff
@@ -156,7 +153,7 @@ def test_commit_a_file(self, mangled_name, content, upload_dircap):
                 mangled_name: [
                     u"dirnode", {
                         u"ro_uri": remote_snapshot_cap.decode("utf-8"),
-                        u"verify_uri": uri_from_string(
+                        u"verify_uri": tahoe_uri_from_string(
```
meejah, line 156: This kind of just "hides" that we're still using the Tahoe API here? (I mean: we're importing from our own namespace, sure, but it's still relying on the Tahoe APIs here). I guess we can still search for "tahoe_uri_from_string" to get rid of it later, so meh ... but the "real" solution is `to_verify_capability()` or similar.

Bigger picture here is that _ideally_ there'd be a library (provided by Tahoe, ideally) that _just_ does capability-string manipulation so we're kind-of prototyping the API we want in `magic_folder.util.capabilities` essentially...

src/magic_folder/test/test_upload.py
```diff
@@ -170,7 +167,7 @@ def test_commit_a_file(self, mangled_name, content, upload_dircap):
         # test whether we got a capability
         self.assertThat(
             remote_snapshot_cap,
-            MatchesPredicate(is_uri,
+            MatchesPredicate(is_immutable_directory_cap,
                              "%r is not a Tahoe-LAFS URI"),
```
meejah, line 171: minor, but the text should maybe be upgraded too (e.g. "not a Tahoe-LAFS immutable-directory URI")

meejah (approved):
looks fine, couple inline comments.
address those or not, and the failing builders plz then merge :)

src/magic_folder/test/test_upload.py
tomprince: Yeah, if this wasn't in test code, I'd have definitely add `to_verify_capability`. It probably makes sense to do that anyway.

PR #490, opened 2021-07-14
Dead test code

tomprince:
I discovered this while looking at the uses for `allmydata.uri` for #489.

`magic_folder.test.common_util` had only one used function, which I moved to `magic_folder.test.cli.test_magic_folder` which was the only place it was used.


exarkun:
I would :fireworks: and merge right now but the lint job is failing.

PR #492, opened 2021-07-15
Fix docker image builds.

### LINKED ISSUE #496: Fix docker image build.
It turns out that #437 broke our docker builds, since the `pip wheel` command to populate a wheelhouse failed. This is due to setuptools not being pinned in the requirements files.

tomprince:
Fixes #496.

The current version of pip-compile doesn't pin `setuptools` by default. This caused the docker image builds to fail, when trying to pre-populate a wheelhouse of python deps.

This add the `--allow-unsafe` (which will become the default) to the `pip-compile-multi` invocation, to include them. It also removes the manual pins that we had in our tox configuration, since they are redundant with the automatic pinning we now had; this is necessary, since otherwise the version of setuptools for the `base.in` and `tox.in` would conflict.


tox.ini
```diff
@@ -153,7 +153,8 @@ commands =
 	     --remove \
 	     requirements/base.txt \
 	     requirements/platform.txt
-         pip-compile-multi -g base -g tox -g test
+         # --allow-unsafe is a future default
```
exarkun, line 156: I'm not sure what this comment is meant to help me understand.

exarkun (approved):
If this fixes the image building jobs then it looks fine to me.

tox.ini
tomprince: It is explaining why we are passing something an option with `unsafe` in the name.

PR #495, opened 2021-07-15
481.status error reporting

### LINKED ISSUE #481: Status API: error reporting
At least some kinds of errors should be reported to UIs via the status API:

- downloader problems, like: can't write file (permissions, disk full)
- uploader problems:
  - can't contact tahoe
  - tahoe gave us an error (investigate: what happens on ZKAP-specific errors)

meejah:
fixes #481 

Report some kinds of errors via the status API. The first two such errors are:
- downloader fails to download a snapshot
- downloader fails to move a file into position


src/magic_folder/downloader.py
```diff
@@ -372,7 +379,14 @@ def _process(self, snapshot):
...
                 ):
-                    staged = yield self._magic_fs.download_content_to_staging(snapshot, self.tahoe_client)
+                    try:
+                        staged = yield self._magic_fs.download_content_to_staging(snapshot, self.tahoe_client)
+                    except Exception:
+                        raise PublicError(
+                            self._clock.seconds(),
+                            "Failed to download snapshot for '{}'.".format(relpath)
+                        )
+
```
tomprince, line 389: My first inclination would be that it is would be better directly report the error here (by calling `.error_occured` and then re-raising the original exception. After thinking about it further, I think that makes sense, as we may not want to let this error propagate to the top-level; in this particular case, we may want to continue scanning the collective, even if we can't update one file.

As an aside, I think this case differs from the `APIError` case, in that in the `APIError` case, there is *by definition* a caller that will handle it, when relevant. That isn't to say that every function that raises it is only called from the API, but that when we may want to report an error via the API, we are being called from there, and so there is something that catches `APIError`.

In this case, there isn't such a top-level handler, so we would need to add them as appropriate, and the places we may need those will grow as improve the software to partially recover from errors.

If we make this change, that also means that we can leave getting the timestamp to the status API side, so we don't need to pass in a reactor here.

src/magic_folder/error.py
```diff
@@ -0,0 +1,36 @@
...
+    division,
+    print_function,
+    unicode_literals,
+)
+
+import attr
+
+
+@attr.s
+class PublicError(Exception):
```
tomprince, line 12: As discussed above, I don't think we need this exception type.

src/magic_folder/status.py
```diff
@@ -33,6 +33,14 @@ class IStatus(Interface):
     messages from the status API.
     """
 
+    def error_occurred(public_error):
+        """
+        Some error happened that should be reported to the user.
+
+        :param PublicError public_error: a plain-language description
+            of the error.
```
tomprince, line 41: I think we should take a string here, rather than an exception.

tomprince (changes requested):

src/magic_folder/downloader.py
meejah: I did have it more like that at first, but then it's like "`error_occurred(PublicError(..)); raise`" most of the time .. and was thinking with a top-level handler we can then raise `PublicError` directly more places (if the need arises -- I'm sure there will be more errors we want to report to "users").

time-stamp-wise, sure maybe that's better just handled directly by `IStatus` instead of at each raise-point.

As far as "continue scanning, or not" this code doesn't change that behavior (before, it would leak the exception out anyway). I'm not sure if there _is_ a right answer here for that .. some errors are probably "per-file" but if it's like "disc full" or "can't write the magic-dir" then you're going to flood a bunch of errors on every scan, basically?

So, what if I take the timestamp out of the `PublicError` ctor (and make the status service insert them instead) and we can file a ticket about (or otherwise address) the wider issue of "do we keep scanning on errors in a folder, or not"..?

src/magic_folder/error.py
meejah: The reason I added this is to be extra-explicit about errors that can be shown to users (and so that when writing / handling them, there's an extra reminder that you're writing for an end-user and not a developer or other expert).

src/magic_folder/downloader.py
meejah: Regarding the handling: I did add a higher-level handler (at the "downloader service" level).

tomprince: > I did have it more like that at first, but then it's like "error_occurred(PublicError(..)); raise" most of the time

That seems like a reasonable thing to have (though it isn't clear to me that the `PublicError` there adds anything). That is exactly the interface I was suggesting.

> with a top-level handler we can then raise PublicError directly more places

I'm not sure what you mean by this? I guess `.snapshots` and `.participants` don't have a status instance to call, maybe? My guess is that those places may not have enough context for a good error message, and anywhere else can report a status error directly as easily as raise `PublicError`.

> As far as "continue scanning, or not" this code doesn't change that behavior

Sure, this was thinking about how we might change the code, and why we might not want to depend on `PublicError` escaping to a particular exception handler. I wasn't suggesting changing anything, just thinking through what impacts the API we pick here would be, if we made that change.

> Regarding the handling: I did add a higher-level handler (at the "downloader service" level).

Yes. My point about possibly handling errors is that we may decide we want to handle errors before they get to there, which may lead to a proliferation of places where we need to catch `PublicError` and report it, in *addition* to all the places where we raise it.

tomprince: > So, what if I take the timestamp out of the PublicError ctor (and make the status service insert them instead)

To be clear, I think we should go with the API that does not use exceptions.

tomprince (approved):
Thanks! This looks good.

src/magic_folder/status.py
```diff
@@ -173,7 +173,7 @@ class PublicError(object):
     avoiding jargon and technical details (except where immediately
     relevant). This is used by the IStatus API.
     """
-    timestamp = attr.ib(validator=attr.validators.instance_of((float, int)))
+    timestamp = attr.ib(validator=attr.validators.instance_of((float, int, long)))
```
tomprince, line 176:
```suggestion
    timestamp = attr.ib(validator=attr.validators.instance_of((float,) + six.integer_types))
```

Issue #484, opened 2021-07-13
Intermittent test_create_then_recover[add_snapshot] failure on macOS


exarkun:
Seen, for example, at https://github.com/LeastAuthority/magic-folder/runs/3051261891?check_suite_focus=true

```
>           raise ExpectedFileMismatchException(path, timeout)
E           ExpectedFileMismatchException: Contents of '/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmp8gAaq4/kitties/sylvester' mismatched after 25s
```

meejah:
I haven't seen this in some time. Will re-open if we do.

(also the example log linked above is deleted).

PR #498, opened 2021-07-19
Fix windows integration tests.

### LINKED ISSUE #503: Fix windows integration tests.
I enabled the windows integration tests in #483, but hadn't looked at the output to notice that they were failing (which was masked by `continue-on-error: True` in the workflow).

tomprince:
Fixes #503.

- Uses the correct paths for executables in the tahoe virtualenv
- Uses stderr to capture eliot logs on windows (since we can't pass a pipe on FD 3)
- Ensures we don't hang the integration tests if a service fails to start.
- Uses psutil to suspend tahoe, since SIGSTOP doesn't exist on windows
- Don't block the reactor when sleeping in integration tests
- Ensure that log output from services is unbuffered on windows.
- Ensure that the folder state database is closed before trying to remove a folder in the integration tests.

This might also address #484.


src/magic_folder/util/eliotutil.py
```diff
@@ -350,19 +371,56 @@ def get_file():
...
 
 
-def log_call_deferred(action_type):
+def log_call_deferred(action_type, include_args=False):
     """
     Like ``eliot.log_call`` but for functions which return ``Deferred``.
     """
+
+    if include_args:
+        if include_args is True:
```
exarkun, line 380: I guess this less than wonderful bool/list signature is copied from Eliot so it makes sense to keep it instead of doing something less confusing...

integration/conftest.py
```diff
@@ -120,6 +106,28 @@ def cleanup():
     return tmp
 
 
+@attr.s
+class VirtualEnv(object):
```
exarkun, line 110: Please add a docstring here.

integration/conftest.py
```diff
@@ -155,12 +163,12 @@ def tahoe_venv(request, reactor):
 
     venv_dir = parser.get("testenv:{}".format(tahoe_env), 'envdir')
 
-    return Path(venv_dir)
+    return VirtualEnv(Path(venv_dir))
```
exarkun, line 166: nice

integration/util.py
```diff
@@ -277,6 +282,17 @@ def leave(self, folder_name):
         """
         magic-folder leave
         """
+        if self._global_config is not None:
+            # If we've accessed the folder state database from the integration
+            # tests, make sure that the connection has been closed before we
+            # try to remove the database. This is necessary on windows,
+            # otherwise the state database can't be removed.
```
exarkun, line 289: It seems like this might suggest that we actually shouldn't even keep a long-lived connection to the database?  Perhaps there are other factors to balance this consideration against ... but "connecting" to a SQLite3 database is not a particularly expensive operations and simplifying resource management by cutting down on the number of long-lived resources to manage might be worth it.

integration/util.py
```diff
@@ -562,11 +603,27 @@ def outReceived(self, data):
...
             sys.stdout.write(data)
 
+    def eliot_garbage_received(self, data):
+        """
+        Called when non-JSON lines are received on FD 3.
+
+        Since FD 3 is suppposed to only have eliot-logs, log them as malformed.
+        """
+        with self._action.context():
+            Message.log(message_type=u"malformed-eliot-log", data=data)
```
exarkun, line 625: Can we also make this into a test-failing event so that if/when such things arise in the future we can catch them quickly?  Or is there so much garbage on FD 3 that it's not even worth trying to make it clean?

integration/util.py
```diff
@@ -835,18 +901,20 @@ def await_file_contents(path, contents, timeout=15):
...
+        yield twisted_sleep(reactor, 1)
     if exists(path):
         raise ExpectedFileMismatchException(path, timeout)
     raise ExpectedFileUnfoundException(path, timeout)
 
+@inline_callbacks
 def ensure_file_not_created(path, timeout=15):
+    from twisted.internet import reactor
     start_time = time.time()
     while time.time() - start_time < timeout:
```
exarkun, line 913: It probably doesn't make any difference, at least not now, but as long as this is reactor-driven code `reactor.seconds` might make more sense than `time.time` here?

integration/util.py
```diff
@@ -924,6 +992,11 @@ def twisted_sleep(reactor, timeout):
     return deferLater(reactor, timeout, lambda: None)
 
 
+def sleep(timeout):
```
exarkun, line 995: This most definitely begs for a docstring and maybe even a more descriptive name.

integration/util.py
```diff
@@ -539,21 +569,32 @@ def __init__(self, magic_text):
...
-        self._eliot_stream = EliotLogStream()
+        self._eliot_stream = EliotLogStream(fallback=self.eliot_garbage_received)
+        self._eliot_stderr = EliotLogStream(fallback=self.err_received)
         self._action = current_action()
         assert self._action is not None
 
     def processEnded(self, reason):
         with self._action:
             Message.log(message_type=u"process-ended")
+        if not self.magic_seen.called:
```
exarkun, line 580: I generally think that application-level inspection of `Deferred.called` indicates something has gone fairly wrong.

Perhaps this particular code looks like it's okay ... it is at least the creator/caller of the Deferred that is doing the checking.  That basically means it's just a shortcut to avoid having to keep track of whether this same unit-of-responsibility has fired the Deferred or not yet.

I still think it would be better to track this information some other way.  Perhaps by dropping `magic_seen` (setting it to `None`) immediately before firing it and then checking it against `None`?

I think it's worth doing this because even though this instance of looking at `.called` is probably okay, I'd rather have a codebase where nothing ever looks at `.called` and not have to stop and think about whether certain uses are okay or not.

integration/util.py
```diff
@@ -801,7 +866,7 @@ def __init__(self, path, timeout):
 
 
 
-@log_call(action_type=u"integration:await-file-contents", include_args=["path", "contents", "timeout"])
+@log_inline_callbacks(action_type=u"integration:await-file-contents", include_args=True)
```
exarkun, line 869: This was blocking and returned None.  Now it is reactor-driven and returns a Deferred.  Probably worth a docstring change.

exarkun (approved):
Thanks.  Some simple comments inline.  Please address and then merge.

integration/util.py
tomprince: I don't think it makes sense for each test that access a folder state database to reach into `MagicFolderConfig.database` and close it (I suspect that attribute should actually be private).

It isn't so much that we want to keep long-lived connections open, but that there isn't a good API for closing them. Part of the issue is we keep a cache of `MagicFolderConfig` obejcts on `GlobalConfigDatabase` (to ensure that we only have a single connection to each state database, and that we have access to it to close it), and there is not an API for removing an object from that cache. (See https://github.com/LeastAuthority/magic-folder/pull/475/commits/f4bbe95402dbddb2b052e77ab29042d2ad4b85ca for details.)  I'm open to other suggestions on how to handle the interactions here.

integration/util.py
tomprince: It is not straight-forward to do this for a couple of reasons:
- the service is not scoped to a single test
- I don't think raising an exception here would cause the test to fail, as it will be swallowed by the reactor.
I filed #510 to address this. As it stands, I don't think we should ever hit this, as FD 3 is explicitly for eliot log messages.

src/magic_folder/util/eliotutil.py
tomprince: Not *exactly*; eliot treats `None` as include everything, instead of `True`.

PR #500, opened 2021-07-19
Stop using mangled paths, except for tahoe directory entries.

### LINKED ISSUE #499: Stop using mangled paths, except for tahoe directory entries.
The reason we use mangled paths, is because we want to use a flat directory structure for a magic-folder. The only place that cares about having a `/` in the name is the directory entry, so we should confine the mangling of names to there, rather than also keeping them mangled in the database or in the snapshot metadata.

This simplifies a lot of code, and also limits the number of places that #395 needs to be dealt with.

tomprince:
Fixes #499.

Doing this now was actually motivated by wanting to move `tahoe_client.add_entry_to_mutable_directory` for #420, but once I had that, it turned out that #499 was very easy to implement on top of that.

### DIFF
--- a/src/magic_folder/downloader.py
+++ b/src/magic_folder/downloader.py
@@ -52,9 +52,6 @@ from twisted.internet.interfaces import (
 from .config import (
     MagicFolderConfig,
 )
-from .magicpath import (
-    path2magic,
-)
 from .participants import IWriteableParticipant
 from .snapshot import (
     create_snapshot_from_capability,
@@ -690,7 +687,7 @@ class DownloaderService(service.MultiService):
             # latest, there is nothing to do .. otherwise, we
             # have to figure out what to do
             try:
-                our_snapshot_cap = self._config.get_remotesnapshot(path2magic(relpath))
+                our_snapshot_cap = self._config.get_remotesnapshot(relpath)
             except KeyError:
                 our_snapshot_cap = None
             if snapshot.capability != our_snapshot_cap:
--- a/src/magic_folder/downloader.py
+++ b/src/magic_folder/downloader.py
@@ -28,8 +28,9 @@ from eliot.twisted import (
 )
 from eliot import (
     log_call,
-    start_action,
     Message,
+    start_action,
+    write_failure,
 )
 
 from twisted.application import (
@@ -52,7 +53,7 @@ from .config import (
     MagicFolderConfig,
 )
 from .magicpath import (
-    magic2path,
+    path2magic,
 )
 from .snapshot import (
     create_snapshot_from_capability,
@@ -212,7 +213,7 @@ class IMagicFolderFilesystem(Interface):
     there are no 'partial' files in the magic-folder.
     """
 
-    def download_content_to_staging(remote_snapshot, tahoe_client):
+    def download_content_to_staging(relpath, remote_snapshot, tahoe_client):
         """
         Prepare the content by downloading it.
 
@@ -220,7 +221,7 @@ class IMagicFolderFilesystem(Interface):
             content (or errback if the download fails).
         """
 
-    def mark_overwrite(remote_snapshot, staged_content):
+    def mark_overwrite(relpath, remote_snapshot, staged_content):
         """
         This snapshot is an overwrite. Move it from the staging area over
         top of the existing file (if any) in the magic-folder.
@@ -231,7 +232,7 @@ class IMagicFolderFilesystem(Interface):
         :return PathState: The path state of the file that was written.
         """
 
-    def mark_conflict(remote_snapshot, staged_content):
+    def mark_conflict(relpath, remote_snapshot, staged_content):
         """
         This snapshot causes a conflict. The existing magic-folder file is
         untouched. The downloaded / prepared content shall be moved to
@@ -247,7 +248,7 @@ class IMagicFolderFilesystem(Interface):
             content.
         """
 
-    def mark_delete(remote_snapshot):
+    def mark_delete(relpath, remote_snapshot):
         """
         Mark this snapshot as a delete. The existing magic-folder file
         shall be deleted.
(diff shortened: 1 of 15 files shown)


src/magic_folder/downloader.py
```diff
@@ -209,15 +208,15 @@ class IMagicFolderFilesystem(Interface):
     there are no 'partial' files in the magic-folder.
     """
 
-    def download_content_to_staging(remote_snapshot, tahoe_client):
+    def download_content_to_staging(relpath, remote_snapshot, tahoe_client):
```
meejah, line 211: Removing the tie between the snapshot and its name seems like a bad move. Why can't the snapshot know its own name?

tomprince: It is generally better to have a single source of truth, and given that we *need* to have the name in the directory entry, it seems preferable to have that be the source of truth. Also, I can imagine a future where we detect renames, in which case, we may end up writing a file to a renamed path.

At this *particular* point, I think it makes sense for `IMagicFolderFilesystem` to not know about snapshots, this is an incremental step towards that. (That is, we should also be passing in the capabilitiy, mtime and author, to the three functions, rather than accessing those attributes on the passed in snapshot).

meejah: What are you saying is "the single source of truth"?

tomprince: The (demangled) name from the DMD that we discovered this snapshot via.

tomprince: Part of the reason that I prefer using that name, is that there can't be collisions by construction[1], but the data format allows for clashes in the metadata name (even if our current code never constructs such a thing).

[1] Unless or mangling isn't unique. Or, if tahoe returned a JSON object with duplicate keys, but that would be a violation of *tahoe*s data model

meejah: So why can't the snapshot know its own name?

tomprince: Do you mean the python `RemoteSnapshot` object?  I don't have any objection major[1] objection to it knowing the name it was access by, though it would require some refactoring to that, since we don't currently have the name where we are constructing the snapshot . Also, for snapshots that we discover as parents, the tie to the name of the child is less strong (though in this case, I don't think we ever *use* the name).

On the other hand, I don't see any clear benefit to having it store the name either. One downside is that a `RemoteSnapshot` would then no longer be a python representation of snapshot capability (this is why it would no longer be a great representation of parent snapshots, as mentioned above).

[1] Though it might be confusing to have both a name, and a potentially different name in the metadata.

tomprince: >  I don't have any objection major objection [...]

To be clear, I don't think we should make `RemoteSnapshot.name` be the name from the DMD. I think it would make the code (in particular `RemoteSnapshotCacheService`) more complicated, and the make the correspondence between the persistent (i.e. on grid) data representation and the in-process one more tenuous. The only advantage I can see is not needing to pass an extra value a few places, which isn't clearly better.

meejah: `RemoteSnapshot.name` _is_ the name from the DMD.

tomprince: > `RemoteSnapshot.name` _is_ the name from the DMD.

No: https://github.com/LeastAuthority/magic-folder/blob/6f8c379e253effd0c5144f7f4a1170acd93fa34c/src/magic_folder/snapshot.py#L428-L443 It is the name from the metadata blob in the snapshot. Given that `create_snapshot_from_capability` only takes the capability, it *can't* have the name from the DMD that the points at that capability.

(Granted, it is currently always the same, but as I argued above, we should get the value from the source of truth, which is not where it comes `RemoteSnapshot.name` comes from.)

meejah: > Granted, it is currently always the same

Yes, that's what I mean. "A snapshot's name" is its entry in the DMD.

tomprince: My point is we should program defensively against those getting out of sync, either due to a bug, or because in the future we (e.g.) detect renames. So, **even though they are currently the same**, we should consistently use only one of the values, rather than sometimes using one and sometimes using the other.

And, before you point out we could bump the version if we want to decouple them, if we program defensively now, we may not need to do that.

meejah: I'm talking about all the places that take an extra "relpath" arg in this PR beside (or sometimes instead of) a `RemoteSnapshot` .. that makes it easy to pass the wrong thing instead of the snapshot knowing its name (and/or relpath).

tomprince: > I'm talking about all the places that take an extra "relpath" arg in this PR [...] that makes it easy to pass the wrong thing

I do think that is a reasonable concern. As it is, we don't have an object that bundles a remote snapshot and the name in the DMD that we read it from (if we are programming defensively, `RemoteSnapshot` *is not that*). I don't think it worth it to introduce one here, particularly as I want most of the places this would affect to stop taking snapshots at all.

For me, it comes down to being resilient against inconsistencies or changes in format of our persistent data (i.e. what we read from the tahoe grid) outweights the risk of passing the wrong arguments. The later is *entirely* under our control and doesn't impose any constraints on what we do in the future. The way we interpret data on-grid *does* constrain how we can evolve our software in the future, so I am much more concerned about that.

meejah (changes requested):
I'm not talking about "the source of truth". Doesn't matter what you pass when constructing a `RemoteSnapshot` (right now it's "the name" but you could of course pass "the relpath" instead or even just de-mangle the thing that's passed currently in the ctor). I'm saying adding a new arg to a bunch of methods doesn't help anything.

Bigger-picture here, for the entire existence of this project it has used the terminology that "the name" of a Snapshot is a mangled/flattened relative-path. Since you seem very sure that this must change, it needs to be more cohesive than adding a new "relpath" arg to a bunch of methods -- doing that just makes things _more_ confused since "name" is still all over the place (and you've also re-used that to mean "relpath" in some places). There's also documentation etc. (including the signature scheme) to fix up if we really to do _need_ to change all this immediately. (Also something besides "name" must be used to avoid confusion .. hence suggesting "`relpath`").

I don't see how any of this is vital to a first internal release though, so it's low-priority for me at the moment.

PR #502, opened 2021-07-20
501.last modified

### LINKED ISSUE #501: Store a last-modified timestamp
The "file-status" status endpoint should properly reflect the time when changes were made (which is likely different from e.g. the modified-time of a Snapshot itself).

magic-folder shall store a "last modified" timestamp in its state database reflecting when an update to that row was made (and include that information in the `/v1/magic-folder/<folder>/file-status` API)

meejah:
fixes #501 

add "last-updated" data to `file-status` endpoint

### DIFF
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -735,6 +735,7 @@ class MagicFolderConfig(object):
     """
     name = attr.ib()
     database = attr.ib()  # sqlite3 Connection
+    _get_current_timestamp = attr.ib(default=time.time)
 
     @classmethod
     def initialize(
@@ -1004,7 +1005,7 @@ class MagicFolderConfig(object):
         action = STORE_OR_UPDATE_SNAPSHOTS(
             relpath=name,
         )
-        now_ns = seconds_to_ns(time.time())
+        now_ns = seconds_to_ns(self._get_current_timestamp())
         with action:
             cursor.execute(
                 """
@@ -1037,7 +1038,7 @@ class MagicFolderConfig(object):
         action = STORE_OR_UPDATE_SNAPSHOTS(
             relpath=name,
         )
-        now_ns = seconds_to_ns(time.time())
+        now_ns = seconds_to_ns(self._get_current_timestamp())
         with action:
             try:
                 cursor.execute(
@@ -1066,7 +1067,7 @@ class MagicFolderConfig(object):
         action = STORE_OR_UPDATE_SNAPSHOTS(
             relpath=name,
         )
-        now_ns = seconds_to_ns(time.time())
+        now_ns = seconds_to_ns(self._get_current_timestamp())
         with action:
             try:
                 cursor.execute(
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -27,6 +27,7 @@ from base64 import (
     urlsafe_b64encode,
 )
 import hashlib
+import time
 
 from hyperlink import (
     DecodedURL,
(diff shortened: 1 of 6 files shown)


src/magic_folder/config.py
```diff
@@ -214,16 +216,17 @@
...
-            [ctime_ns]      INTEGER NOT NULL, -- mtime of current snapshot
-            [size]          INTEGER NOT NULL  -- size of current snapshot
+            [name]             TEXT PRIMARY KEY, -- mangled name in UTF-8
+            [snapshot_cap]     TEXT,             -- Tahoe-LAFS URI that represents the most recent remote snapshot
+                                                 -- associated with this file, either as downloaded from a peer
+                                                 -- or uploaded from local changes
+            [mtime_ns]         INTEGER NOT NULL, -- ctime of current snapshot
+            [ctime_ns]         INTEGER NOT NULL, -- mtime of current snapshot
+            [size]             INTEGER NOT NULL, -- size of current snapshot
+            [last_updated_ns]  INTEGER NOT NULL  -- timestamp when last changed
```
tomprince, line 229: I'm slightly tempted to say this should not be in ns, given that we don't care about that level of accuracy for it, but consistency is probably more valuable.

src/magic_folder/config.py
```diff
@@ -1161,24 +1174,25 @@ def get_all_current_snapshot_pathstates(self, cursor):
...
-                mtime_ns DESC
+                last_updated_ns DESC
             """,
         )
 
         return [
             (
                 row[0],  # name
                 PathState(mtime_ns=row[1], ctime_ns=row[2], size=row[3]),
+                ns_to_seconds(row[4]),  # last_updated_ns
```
tomprince, line 1195: Although the fact that we don't expose anything other than seconds outside the database leans me towards storing seconds in the db.

src/magic_folder/config.py
```diff
@@ -1001,10 +1004,18 @@ def store_uploaded_snapshot(self, cursor, name, remote_snapshot):
         action = STORE_OR_UPDATE_SNAPSHOTS(
             relpath=name,
         )
+        now_ns = seconds_to_ns(time.time())
```
tomprince, line 1007: I think we should wrap this in a function (at least so we can patch in tests).

tomprince (approved):
This looks good. I do vaguely wonder if we may actually want more than one timestamp, but in the absence of better requirements, this seems find to land now.

src/magic_folder/config.py
meejah: I don't think it's a good idea for _any_ of these to be nanoseconds (for the same reason) but yes, consistency is better.

tomprince: There is good reason for mtime/ctime to be nanoseconds, although when I added those values, I never meant them to be exposed to users.  I wonder if it would make sense to track the user-visible modified time separately in seconds (along with last-updated time)? The `PathState` tuple was really intended to be a mostly opaque snapshot of (part of) the file's stat-info, that we can use to notice when the file has changed; for that use case, we want as much accuracy as possible.

meejah: Users do not see nanoseconds.

tomprince: I ... know. My suggestion was to decouple the user-visible data we track, from the internal state tracking. It isn't obvious that they always want to be in sync. For example, if somebody `touch`s a file[1], we must update the internal state tracking, but we might want to not update the user-visible modified time, given that the file wasn't actually modified. Just because the two values happen to be equal now, doesn't mean we should store them in the same column, if they have different meanings.

[1] To be extra clear, we *don't* do this now, and we may not want to do this in the future. But I think it is entirely reasonable to think about potential future cases, when thinking about how to model data.

tomprince: And, to be clear, neither of the points are blocking landing this.

PR #514, opened 2021-07-28
Don't pass snapshot to IMagicFileSystem.

tomprince:
@meejah This is the followup to #500 that we discussed.

- Don't pass snapshot to IMagicFileSystem.
- Remove .name from RemoteSnapshot.

Once #500 lands, I'll re-target this to main, but for the moment, it is against the current (as of this PR being opened) state of #500.

### DIFF
--- a/docs/downloader.rst
+++ b/docs/downloader.rst
@@ -28,7 +28,7 @@ It is represented by an immutable directory and contains:
   Deletion Snapshots contain no such key.
 - ``metadata``: information about the Snapshot, a read-only Capability pointing to a JSON-serialized dict containing:
   - ``snapshot_version``: int(1) currently
-  - ``name``: the name of this snapshot (a mangled relative path)
+  - ``name``: the name of this snapshot (a relative path)
   - ``author``: a dict containing:
     - ``name``: arbitrary name
     - ``verify_key``: base64-encoded public key of the author
--- a/src/magic_folder/api_cli.py
+++ b/src/magic_folder/api_cli.py
@@ -100,8 +100,8 @@ def dump_state(options):
     print("  magic_path: {}".format(config.magic_path.path), file=options.stdout)
     print("  collective: {}".format(config.collective_dircap), file=options.stdout)
     print("  local snapshots:", file=options.stdout)
-    for snap_name in config.get_all_localsnapshot_paths():
-        snap = config.get_local_snapshot(snap_name)
+    for relpath in config.get_all_localsnapshot_paths():
+        snap = config.get_local_snapshot(relpath)
         parents = ""
         q = deque()
         q.append(snap)
@@ -109,15 +109,15 @@ def dump_state(options):
             s = q.popleft()
             parents += "{} -> ".format(s.identifier)
             q.extend(s.parents_local)
-        print("    {}: {}".format(snap_name, parents[:-4]), file=options.stdout)
+        print("    {}: {}".format(relpath, parents[:-4]), file=options.stdout)
     print("  remote snapshots:", file=options.stdout)
-    for snap_name, ps, last_update, upload_duration in config.get_all_current_snapshot_pathstates():
+    for relpath, ps, last_update, upload_duration in config.get_all_current_snapshot_pathstates():
         try:
-            cap = config.get_remotesnapshot(snap_name)
+            cap = config.get_remotesnapshot(relpath)
         except KeyError:
             cap = None
             continue
-        print("    {}:".format(snap_name), file=options.stdout)
+        print("    {}:".format(relpath), file=options.stdout)
         print("        cap: {}".format(cap), file=options.stdout)
         print("        mtime: {}".format(ps.mtime_ns), file=options.stdout)
         print("        size: {}".format(ps.size), file=options.stdout)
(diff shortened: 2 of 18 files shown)


meejah (changes requested):
This (along with 500) definitely gets us _closer_ to solving the issues with this refactoring that we chatted about.

However, the problem that "name" now refers to relpath sometimes and mangled-name some other times is still worsened by these PRs (e.g. the database column is still "name" but got re-used as relpath, and some other inline comments to this effect).

To re-iterate from the conversation you mention: whether things internally use mangled-name or relpath is rather arbitrary -- but they _absolutely should be consistent_ and so I don't want to get stuck in a half-way position where there's both (since it's already confusing enough). Up until now, the arbitrary decision is that "name means mangled-name" and so we can't just start using "name means relpath, but only here and over there" .. we should use "relpath" to mean relative path and "name" (or "manged_name") to mean the other thing. Also from that conversation is that this may have implications on the signature spec as well (although also maybe not); has that been analyzed?

tomprince:
> However, the problem that "name" now refers to relpath sometimes and mangled-name some other times is still worsened by these PRs (e.g. the database column is still "name" but got re-used as relpath, and some other inline comments to this effect).

I've not looked at any magic-folder code for close to a week, but where do we ever (after these two PRs) use `name` to refer to the mangled name? As far as I can tell, the only place the mangled form is given a name after this is `encoded_relpath_u` in `_CollectiveDirnodeParticipant.files`. I'll admit that relpath and name are used interchangably, which can be confusing but not ambiguous.

tomprince:
> However, the problem that "name" now refers to relpath sometimes and mangled-name some other times is still worsened by these PRs (e.g. the database column is still "name" but got re-used as relpath, and some other inline comments to this effect).

As I mentioned above, `name` is never used for the mangled name. I have gone and fixed up the comments that suggest it is a mangled name (which I'll push shortly). I'd be happy to go through and change all references from name to relpath, but I'd prefer to do that in a followup PR.

> To re-iterate from the conversation you mention: whether things internally use mangled-name or relpath is rather arbitrary 

Yes and no. Certainly we can use whatever for internally we want, we could even use something else like the base64 encoded version or a rot13[1] encoded version. However, given that relpath isn't something arbitrary (unlike the mangling we've chosen), picking that form as the one we use internally isn't entirely arbitrary.

> but they absolutely should be consistent and so I don't want to get stuck in a half-way position where there's both

I don't think this leaves us in a half-way position (nor does #500 on its own), except for a few out-of-date comments, which I'll fix.

> Up until now, the arbitrary decision is that "name means mangled-name"

I do keep finding indications in old documentation that that *wasn't* always the case. For example https://github.com/LeastAuthority/magic-folder/blob/main/docs/snapshots.rst#signature-details talks (implicitly, by mentioning a mangled version of it) about the name being the unmangled version.

> we should use "relpath" to mean relative path and "name" (or "manged_name") to mean the other thing.

After this, we use `relpath` and `name` interchangeably to refer to the relative path and don't refer to the mangled name at all.

> Also from that conversation is that this may have implications on the signature spec as well (although also maybe not); has that been analyzed?

Reading https://github.com/LeastAuthority/magic-folder/blob/main/docs/snapshots.rst#signature-details suggests that the signature spec envisioned using the unmangled name already; it seems these two PRs bring the code (closer) in-line with the signature spec[2].

---

I'd be happy to make the changes you suggested about `s/name/relpath` if that gets #500 and this landed, though as I mentioned it isn't ambiguous.


[1] though what that means for full utf-8 name could be debated :wink: 
[2] though the spec does seem want us to use the name from the DMD to verify, rather than the embedded name which is what we currently do. Our code isn't currently structured that we have access to the DMD name at the point we verify the signature.

meejah:
> After this, we use relpath and name interchangeably to refer to the relative path and don't refer to the mangled name at all.

yes, this is the problem. I think we should be clear and use `relpath` for everything that's a relative path (including the database columns, arg names, etc). "name" was never a great label anyway, and makes even less sense if it's actually a relative-path and not an arbitrary label (as the generic "name" at least implies).

tomprince:
> > After this, we use relpath and name interchangeably to refer to the relative path and don't refer to the mangled name at all.
> 
> yes, this is the problem. I think we should be clear and use `relpath` for everything that's a relative path (including the database columns, arg names, etc). "name" was never a great label anyway, and makes even less sense if it's actually a relative-path and not an arbitrary label (as the generic "name" at least implies).

I agree we should change this. Is doing it now a blocker? (I'm partly reluctant to do it here, as I suspect it might lead to a lot of code conflicts).

meejah:
> I agree we should change this. Is doing it now a blocker? (I'm partly reluctant to do it here, as I suspect it might lead to a lot of code conflicts).

Yes, because re-using "name" to mean "relpath" (sometimes) is confusing.

I don't see that any of this re-factoring is necessary for the real feature .. but as I said, it's hard to tell because this is _just_ the refactoring, and not the real feature.

tomprince:
> I don't see that any of this re-factoring is necessary for the real feature .. but as I said, it's hard to tell because this is _just_ the refactoring, and not the real feature.

Generally no refactoring is necessary. I do tend to split work up into "making the change easy" and then "making the easy change", even if it ends up being more work than just making the change. However, my experience is that the extra work starts paying itself off fairly quickly. I also enjoy refactoring and code gardening, so am probably more aggressive about refactoring than might be strictly necessary; but it part of how I understand code and the changes I need to make and manage complexity.

tomprince:
I've gone and changed everything (except the db and the snapshot metadata) to use relpath.

PR #518, opened 2021-07-29
prototype-endpoint-reporting

### LINKED ISSUE #516: `magic-folder-api` fails with "Error: invalid port 0" when global configuration uses tcp endpoint with port number0
When `magic-folder-api` is pointed at a configuration that is configured with a TCP endpoint address using port 0 (for OS-managed automatic free port assignment) it fails to operate, reporting only:

```
Error: invalid port 0
```

In this case a running magic-folder process will be listening on *some* actual TCP port number of `magic-folder-api` is not figuring out what this is.

The expected behavior is that wherever the running magic-folder process is listening, `magic-folder-api` can find it to invoke APIs.

tomprince:
Fixes #516.

### DIFF
--- a/src/magic_folder/cli.py
+++ b/src/magic_folder/cli.py
@@ -560,6 +560,9 @@ class BaseOptions(usage.Options):
     def client(self):
         if self._client is None:
             from twisted.internet import reactor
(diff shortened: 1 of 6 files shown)


meejah:
superceded

PR #520, opened 2021-07-29
Provide a wrapper for IStatus that passes the folder name automatically.

### LINKED ISSUE #519: Provide a wrapper for `IStatus` that passes the folder name automatically.
All the methods of the status service take a folder-name. Since each component service of `MagicFolder` only ever reports errors for the corresponding folder, it would be convenient if they did not need to access and pass the folder name whenever they call the status service.

tomprince:
Fixes #519.

### DIFF
--- a/src/magic_folder/status.py
+++ b/src/magic_folder/status.py
@@ -5,7 +5,10 @@ from __future__ import (
     unicode_literals,
 )
 
+from functools import partial
 import json
+
+from six import string_types
 import attr
 from collections import (
     defaultdict,
@@ -15,6 +18,7 @@ from zope.interface import (
     Interface,
     implementer,
 )
+from zope.interface.interface import Method
 
 from autobahn.twisted.websocket import (
     WebSocketServerFactory,
@@ -38,7 +42,7 @@ class IStatus(Interface):
     messages from the status API.
     """
 
-    def error_occurred(public_error):
+    def error_occurred(folder, message):
         """
         Some error happened that should be reported to the user.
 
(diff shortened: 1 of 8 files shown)


src/magic_folder/status.py
```diff
@@ -38,7 +42,7 @@ class IStatus(Interface):
     messages from the status API.
     """
 
-    def error_occurred(public_error):
+    def error_occurred(folder, message):
```
meejah, line 45: Hmm .. although I suppose _most_ errors will be "for a folder" I imagine there are cases of more generic errors we'll want to report. Forcing this API to have a "folder" arg just to do the wrapper-thing "more generically" feels pretty forced.  (Or is it just the interface here that's already-wrong since the args seem to refer to the "exception" version).

Anyway, "burning in" that all errors only occur in the context of a single magic-folder doesn't feel correct.

src/magic_folder/status.py
```diff
@@ -368,3 +372,57 @@ def download_finished(self, folder, relpath):
...
+        for name, method in iface.namesAndDescriptions():
+            if not isinstance(method, Method):
+                continue
+            if method.positional[0] == relative:
+                setattr(cls, name, _ProxyDescriptor(original, relative, name))
+        return cls
+    return decorator
+
+
+@relativeProxyFor(IStatus, "_status", "folder")
```
meejah, line 420: I see where you're going with this .. but now we're tying _all_ (current and future) "status" APIs to require a "folder" as the first argument.
It also makes this code somewhat hard to read? (e.g. instead of having an `IStatus` interface to look at, you have to trace back here and figure out that it's an `IStatus` under the hood with magic first arg before then finding `IStatus` to see what it actually does).

I guess overall I'm "-0" on making this so generic for a single use-case (i.e. just to avoid accessing the right folder-name for the half-dozen call-sites that require those).

src/magic_folder/status.py
tomprince: This is just fixing up the interface to match the implementation.

src/magic_folder/status.py
tomprince: > but now we're tying all (current and future) "status" APIs to require a "folder" as the first argument.

I very deliberately designed this to *not* constrain us that way. This *only* proxies method that take `folder` as the first argument, and ignores methods that don't have that as a first argument (and non-methods). While that does limit what is available via `FolderStatus`, `IStatus` can have other methods that are accessible directly.

The current implementation keeps the underlying `IStatus` private, so we can't access those methods from a `FolderStatus`, but my feeling is we shouldn't need to (and we can always revisit that).

My immediate motivation for this was wanting to pass a `FolderStatus` to a function that doesn't currently have access to the name, without also needing to pass the name in (`magic_folder.scanner.find_updated_files`). I also noticed when the status code initially landed that you changed at least one object around that *didn't* have the name, just to be able to report status. I considered suggesting something like this (at least the interface, not necessarily the implementation).

src/magic_folder/status.py
tomprince: As an aside, I agree that we should be able to report errors that aren't tied to a folder; I think when I was writing this, one of the reasons I had in my head for filtering methods to proxy was being able to report global errors.

src/magic_folder/status.py
meejah: >  I also noticed when the status code initially landed that you changed at least one object around that didn't have the name, just to be able to report status.

I feel like "approximately everything" is fair game to have access to "the config" so don't see that as a problem.

meejah: > This only proxies method that take folder as the first argument, and ignores methods that don't have that as a first argument (and non-methods). 

Hmm, okay I see that now ..  that's good.

src/magic_folder/status.py
```diff
@@ -368,3 +372,57 @@ def download_finished(self, folder, relpath):
...
+    def __get__(self, oself, type=None):
+        if oself is None:
+            raise NotImplementedError()
+        original = getattr(oself, self.original)
+        return partial(
+            getattr(original, self.method),
+            getattr(oself, self.relative),
+        )
+
+def relativeProxyFor(iface, original, relative):
```
meejah, line 395: "Our" code is all snake_case right? This probably should be as well.

meejah (approved):
At least one remaining inline thing (snake-case), otherwise looks fine.

src/magic_folder/status.py
tomprince: Yeah, I was looking at `proxyForInterface` when I was writing this, and copied the casing there. I've updated it.

meejah: So this right here is why I find force-pushing to existing PRs hard: I don't see the change here, I have to look at the entire diff again...

PR #521, opened 2021-07-29
Add helper for generating small file content that does not fit a tahoe lit uril.

tomprince:
This makes the integration tests logs easier to read, as we don't have thousand
line long log messages.

### DIFF
--- a/integration/test_synchronize.py
+++ b/integration/test_synchronize.py
@@ -25,6 +25,18 @@ from .util import (
     twisted_sleep,
 )
 
+def non_lit_content(s):
+    # type: (unicode) -> unicode
+    """
+    Pad the given string so it is long enough to not fit in a tahoe literal
+    URI.
+    """
+    # The max size of data that will be stored in a literal tahoe cap is 55.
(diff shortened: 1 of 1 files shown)


integration/test_synchronize.py
```diff
@@ -25,6 +25,14 @@
     twisted_sleep,
 )
 
+def non_lit_content(s):
+    # type: (unicode) -> unicode
+    """
+    Pad the given string so it is long enough to not fit in a tahoe literal
+    URI.
+    """
+    return "{} {}\n".format(s, "." * max(55 - len(s), 0))
```
meejah, line 34: I'd greatly prefer to not have a magic-number here .. but of course I'd _also_ like to not have to import `allmydata.*` code .. can you at least document where this comes from so that later authors can confirm its still valid.

meejah (approved):
linter is sad, and one inline fix then looks fine

integration/test_synchronize.py
tomprince: Given that the constant we'd need to import is `allmydata.immutable.upload.Uploader.URI_LIT_SIZE_THRESHOLD`, I definitely think we shouldn't import it. I've left a comment (and included the name of the constant).

Issue #517, opened 2021-07-28
re-populate "status" object at startup


meejah:
When magic-folders first starts up it should ensure that the "status" object is in an appropriate state.

When a `LocalSnapshot` is created, for example, it is considered "queued". This will not be reflected on startup if a `LocalSnapshot` exists (i.e. that snapshot was not uploaded before magic-folders stopped).

Issue #523, opened 2021-07-29
Provide "uploaded time" in status


meejah:
To communicate when a file was last uploaded, a timestamp should be included when the last upload finished.

meejah:
Actually, according to code (and subsequent manual tests) the `last-updated` time already reflects the uploaded time. The last-updated time is changed whenever any change is recorded (so "queued a local snapshot" or "uploaded to Personal DMD" will both change the "last updated" timestamp).

So you can say if a file is up-to-date and when it was last changed: if the relative-path is in `uploads` (or `downloads`) then there are changes pending. The most-recent upload (or download) time is reflected in `last-updated` timestamp.

A locally-changed (or created) file will go through states:

- in "uploads" list, with `queued-at` time (only)
- in "uploads" list, with `queued-at` and `started-at` timestamps
- gone from "uploads" list, present in "recent" list (with `last-updated` time being shortly after the Personal DMD entry was updated)

tomprince:
I will note that we record that timestamp *before* we link the snapshot cap into the personal DMD, which effectively means that it doesn't mean the file is persistent on-grid. I guess the file will still be reported as an upload in the status in that case[1], so if it isn't in the uploads list, the time will be (hopefully only) shortly before it is linked to the DMD. That said, I think if we want to report the time when we linked the cap into the DMD, we should record that time explicitly (particularly if we end up doing something like https://github.com/LeastAuthority/magic-folder/pull/471).


(Ignoring what happens when we restart)

meejah:
If we're being pedantic: the time is currently noted _after_ the upload takes place. Thus, it is "on the grid".

It is true that it is _not_ linked in the DMD at this point. We could certainly easily update the timestamp again at that point if its interesting -- but UIs have no way to distinguish which timestamp they're seeing anyway.

Issue #515, opened 2021-07-28
a non-regular file inside a filesystem directory magic-folder is pointed at prevents contents from being uploaded


exarkun:
If magic-folder is added pointed at a directory which itself contains something that is not a "regular" file - eg a directory or a symlink - then the contents of the directory are not uploaded.  The following appears repeatedly in the log:

```
af63034d-2bce-43e6-8bed-07748160ebbd
└── scanner:find-updates/1 ⇒ started 2021-07-28 19:13:08Z ⧖ 0.002s
    └── scanner:find-updates/2 ⇒ failed 2021-07-28 19:13:08Z
        ├── exception: exceptions.UnboundLocalError
        └── reason: local variable 'path_state' referenced before assignment

05a802a8-75d0-48d4-b9e2-c3d96c369d0a
└── eliot:traceback/1 2021-07-28 19:13:08Z
    ├── exception: exceptions.UnboundLocalError
    ├── reason: local variable 'path_state' referenced before assignment
    └── traceback: Traceback: <type 'exceptions.UnboundLocalError'>: local variable 'path_state' referenced before assignment⏎
        /nix/store/j0q2p45gpj4w8p6zaixa3574a8c4lip4-python-2.7.17-env/lib/python2.7/site-packages/twisted/internet/defer.py:501:errback⏎                                                                                                                                      
        /nix/store/j0q2p45gpj4w8p6zaixa3574a8c4lip4-python-2.7.17-env/lib/python2.7/site-packages/twisted/internet/defer.py:568:_startRunCallbacks⏎                                                                                                                           
        /nix/store/j0q2p45gpj4w8p6zaixa3574a8c4lip4-python-2.7.17-env/lib/python2.7/site-packages/twisted/internet/defer.py:654:_runCallbacks⏎                                                                                                                                
        /nix/store/j0q2p45gpj4w8p6zaixa3574a8c4lip4-python-2.7.17-env/lib/python2.7/site-packages/twisted/internet/defer.py:1475:gotResult⏎                                                                                                                                   
        --- <exception caught here> ---⏎
        /nix/store/j0q2p45gpj4w8p6zaixa3574a8c4lip4-python-2.7.17-env/lib/python2.7/site-packages/twisted/internet/defer.py:1416:_inlineCallbacks⏎                                                                                                                            
[...]

PR #525, opened 2021-07-30
Status updates

meejah:
While looking into #523 I made some (different) improvements to the status / etc API:

- use `queued-at` instead of `queued_at` (for example)
- get rid of the single-entry `dict`s in "recent" status
- add an "upload duration"
- properly represent "uploads" queue after a restart where multiple LocalSnapshots already exist
- better `magic-folder-api dump-state` output/indenting, including the upload-duration + speed.

Logically separate PRs, probably, but .. ideally @crwood has feedback (and I can remove anything we don't already like).


crwood (commented):
These all look like useful, positive changes to me (especially preserving the queue between restarts) -- thank you!

The single-entry dicts in `recent` thing was mildly puzzling to me. Perhaps the same treatment should be given to the `uploads` and `downloads` entries too? Unlike `recent` and `errors` -- which point at lists -- `uploads` and `downloads` point at dicts -- keyed by relpath -- which thereby would seem to prevent displaying two upload operations for the same file. Maybe this is intentional (I don't know) but I'm not sure that/why a consumer of this API would would ever need to do a lookup by _relpath_ specifically here (e.g., in the form `uploads.get("SomeFile.txt")` or whatever). Perhaps, for conistency's sake, each of these sections -- `uploads`, `downloads`, `recent`, and `errors` -- should point to lists of dicts (ordered by `last-updated` time in the case of `recent` but `queued-at` in the case of `uploads`/`downloads`)?

meejah:
This fixes https://github.com/LeastAuthority/magic-folder/issues/517

meejah:
okay, I made those changes @crwood

meejah:
(That is, uploads and downloads are now also lists-of-dicts and are reverse-sorted on queued-at time).

Note that there will only ever be 0 or 1 entries for a given relpath (if there are multiple local snapshots for a file it's still just "a pending operation on `x`" and the queued-at time will reflect the oldest local snapshot).

tomprince:
This is perhaps out-of-scope for this PR, but I'm going to bring it up here, since this about changing the shape of some of the status information we report.

While I was working on #520 and the fix for #515, I ran into the fact the status service now has a hard-dependency on the magic-folder state database. I had originally (https://github.com/LeastAuthority/magic-folder/pull/340#discussion_r640847572) said we didn't need a fake status implementation, since it was already entirely in-memory anyway, but that is no longer true.

I was wondering if the recent changes data belongs as part of the status service? Removing that would remove the need for the status service to touch the database. Given that we get the data from the database, it doesn't seem to really make sense to be part of a push API, since we don't send a message when it changes, only when some other operation triggers an unrelated update; in particular when a client connects, we will send stale data[1]. This is perhaps slightly mitigated but the fact that we probably do send in message in those cases; that doesn't diminish the fact that there seems to be a mismatch between the the nature of the data and the push API.

[1] Unless this is the first client to connect, and nothing has happened that *would* have triggered a message.

src/magic_folder/uploader.py
```diff
@@ -410,9 +426,12 @@ def startService(self):
         Start UploaderService and initiate a periodic task
         to poll for LocalSnapshots in the database.
         """
-
         service.Service.startService(self)
 
+        # if we started with any local snapshots already in the
+        # database we must reflect this in our status service.
+        self._remote_snapshot_creator.initialize_upload_status()
```
tomprince, line 433: We should really wait to send status to any client, until we've updated the status with the appropriate information. I guess right now, we don't yield to the reactor between starting the web service and this call, but I think it would be better to explicitly ensure this is the case. (Particularly since I'm intending to make starting the magic folders asynchronous as part of #420, so we can check the on-grid version first).

My initial thought is that the `MagicFolder.ready` deferred should wait on each component service being ready (which would include populating status), and then `MagicFolderService` would notify the status service once all those have fired. That said, the code here *does* shorten the window where we could present stale data, so this can be handled in a followup (though I wouldn't consider #517 fixed until that is done).

meejah:
> Removing that would remove the need for the status service to touch the database. Given that we get the data from the database, it doesn't seem to really make sense to be part of a push API

I don't think there's any reason to _avoid_ using the database in the `/status` API.

The point of the `/status` API is to avoid UIs having to poll. If we moved the "recent" list to a polling API then GridSync would either have to poll it constantly, or _at least_ whenever a new message came in on the status API. If the status API _isn't_ reflecting a current state somehow, that's a bug -- but I don't think we should base that decision on "does (some of) the information reside in a database or not".

Now, maybe there's a use-case to have an _additional_ way to access that -- which is kind of what `.../file-status` is I suppose -- but so long as its useful for at least gridsync it should stay.

In any case, definitely out-of-scope for this PR.

crwood (approved):
I'm not sure what's going on with the Windows CI test failures but, otherwise, this looks good to me.

src/magic_folder/api_cli.py
```diff
@@ -106,14 +108,20 @@ def dump_state(options):
...
-        path_state = config.get_currentsnapshot_pathstate(snap_name)
-        print("    {}: {}".format(snap_name, cap), file=options.stdout)
-        print("    {}  {}".format(" "*len(snap_name), path_state), file=options.stdout)
+            continue
+        print("    {}:".format(snap_name), file=options.stdout)
+        print("        cap: {}".format(cap), file=options.stdout)
+        print("        mtime: {}".format(ps.mtime_ns), file=options.stdout)
+        print("        size: {}".format(ps.size), file=options.stdout)
+        if upload_duration:
+            duration = upload_duration / 1000000000.0  # humans care about seconds not ns
```
tomprince, line 122: We have a function for this.

src/magic_folder/config.py
```diff
@@ -988,7 +989,7 @@ def delete_localsnapshot(self, cursor, name):
 
 
     @with_cursor
-    def store_uploaded_snapshot(self, cursor, name, remote_snapshot):
+    def store_uploaded_snapshot(self, cursor, name, remote_snapshot, upload_started_at):
```
tomprince, line 992: We also need to adjust `store_downloaded_snapshot` as well (at least to null it out). I really think, though, that this should be in a table tied to the remote snapshot, *not* in the current snapshot table.

src/magic_folder/api_cli.py
```diff
@@ -9,6 +9,8 @@
 import json
 from collections import deque
 
+import humanize
```
tomprince, line 12: If we are going to use this directly, we should explicitly depend on it.

src/magic_folder/api_cli.py
meejah: I didn't use it here because it truncates to "seconds" .. and I figured this is one case we probably _do_ want fractional seconds so that speed calculations (esp. for smaller files) are more accurate. Maybe _all_ seconds want fractional parts, though, and `ns_to_seconds` should just return float?

tomprince: Hmm. I think all the places we currently use it, second accuracy is reasonable to have; which inclines me to not change the current interface. Can you leave a comment here that explains why you aren't using the helper?

src/magic_folder/config.py
meejah: It should not be nulled out; the last upload happened still. We're just remembering how long that took.

Perhaps we also want to remember how long the last download took, too (since it's highly likely to be different)...

tomprince: It will not be the upload duration of the current remote-snapshot, which is what the table it is currently in is tracking. It is incorrect to keep the upload duration *in this table* if we change the remote snapshot in the table.

That is why I think we should move it to a different table, because then we can keep the data tied to the right snapshot.

src/magic_folder/api_cli.py
meejah: is there any documentation for `requirements/*`?

tomprince: No, sorry. I guess https://pip-compile-multi.readthedocs.io/en/latest/boilerplate.html is probably close, although the command to run is `tox -e pin-requirements`. You can also add `-- --no-upgrade ` to that, to only resolve new dependencies, not update any other dependencies.

src/magic_folder/config.py
meejah: What table?

tomprince: It would need to be a new table, since we don't have a table that record remote snapshots, other than the pointer to the current one.

meejah: Okay I nulled it out. I think making an entire snapshot-caching system is beyond the point of the immediate ask of #523

src/magic_folder/api_cli.py
meejah: okay i added instructions and a req.

src/magic_folder/api_cli.py
meejah: made a new helper with a docstring

meejah:
Wait, now nix build fails??

(All I did here was add `humanize` as a direct dependency; it was already a dependency)

exarkun:
> Wait, now nix build fails??
> 
> (All I did here was add `humanize` as a direct dependency; it was already a dependency)
> 
> @exarkun any hints?

nixpkgs appears to have humanize-0.5.1 but the dependency is for humanize>=1.0.0.

Can magic-folder get away with 0.5.1?  If not then we can look into getting a newer version packaged (which could range from "trivial" if we just need to bump the version and point at the new tarball ... to impossible if the newer version depends on something else that depends on something else that depends on something else that ...).

exarkun:
> > Wait, now nix build fails??
> > (All I did here was add `humanize` as a direct dependency; it was already a dependency)
> > @exarkun any hints?
> 
> nixpkgs appears to have humanize-0.5.1 but the dependency is for humanize>=1.0.0.
> 
> Can magic-folder get away with 0.5.1?

I guess it *can* because that is what it was using on the NixOS builder before, when it was only a transitive dependency.

meejah:
Hmm, okay ... i only set it to 1.0.0 because that's what "pip freeze" said after I did the "update dependencies" thing; I'll set it to the lower one.

PR #527, opened 2021-07-30
Improve versioning in the Nix packaging

### LINKED ISSUE #526: The Nix packaged magic-folder does not build properly except from a git checkout because it has no `_version.py`
`setup.py` does some tricky things with `git` to automatically determine a version and then write it to `_version.py` for subsequent use after the git state is no longer available.

If `setup.py update_version` is run and the working tree is not a git checkout (or whatever you call a directory with a `.git` containing the usual stuff) then it silently does nothing.  Subsequently, `magic_folder/__init__.py` is broken because it cannot import `magic_folder._version`.

In most usage so far, the in-repo Nix packaging does not trip over this because it declares the magic-folder source to be the working tree itself and that working tree is almost always a git checkout (eg on CI, in a developer's working environment, etc).

However, any source tarball is generally missing `.git` and its contents.  This applies to a very common out-of-tree Nix invocation which uses the GitHub "archive" feature to get a source tarball without `.git`.  Consequently, in this usage, there is no `_version.py` and the build fails.

This blocks efforts to integrate stand-alone magic-folder with GridSync on NixOS which effectively means it blocks running GridSync w/ stand-alone magic-folder on NixOS entirely since the AppImage distribution does not work on NixOS (I don't understand why) and a pip-based installation involves building most of Qt which takes many hours, even ignoring the tedious process of setting up and maintaining the build environment.

exarkun:
Fixes #526

### DIFF
--- a/nix/magic-folder.nix
+++ b/nix/magic-folder.nix
@@ -1,7 +1,7 @@
 { lib, python, buildPythonPackage, pythonPackages, tahoe-lafs, klein, git }:
 buildPythonPackage rec {
   pname = "magic-folder";
-  version = "2020-02-05";
+  version = "2020-07-30";
   src = lib.cleanSource ../.;
 
   propagatedBuildInputs = with pythonPackages; [
--- a/nix/magic-folder.nix
+++ b/nix/magic-folder.nix
@@ -2,7 +2,7 @@
 buildPythonPackage rec {
   pname = "magic-folder";
   version = "2020-02-05";
-  src = ../.;
+  src = lib.cleanSource ../.;
 
   propagatedBuildInputs = with pythonPackages; [
     configparser
--- a/nix/magic-folder.nix
+++ b/nix/magic-folder.nix
@@ -17,7 +17,20 @@ buildPythonPackage rec {
   ];
 
   postPatch = ''
-    PATH="$PATH:${git}/bin" ${python}/bin/python setup.py update_version
+    # Generate _version.py ourselves since we can't rely on the Python code
+    # extracting the information from the .git directory that is likely not
+    # available.
+    cat > src/magic_folder/_version.py <<EOF
+
+# This _version.py is generated from metadata by nix/magic-folder.nix.
+
+__pkgname__ = "magic-folder"
+real_version = "${version}"
+full_version = "${version}"
+branch = "main"
+verstr = "${version}"
+__version__ = verstr
+EOF
   '';
 
   checkPhase = ''
--- a/nix/magic-folder.nix
+++ b/nix/magic-folder.nix
@@ -1,7 +1,7 @@
 { lib, python, buildPythonPackage, pythonPackages, tahoe-lafs, klein, git }:
 buildPythonPackage rec {
   pname = "magic-folder";
-  version = "2020-07-30";
+  version = "2021.7.30.dev0";
   src = lib.cleanSource ../.;
 
   propagatedBuildInputs = with pythonPackages; [
--- a/nix/magic-folder.nix
+++ b/nix/magic-folder.nix
@@ -1,7 +1,7 @@
 { lib, python, buildPythonPackage, pythonPackages, tahoe-lafs, klein, git }:
 buildPythonPackage rec {
   pname = "magic-folder";
-  version = "2021.7.30.dev0";
+  version = "0.1.0";
   src = lib.cleanSource ../.;
 
   propagatedBuildInputs = with pythonPackages; [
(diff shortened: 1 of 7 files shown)


nix/magic-folder.nix
```diff
@@ -17,7 +17,20 @@ buildPythonPackage rec {
...
+    cat > src/magic_folder/_version.py <<EOF
+
+# This _version.py is generated from metadata by nix/magic-folder.nix.
+
+__pkgname__ = "magic-folder"
+real_version = "${version}"
+full_version = "${version}"
+branch = "main"
+verstr = "${version}"
+__version__ = verstr
```
tomprince, line 32: I guess nothing currently cares, but I'm slightly leary of having a non-PEP 440 version here.

Thinking about it a little more, I wonder if we can get away with using `"unknown"` for the version if we don't have one (which is what tahoe does); we could set it in `magic_folder/__init__.py` if we can't import from `._version`? Though that would mean we won't notice if we deploy without a proper version. That still would be a non-PEP 440 version, but not version-like, which I think is better; it also feels more honest that a random date that doesn't get updated.

We could address the later issue for gridsync by adding some sort of integration test in that asserts the version isn't unknown.

nix/magic-folder.nix
```diff
@@ -17,7 +17,20 @@ buildPythonPackage rec {
...
+    # available.
+    cat > src/magic_folder/_version.py <<EOF
+
+# This _version.py is generated from metadata by nix/magic-folder.nix.
+
+__pkgname__ = "magic-folder"
+real_version = "${version}"
+full_version = "${version}"
+branch = "main"
+verstr = "${version}"
```
tomprince, line 31: Looking at this, it appears that none of these are used (I think we inherited them from tahoe). I'd be inclined to remove all of them, and just set `__version__` directly (here and/or in `setup.py`).

nix/magic-folder.nix
```diff
@@ -1,8 +1,8 @@
 { lib, python, buildPythonPackage, pythonPackages, tahoe-lafs, klein, git }:
 buildPythonPackage rec {
   pname = "magic-folder";
-  version = "2020-02-05";
-  src = ../.;
+  version = "2020-07-30";
```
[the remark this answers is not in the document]
tomprince: As a followup, if we go with `unknown` below, do we need to specify a version that will just keep getting out of date?

It looks like `buildPythonPackage` depends on `version`, but only to construct `name` from it and `pname` (per https://github.com/NixOS/nixpkgs/blob/40c20baacdbda70e7ac34f108634d4476010b2ee/pkgs/development/interpreters/python/mk-python-derivation.nix#L30).

I think 
```nix
buildPythonPackage rec {
    name = "magic-folder";
    meta = {
         mainProgram = "magic-folder";
    };
    ...
}
```
would have the same effect, without having a version number we need to update.

tomprince (changes requested):

nix/magic-folder.nix
exarkun: I can replace the "-" with "." to make it PEP 440.  I could also mark this as a "dev" or "pre" version or some such to make it more clear it's not an actual final/release version.

nix/magic-folder.nix
exarkun: That seems fine to me and maybe also related to meejah's comment on the linked issue.  It seems beyond the scope of this effort though.  I just mirrored what setup.py is doing here.

nix/magic-folder.nix
tomprince: I'm definitely wary about using a version with a different versioning scheme than the project uses[1]. I know that apple used (uses?) a different version scheme for clang than upstream, which has caused confusion. That is why I'd prefer `"unknown"` than something date based, even if it changed to be PEP440-compliant (without deciding to go with calver for magic-folder).

[1] I'd be fine switching to something calver based, but if we do that, we should be explicit about that change.

nix/magic-folder.nix
tomprince: True, though if we go with the option I suggested above of allowing `_version.py` to not exists, that would address both issues, without being out-of-scope.

exarkun: I'd just as soon skip the extra rounds of review while we converge on a satisfactory replacement for `_version.py` though :)

tomprince: :shrug: The important thing to me here is to not embed a weird (i.e. incompatible with "upstream") version in the python package.

meejah:
I filed https://github.com/LeastAuthority/magic-folder/issues/532

nix/magic-folder.nix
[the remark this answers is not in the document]
exarkun: Instead, I opted for a real version number that is in sync with that defined in the other package-related code/config.

nix/magic-folder.nix
exarkun: The version numbers now match.

nix/magic-folder.nix
exarkun: The version number is now PEP 440-based.  It still adheres to the semver-ish scheme that the other packaging was using.  I don't mind switching to calver or something but that seems like a separate change.

meejah (approved):
Looks good to me.

There is a ticket (#532) to expand the release plan; that would be an appropriate place to determine what the current fashionable way for Python programs to determine and publish their version is.

meejah:
p.s. maybe a link to #532 in the proto-release-doc would be good too

Issue #524, opened 2021-07-29
Provide an API to show "all tahoe object-sizes" in magic-folder


meejah:
In order to get good pricing estimates from ZKAPauthorizer it needs to know how many and of what size every Tahoe capability magic-folders currently cares about.

This means returning, per magic-folder, a list of integers. For each Snapshot, that means: the size of the "content" capability, the size of the "metadata" capability and the size of the Snapshot capability itself.

meejah:
I know we talked about estimating these, but with a couple updates to the database we can be exact since for all remote-snapshots we have the 3 capabilities (the Snapshot cap, metdata cap and content cap). We currently don't cache the latter two.

(Capabilities contain their exact size)

tomprince:
@crwood Would it be more useful to have the list of sizes of things we have already uploaded to the grid, Or the sizes we will have uploaded[1], once everything is done syncing (assuming no changes happen in the interim?

[1] Ignoring, I think intermediate snapshots; just the size of the reachable objects at that time.

crwood:
@tomprince I'm thinking it would be more useful to have the list of filesize for objects that have already uploaded to the grid (since this API will be used to determine the user's projected "monthly ZKAP cost" for all files currently stored in the grid).

tomprince:
@crwood If we are looking to get the projected monthly cost, it seems like including stuff that we are about to upload will give a better projection of that, than the current file sizes; at least on an ongoing basis. Unless the existing snapshots are renewed before we upload new ones (for which there is definitely a window), the user *won't* be charged for file already on-grid that we have newer snapshots for.

Issue #513, opened 2021-07-28
Newly-added magic-folders consistently fail to upload their contents upon being added


crwood:
I'm trying to add a directory of Cat Pictures to a storage grid using Magic-Folder. Upon adding the directory as a new magic-folder, I observe that the files are immediately inserted into the `uploads` queue (as per the websocket/"status" API message I receive), for example:

```
[...]

Issue #505, opened 2021-07-20
PR #502 breaks "file-status" HTTP API with "sqlite3.OperationalError: no such column: last_updated_ns"


crwood:
Upon updating and testing Magic-Folder after #502 (c5760cd619370ee6b67f37c7f5c85fa75c07138c), I observe an unhandled exception (spewed to `magic-folder run`'s stdout) and an HTTP 500 error when requesting a `GET  /v1/magic-folder/<folder-name>/file-status` as follows:

```
Traceback (most recent call last):
   File "/home/user/.pyenv/versions/2.7.18/lib/python2.7/site-packages/autobahn/websocket/protocol.py", line 1242, in consumeData
     self.processHandshake()
   File "/home/user/.pyenv/versions/2.7.18/lib/python2.7/site-packages/autobahn/websocket/protocol.py", line 2856, in processHandshake
     txaio.add_callbacks(f, self.succeedHandshake, forward_error)
   File "/home/user/.pyenv/versions/2.7.18/lib/python2.7/site-packages/txaio/tx.py", line 506, in add_callbacks
     future.addCallbacks(callback, errback)
   File "/home/user/.pyenv/versions/2.7.18/lib/python2.7/site-packages/twisted/internet/defer.py", line 311, in addCallbacks
     self._runCallbacks()
 --- <exception caught here> ---
   File "/home/user/.pyenv/versions/2.7.18/lib/python2.7/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
     current.result = callback(current.result, *args, **kw)
   File "/home/user/.pyenv/versions/2.7.18/lib/python2.7/site-packages/autobahn/websocket/protocol.py", line 3049, in succeedHandshake
     self._onOpen()
   File "/home/user/.pyenv/versions/2.7.18/lib/python2.7/site-packages/autobahn/twisted/websocket.py", line 301, in _onOpen
     self.onOpen()
   File "/home/user/code/magic-folder/src/magic_folder/status.py", line 107, in onOpen
     self.status.client_connected(self)
   File "/home/user/code/magic-folder/src/magic_folder/status.py", line 233, in client_connected
     self._last_state = self._marshal_state()
   File "/home/user/code/magic-folder/src/magic_folder/status.py", line 285, in _marshal_state
     for name in self._config.list_magic_folders()
   File "/home/user/code/magic-folder/src/magic_folder/status.py", line 285, in <dictcomp>
     for name in self._config.list_magic_folders()
[...]

PR #528, opened 2021-07-30
524.tahoe objects

meejah:
Add a `.../tahoe-objects` API to retrieve the sizes of all Tahoe objects we care about.

The immediate use-case for this is to pass to ZKAPAuthorizer to compute how much it will cost to renew the leases on all of them.

### DIFF
--- a/docs/interface.rst
+++ b/docs/interface.rst
@@ -192,6 +192,14 @@ The list is ordered from most-recent to least-recent timestamp.
 ``size`` is in bytes.
 
 
+GET `/v1/magic-folder/<folder-name>/tahoe-objects`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Returns a list of integers representing the sizes of all individual capabilities that this folder is using.
+That means a size for each Snapshot capability and its corresponding metadata capability and content capability.
+The list is flat; if there are 2 Snapshots on the grid this will return 6 integers.
+
+
 Status API
 ----------
 
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -1053,10 +1053,16 @@ class MagicFolderConfig(object):
                 action.add_success_fields(insert_or_update="insert")
             except (sqlite3.IntegrityError, sqlite3.OperationalError):
                 cursor.execute(
-                    "UPDATE current_snapshots"
-                    " SET snapshot_cap=?, mtime_ns=?, ctime_ns=?, size=?, last_updated_ns=?, metadata_cap?, content_cap=?"
-                    " WHERE [name]=?",
-                    (snapshot_cap, path_state.mtime_ns, path_state.ctime_ns, path_state.size, now_ns, name, remote_snapshot.metadata_cap, remote_snapshot.content_cap),
+                    """
+                    UPDATE
+                        current_snapshots
+                    SET
+                        snapshot_cap=?, mtime_ns=?, ctime_ns=?, size=?, last_updated_ns=?, metadata_cap=?, content_cap=?
+                    WHERE
+                        [name]=?
+                    """,
+                    (snapshot_cap, path_state.mtime_ns, path_state.ctime_ns, path_state.size,
+                     now_ns, remote_snapshot.metadata_cap, remote_snapshot.content_cap, name),
                 )
                 action.add_success_fields(insert_or_update="update")
 
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -1146,6 +1146,42 @@ class MagicFolderConfig(object):
                 return row[0].encode("utf-8")
             raise KeyError(name)
 
+
+    @with_cursor
+    def get_remotesnapshot_caps(self, cursor, name):
+        """
+        return all three capabilities corresponding to a given remote snapshot
+
+        :param unicode name: The name to match.  See ``LocalSnapshot.name``.
+
+        :raise KeyError: If no snapshot exists for the given name.
+
+        :returns: A 3-tuple of (snapshot-cap, content-cap, metadata-cap)
+        """
+        action = FETCH_CURRENT_SNAPSHOTS_FROM_DB(
+            relpath=name,
+        )
+        with action:
+            cursor.execute(
+                """
+                SELECT
+                    snapshot_cap, content_cap, metadata_cap
+                FROM
+                    current_snapshots
+                WHERE
+                    [name]=?
+                """,
+                (name,)
+            )
+            row = cursor.fetchone()
+            if row and row[0] is not None:
+                return (
+                    row[0].encode("utf-8"),  # snapshot-cap
+                    row[1].encode("utf-8"),  # content-cap
+                    row[2].encode("utf-8"),  # metadata-cap
+                )
+            raise KeyError(name)
+
     @with_cursor
     def get_currentsnapshot_pathstate(self, cursor, name):
         """
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -225,6 +225,8 @@ _magicfolder_config_schema = Schema([
             [snapshot_cap]     TEXT,             -- Tahoe-LAFS URI that represents the most recent remote snapshot
                                                  -- associated with this file, either as downloaded from a peer
                                                  -- or uploaded from local changes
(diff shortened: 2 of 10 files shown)


tomprince:
While I'd prefer being accurate, I worry about the cost of continually adding new fields to `current_snapshots`. I think I err slightly towards preferring to use estimates, rather than storing a bunch of extra information in that table.

If instead, we had a table of remote snapshots that we stored the information in, that would change my feelings, but that feels like a bigger change than we should be making here.

I'll also note that this doesn't include the size of any of the DMDs.

meejah:
What is "the cost" of adding two columns?

tomprince:
As I mentioned in slack, I also wonder if we should use the most recent local-snapshot size, even if it hasn't been uploaded, in the data we report (instead of the remote snapshot size).

There is also the transient cost of all the intermediate snapshots that we'll upload, but those don't represent an on-going cost., so isn't relevant to the current intended use for this API. (Though it would be good to make that clear somewhere).

meejah:
> As I mentioned in slack, I also wonder if we should use the most recent local-snapshot size, even if it hasn't been uploaded, in the data we report (instead of the remote snapshot size).

I think we should only report about actual capabilities we already know of (and care about).

tomprince:
> What is "the cost" of adding two columns?

Well, at the most basic, it is the difference between a ~30 line diff and a ~120 line diff. In addition, as I said, I think we may want to, at some point, record all the remote snapshots we've seen (or at least all our ancestors), in which case having this information here too would not be normalized[1]. It also means we need to download the snapshot cap, if we want to store it in this table; not an issue now and I don't have a case in mind where we'd want to, but it does constrain us.

In any case, changing formats of on-disk structures has a cost, since changing them requires handling migrations (as @exarkun pointed out to me in https://github.com/LeastAuthority/magic-folder/pull/451#discussion_r660871818 we have a schema migration system, we should probably start using it soon). That means that it somewhat more important to get them right than in-memory and so worth thinking about more.

From slack:
> A local snapshot will "soon" become a remote anyway, so I don't see much value in predicting the future like that...but again that could be a future enhancement if we find it's valuable

If somebody adds a 5GB file to a magic-folder, my guess is they would expect the cost estimate to update ~immediately, even before the file is on-grid. In some sense it isn't really predicting the future, but rather reporting, "once we've finished syncing, this is how much space we will be using".

[1] And sure, we could remove these fields then, but that is more work.

meejah:
There is no need to do any migration-handling until after a first release.

meejah:
> If somebody adds a 5GB file to a magic-folder, my guess [..]

...sure, and if your guess proves to be correct then we can make a future enhancement.

meejah:
> I think we may want to, at some point, record all the remote snapshots we've seen (or at least all our ancestors), [..]

As I said, "possible future enhancement" should we find that's a useful thing.

meejah:
> I'll also note that this doesn't include the size of any of the DMDs.

We can't include those without doing two extra tahoe round-trips (mutables don't have their size in the capability).

tomprince:
I will say, despite my reluctance to land this code, I think the idea you had about getting sizes is a good one.

I agree that keeping remote snapshots is something for the future (if ever), as I said in my original comment. I do think there is value in considering it as a possibility, and how we might get from here to there if we did. There are (at least) three approaches to addressing #524.
- Walking the tree and getting sizes (either here or in the tahoe plugin. This would get us an exact result, but has the downside of requiring a bunch of network round trips.
- We can estimate the size of the snapshot cap and metadata cap. This won't give us an accurate intermediate result, but as we discussed the final result (in terms of cost) won't be affected by our estimation.
- The one you have here, which caches the various caps, and uses them to calculate exact sizes. This has the benefit of giving an accurate result without needing any network round trips. However, it comes at the cost of more code; and (in my opinion) a more complex db schema, that may involve a more complex db migration in the future. While that last is debatable, we can side step the issue by choosing to do estimates instead.

I will say again, my reaction to seeing this was something like "Cool, if we can get accurate information without a bunch of round trips, so much the better. ,,, Hmm, I'm not sure if the direction the schema is a step in the right direction and there is a lot of code required for it including a bunch of unrelated tests... While I think this was good path to explore down, having done so, I think the trade-off of only having estimates is worth it, despite my distaste for that."

> There is no need to do any migration-handling until after a first release.

While that is true, as was pointed out in the comment I linked, it may be worthwhile to get some practice with it. And, there has been at least once bug (#505) filed due to not dealing with, And there are costs to dealing with now too (one obvious one is having a harder schema to read). 

> > I'll also note that this doesn't include the size of any of the DMDs.
> 
> We can't include those without doing two extra tahoe round-trips (mutables don't have their size in the capability).

We could probably do a reasonable estimate, though.

tomprince:
> > If somebody adds a 5GB file to a magic-folder, my guess [..]
> 
> ...sure, and if your guess proves to be correct then we can make a future enhancement.

Eh, I think either could be considered an enhancement (of supporting both) of the other. As I understand it, the question we are trying to answer is "if I didn't make any more changes, how long will the credits I have last me"[1]. Even if the user makes no more changes, we know that we will upload all our local snapshots, so we should taken them into account when answering that question. But really, it is not that one or the other is an enhancement, but that the specification was not clear enough; I've asked for clarification: https://github.com/LeastAuthority/magic-folder/issues/524#issuecomment-890279767

There is also the cost of uploading all those local snapshots, but my impression of the interfaces is that the assumption is, that ongoing costs will outweigh the immediate costs of those uploads. In any case, having a mix of costs that are one-time vs ongoing is definitely not something we should be handling here.

[1] Originally I had worded this "... how much will storing this cost me", but I guess the above one better reflects our payment model and UI.

crwood:
>If somebody adds a 5GB file to a magic-folder, my guess is they would expect the cost estimate to update ~immediately, even before the file is on-grid.

I'm pretty sure that ZKAPAuthorizer alone already can/does already effectively supply this information: because tokens are spent before an upload completes, users will be able to see, e.g., that the amount of ZKAPs available has decreased by 5GBs-worth once that file begins uploading (and before it finishes). This, of course, says nothing of the estimated monthly-cost for all currently-stored-object, but, as long as the API in question here provides a list of file-sizes for all currently-stored-objects on demand, that cost/value can be calculated and presented to the user at any time we choose (e.g., whenever a folder has finished syncing, whenever an upload has completed, etc.).

tomprince:
> > If somebody adds a 5GB file to a magic-folder, my guess is they would expect the cost estimate to update ~immediately, even before the file is on-grid.
> 
> I'm pretty sure that ZKAPAuthorizer alone already can/does already effectively supply this information

No. Even after #513 is resolved, if scan several large (e.g. 1G+) files, the tahoe client node (and thus ZKAPAuthorizer) won't know about all of the files until we start to upload the *last* one.

Consider the case of adding a large number of high-resolution files to a magic-folder (perhaps adding a new magic-folder pointing at an existing directory of them). My suggestion is we start reporting *immediately* how much it will cost to store that folder month-to-month, even if it will take days to upload that folder. On the other hand, the implementation suggested here will initially show a cost of 0 for the folder, and then slowly creep up over the days it takes to upload that folder.

meejah:
Can't we tackle being even-more-correct (that is, including sizes of pending uploads) as a followup? Why does that need to block this?

tomprince:
> Can't we tackle being even-more-correct (that is, including sizes of pending uploads) as a followup? Why does that need to block this?

There are a couple of reasons:
- I don't think the database changes in this PR are something that should land (I'm not going to reiterate the discussion above), and they or something else are necessary for this implementation.
- This change (mostly[1]) isn't an incremental step towards the implementation I suggest, since the logic will be quite different.

Particularly given the first point, it doesn't feel like this is worth landing. Also, the amount of work to implement my proposal feels small enough, that landing this now doesn't seem like it would save significant time or effort. I don't think the sunk cost of having written this code means we should land it; as I said, I think it was valuable effort, but the appropriate thing to after having explored this avenue is to not continue down it, and instead pick a different path.

[1] The API boilerplate bits could be re-used.

laurih01:
I tried to follow along all the comments on this, so forgive me if I'm asking something that is already answered - Is there a suggestion or alternate code to do instead of this pull request? If not, how do we move forward?

meejah:
> Well, at the most basic, it is the difference between a ~30 line diff and a ~120 line diff.

"for the record", the 'estimate sizes' PR is +254 lines, bigger than this one. If that's the metric (which is a bad one) then I guess this PR "wins".

PR #536, opened 2021-08-05
fix windows unit-test that's failing

meejah:
Attempts to fix the failing windows unit-test.

- `long` not `int` (fixed)
- sorting seems off ...

### DIFF
--- a/misc/build_helpers/sqlite_version.py
+++ b/misc/build_helpers/sqlite_version.py
@@ -1,0 +1,9 @@
+"""
+dump the underlying sqlite3 version
+"""
+
+import sqlite3
+
+print("sqlite3 version: {}".format(
+    sqlite3.sqlite_version)
+)
--- a/newsfragments/541.minor
+++ b/newsfragments/541.minor
@@ -1,0 +1,1 @@
+follow for more debugging
--- a/tox.ini
+++ b/tox.ini
@@ -29,6 +29,7 @@ commands =
 	 # version pinning we do limits the variability of this output
 	 # somewhat.
 	 pip freeze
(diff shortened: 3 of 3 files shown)


meejah:
It _looks_ like the database sorting is just broken on Windows?

I wanted to know what `sqlite3` version that build used, but I can't seem to find out where `sqlite3` comes from on Windows. The `pip freeze` output in Tox doesn't show any package named that (nor anywhere else in the raw logs from the Windows builder).

meejah:
(it was pointed out that sqlite3 is in the stdlib)

meejah:
After skipping that test in this branch, there's also an Eliot test failing ...

(Surely if "ORDER BY ... DESC" is broken on windows, _someone_ else would have reported? I've not had success finding that person w/ duckduckgo though...)

tomprince:
Looking at some results, it appears that the time resolution on windows is low enough that all the entries get the same time, so the sort order is random. #541 should demonstrate that is the case, by forcing the time we use to be always increasing.

meejah:
`test_limit` is using explicit timestamps already though?

tomprince:
> `test_limit` is using explicit timestamps already though?

Not for `last_updated_ns`, which is what we are sorting by. That can't be controlled by the caller.

I'm going to close this in favor of #541.

meejah:
What about the "dump actual sql version"  stuff?

tomprince:
> What about the "dump actual sql version" stuff?

Sure, we could keep that, though given that it *wasn't* an sqlite3 bug, I suspect that is less valuable. I'd be inclined to cherry-pick that code to its own PR though. (Or at least rebase this one to just have those changes).

PR #539, opened 2021-08-05
update conflict spec

### LINKED ISSUE #537: "conflicts" HTTP API
There is at least the start of a spec here: https://github.com/LeastAuthority/magic-folder/blob/main/docs/proposed/conflict-api.rst

This should be updated to reflect the newer API layout and to use JSON bodies (not query args).

meejah:
fixes #537 

- Updates the HTTP API layout 
- makes the filesystem API explicit
- ensure words handle "more than two" conflicts (multi-conflict) case

### DIFF
--- a/docs/proposed/conflict-api.rst
+++ b/docs/proposed/conflict-api.rst
@@ -11,42 +11,77 @@ This document is aimed at programmers working on magic-folder. It is a proposed 
 Motivation
 ==========
 
-It is desirable to have an explicit API for dealing with conflicts.
-This allows UI programs to be sure they're accurately indicating intent to magic-folders.
+Under normal operation, magic-folders scans other participants' Snapshots and reflects those changes locally; sometimes, these changes can conflict.
+
+It is desirable to have an explicit API for noticing and resolving conflicts.
+The core of magic-folders operation with users is the filesystem.
+This makes the filesystem also an API and conflicts must be part of that API.
+The command-line interface shall have sub-commands for listing and resolving conficts; these commands will use the HTTP APIs.
+
+
+Filesystem API for Conflicts and Resolution
+===========================================
+
+When a Snapshot is found to conflict with a particluar local file (say ``foo``) a "conflict file" is written beside it, reflecting the other participants' content (``foo.conflict-laptop``).
+That is, ``foo.conflict-laptop`` indicates that the participant "``laptop``" has a conflicting update
+The file ``foo.conflict-laptop`` will contain the downloaded ``.content`` of "``laptop``"'s Snapshot.
+The content of ``foo`` remains what it was when the conflict was detected.
+(Note that when multiple participants exist it's possible to have multiple ``*.confict-*`` files pertaining to a single local file).
+
+List Conflicts
+~~~~~~~~~~~~~~
+
+One can use normal directory-browsing tools such as ``ls`` to notice conflict files.
+
+
+Resolve a Conflict
+~~~~~~~~~~~~~~~~~~
+
+When _all_ ``<relpath>.confict-*`` files for a given root are deleted, the conflict is deemed resolved.
+The resolution is whatever the contents of ``<relpath>`` are currently.
+
+So, to resolve a conflict as "take theirs", one could run: ``mv foo.conflict-laptop foo`` if there was a single conflict from participant "``laptop``".
+
+To resolve a conflict as "take mine", one simply deletes ``foo.conflict-laptop`` if there was a single conflict from participant "``laptop``".
+
 
 
 HTTP API for Conflicts and Resolution
 =====================================
 
 
-
-``GET /v1/conflicts/<folder-name>``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+``GET /v1/magic-folder/<folder-name>/conflicts``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Returns a list (possibly empty) of local filesystem paths corresponding to each Snapshot that is currently in a Conflict state in the given magic-folder.
 
 Our content is in the path itself.
-The conflicting "other" content is in ``<path>.conflict-<name>`` where ``<name>`` is the petname of the user who is provided the conflicted content.
-
-Justification: we need somewhere for "theirs" versus "my" content .. I think we should still reflect this on the filesystem, even if the *API to manipulate it* is no longer there.
-This makes it more obvious for CLI users that they should check the conflicts list; the only alternative would seem to be "run some command occasionally to check for conflicts".
-I personally will forget to run this command.
+The conflicting "other" content is in ``<path>.conflict-<name>`` where ``<name>`` is the petname of the participant who is provided the conflicted content.
 
 
-``POST /v1/resolve_conflict/<folder-name>?path=<some-path>&resolution=<theirs|mine>``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The ``path`` query argument is required.
+``POST /v1/magic-folder/<folder-name>/resolve-conflict
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A JSON body is passed to this endpoint, following this example::
+
+    {
+        "relpath": "local/path",
+        "resolution": "author"
+    }
+
+The ``relpath`` key is required.
 It must be a filesystem path relative to the selected magic-folder.
 
-The ``resolution`` query argument is required.
-It must be either the string ``theirs`` or the string ``mine``.
+The ``resolution`` key is required.
+It must be the name of a current participant in the given magic-folder.
 
-It is an error if the given ``path`` in the given magic-folder is not currently in a conflicted state. In this case the response code is **404 Not Found** (XXX is this appropriate?)
+It is an error if the given ``relpath`` in the given magic-folder is not currently in a conflicted state.
 
-If the resolution is ``theirs`` then the file at ``<path>.theirs.<name>`` is moved to ``<path>`` and a new (local) Snapshot is created (with two parents).
+If the ``resolution`` is our user-name then all conflict files are deleted new (local) Snapshot is created (with parents corresponding to all conflicting participants).
 
-If instead the resolution is ``mine`` then the file at ``<path>.theirs.<name>`` is deleted and a new (local) Snapshot is created (with two parents).
+If instead the resolution is some other participant, then the content of ``<relpath>.conflict-<participant>`` is moved to ``<relpath>`` and any other conflict files are deleted.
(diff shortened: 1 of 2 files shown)


docs/proposed/conflict-api.rst
```diff
@@ -11,42 +11,77 @@ This document is aimed at programmers working on magic-folder. It is a proposed
...
+List Conflicts
+~~~~~~~~~~~~~~
+
+One can use normal directory-browsing tools such as ``ls`` to notice conflict files.
+
+
+Resolve a Conflict
+~~~~~~~~~~~~~~~~~~
+
+When _all_ ``<relpath>.confict-*`` files for a given root are deleted, the conflict is deemed resolved.
```
tomprince, line 40: Somewhere, we should indicate how this gets represented as snapshots. Do we also want to consider how to represent partially resolved conflicts?

There should also be some discussion about how we will notice and accept the resolution of a conflict by another participant, since each participant will have conflict files for every other participant that is in conflict. We probably also need to consider what to do about changes to the conflict files when accepting a conflict resolution (this seems like itself a form of conflict).

docs/proposed/conflict-api.rst
```diff
@@ -11,42 +11,77 @@ This document is aimed at programmers working on magic-folder. It is a proposed
...
+The core of magic-folders operation with users is the filesystem.
+This makes the filesystem also an API and conflicts must be part of that API.
+The command-line interface shall have sub-commands for listing and resolving conficts; these commands will use the HTTP APIs.
+
+
+Filesystem API for Conflicts and Resolution
+===========================================
+
+When a Snapshot is found to conflict with a particluar local file (say ``foo``) a "conflict file" is written beside it, reflecting the other participants' content (``foo.conflict-laptop``).
+That is, ``foo.conflict-laptop`` indicates that the participant "``laptop``" has a conflicting update
```
tomprince, line 26: Right now, these files are not treated specially by magic-folder, so are themselves snapshotted. That isn't entirely unreasonable (it is what [syncthing](https://docs.syncthing.net/users/syncing.html#conflict-handling) does, for example). However, if we do want that, we should be explicit about it.

If we do treat them normally, we likely want to explicitly record them when we create them (which means we could store the conflicted remote snapshot under the conflict name in our personal DMD). This also leads me to wonder if we should store the local version under a conflict name as well, when we trigger a conflict (so we'd end up with an identical set of `.conflict-*` files on every device, which we'd eventually get anyway, as we sync between devices).

Also, do we want to handle snapshotting a file specially (i.e. include some extra metadata), when we take a snapshot while conflict files exists?

If we *don't* want to sync `.conflict-*` files as normal files, we should file an issue so that we can implement that change.

docs/proposed/conflict-api.rst
```diff
@@ -11,42 +11,77 @@ This document is aimed at programmers working on magic-folder. It is a proposed
...
-``GET /v1/conflicts/<folder-name>``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+``GET /v1/magic-folder/<folder-name>/conflicts``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Returns a list (possibly empty) of local filesystem paths corresponding to each Snapshot that is currently in a Conflict state in the given magic-folder.
 
 Our content is in the path itself.
-The conflicting "other" content is in ``<path>.conflict-<name>`` where ``<name>`` is the petname of the user who is provided the conflicted content.
+The conflicting "other" content is in ``<path>.conflict-<name>`` where ``<name>`` is the petname of the participant who is provided the conflicted content.
```
tomprince, line 59: I think it probably makes sense to include the list of conflict files explicitly too.

docs/proposed/conflict-api.rst
```diff
@@ -11,42 +11,77 @@ This document is aimed at programmers working on magic-folder. It is a proposed
...
 
-It is an error if the given ``path`` in the given magic-folder is not currently in a conflicted state. In this case the response code is **404 Not Found** (XXX is this appropriate?)
+It is an error if the given ``relpath`` in the given magic-folder is not currently in a conflicted state.
 
-If the resolution is ``theirs`` then the file at ``<path>.theirs.<name>`` is moved to ``<path>`` and a new (local) Snapshot is created (with two parents).
+If the ``resolution`` is our user-name then all conflict files are deleted new (local) Snapshot is created (with parents corresponding to all conflicting participants).
 
-If instead the resolution is ``mine`` then the file at ``<path>.theirs.<name>`` is deleted and a new (local) Snapshot is created (with two parents).
+If instead the resolution is some other participant, then the content of ``<relpath>.conflict-<participant>`` is moved to ``<relpath>`` and any other conflict files are deleted.
+Then a new (local) Snapshot is created (with parents corresponding to all conflicting participants).
```
tomprince, line 84: I'm not sure exactly how this is intended to be used, but it seems there is somewhat of a race condition here, we could have detected a conflict with a new device between when the user indicated to their UI that they had resolved all the conflicts, and when we get notified. (Or if conflict files are themselves snapshotted, one of them could have been updated in that interval.

I think we may want to pass the participants in the conflict we intend to resolve here, rather than assuming all of them (or some otherway of indicating which state is being resolved).

docs/proposed/conflict-api.rst
tomprince: There is also the question if a particular particpant (say `laptop`) update there version of `foo` and and also `foo.conflict-laptop` is update differently? Even we we don't snapshot `.conflict-*` files, we need to address what happens if somebody makes local changes to one, and the corresponding remote update the file. Particularly the HTTP conflict resolution method seems to indicate that making edits to a conflict file and then calling that API is an expected way to deal with conflicts.

tomprince (approved):
This looks like a good start to documenting things. I do have a bunch of questions, mostly about corner cases, but also whether `.conflict-*` files should be snapshoted.

None of these questions need to be answered here, but they should be added as open questions if they aren't.

docs/proposed/conflict-api.rst
meejah: They are indicated as snapshots with multiple parents.

docs/proposed/conflict-api.rst
meejah: > If we don't want to sync .conflict-* files as normal files, we should file an issue so that we can implement that change.

Yes, this is what we want to do. There's at least a comment to that effect in the uploader i believe.

The only file that should ever be edited is the base one (e.g. `foo`). If we want a thing besides "take all changes from X" (including "us") then that's a new HTTP API that should accept a body that is the new text (i.e. distinct from the file-system-based API where you'd edit `foo` to contain the desired resolution, then delete all the conflict files).

meejah: > There is also the question if a particular particpant (say laptop) update there version of foo and and also foo.conflict-laptop is update differently?

Can't fully parse that sentence, but once there is a conflict, the downloader stops downloading things for that file so it shouldn't be "updating" any existing conflict-files (and you shouldn't edit them locally, but if you do nothing will happen).

docs/proposed/conflict-api.rst
meejah: There's no "partially resolved" state, so not sure what you mean by that?

docs/proposed/conflict-api.rst
meejah: Why? The file-names are well-defined, right?

docs/proposed/conflict-api.rst
meejah: We stop downloading once a conflict is detected so I don't think that can happen.

(Anyway, we don't implement this part yet -- there's no way out of a conflict yet, so this can be updated further when that is implemented)

docs/proposed/conflict-api.rst
tomprince: > > There is also the question if a particular particpant (say laptop) update there version of foo and and also foo.conflict-laptop is update differently?
> 
> Can't fully parse that sentence, but once there is a conflict, the downloader stops downloading things for that file so it shouldn't be "updating" any existing conflict-files 

As long as we record the appropriate parent when we resolve a conflict, we can probably get away with not updating the conflict file if the peer updates their version. That will mean the user will need to immediately resolve another conflict with the same peer, though.

If the behavior you've described is intended, it should be documented. I had assumed that it might make sense to update the `.conflict-*` file if the peer made more changes and the `.conflict-*` file was unchanged locally. If we could magically notice when a user realized there was conflict and started trying to resolve it, the ideal behavior would likely be to keep the `.conflict-*` up-to-date with the peer until the user started trying to resolve it. Given that magic is impossible, I'm not sure where on the scale of keeping the first version and keeping it up-to-date makes the most sense as a default.

> (and you shouldn't edit them locally, but if you do nothing will happen).

I've often found it useful to edit both files, when trying to merge changes, so I don't think it is reasonable to expect the user to not edit them locally.

> once there is a conflict, the downloader stops downloading things for that file

The implementation in #540 seems to force conflicts to always be resolved pair-wise. I'd've assumed that you'd want to at least publish the conflicts with every peer we can find, rather than just the first one we encounter, even if we don't update them in response to further changes in the peer.

docs/proposed/conflict-api.rst
tomprince: Well, we'd need to provide a list of participant names we conflict with at least, for them to find the filenames. (Though I'd just as soon include both)

Otherwise, they'd need to list the directory to find all the `.conflict-*` files though. If they need to that anyway, what value does this provide over just listing the folder and noticing the `.conflict-*` files.

docs/proposed/conflict-api.rst
tomprince: What I mean by "partially resolved" is having resolved the conflict with `device-b` (but not `device-a`). It could be nice to record a snapshot that has `device-b` but not `device-a` as a parent. We could (for example) deem that to be what happens when `*.conflict-device-b` is removed.

I'm not sure that is the way we'd want to implement it in the filesystem, and I don't even know if we want to support it. But I think it is reasonable to consider that as a feature we may want, and how it would be implemented. And it is probably worth  considering before we finalize the filesystem API for resolving a conflict.

meejah: What does that even mean, though, semantically?

docs/proposed/conflict-api.rst
meejah: There is no way to resolve conflicts implemented in #540

docs/proposed/conflict-api.rst
meejah: Yes, the spec says to include a list of the (conflicting) participant-names.

docs/proposed/conflict-api.rst
meejah: > I've often found it useful to edit both files, when trying to merge changes, so I don't think it is reasonable to expect the user to not edit them locally.

Obviously you can do whatever you like, but there's no semantics around doing so (is what I'm saying above).

docs/proposed/conflict-api.rst
tomprince: It would be the same as doing something like say a 2-way merge vs. a n-way merge in git.

Another way of looking at it is that is the same as if we pretended that we hadn't yet noticed a conflict with `device-a`, when resolving the conflict with `device-b`.

PR #540, opened 2021-08-06
538.conflicts db api

### LINKED ISSUE #538: Note conflicts internally
Currently, the only way a conflict is noted is via writing a specially-named file (e.g. "`foo.conflict-laptop`" for a conflict from participant `laptop` on file `foo`).

magic-folder should maintain its own notion of whether a file is in conflict or not.

(In the future, the HTTP API will provide a way to resolve conflicts, see https://github.com/LeastAuthority/magic-folder/issues/537 )

For now, all we should implement is the HTTP API to list conflicts.

meejah:
Fixes #538

Add a "conflicts" table, uses it in the downloader and includes some information about conflicts in the status API and via a `.../conflicts` API.

### DIFF
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -1263,8 +1263,6 @@ class MagicFolderConfig(object):
                 (relpath, ),
             )
             rows = cursor.fetchall()
-            if not rows:
-                return None
             return [
                 Conflict(snap_cap, author_name)
                 for author_name, snap_cap in rows
--- a/src/magic_folder/config.py
+++ b/src/magic_folder/config.py
@@ -743,7 +743,7 @@ class Conflict(object):
     Represents information about a particular conflict.
     """
     snapshot_cap = attr.ib()  # Tahoe URI
-    author_names = attr.ib(validator=instance_of(list))
+    author_name = attr.ib(validator=instance_of(unicode))
 
 
 @attr.s
@@ -1235,13 +1235,9 @@ class MagicFolderConfig(object):
             conflicts = dict()
             for relpath, author_name, snap_cap in cursor.fetchall():
                 try:
-                    conflicts[relpath].author_names.append(author_name)
-                    assert snap_cap == conflicts[relpath].snapshot_cap, "Capabilities don't match"
+                    conflicts[relpath].append(Conflict(snap_cap, author_name))
                 except KeyError:
-                    conflicts[relpath] = Conflict(
-                        snapshot_cap=snap_cap,
-                        author_names=[author_name],
-                    )
+                    conflicts[relpath] = [Conflict(snap_cap, author_name)]
             return conflicts
 
     @with_cursor
@@ -1249,8 +1245,8 @@ class MagicFolderConfig(object):
         """
         :param text relpath: snapshot relpath
 
-        :returns list: Conflict instance, or None if there are no
-            conflicts at all for `relpath`
+        :returns list: list of Conflict instances, or None if there
+            are no conflicts at all for `relpath`
         """
         with start_action(action_type="config:state-db:list-conflicts"):
             cursor.execute(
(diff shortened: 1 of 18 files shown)


src/magic_folder/downloader.py
```diff
@@ -304,6 +304,12 @@ def _process(self, snapshot):
                           capability=snapshot.capability,
                           ) as action:
             relpath = magic2path(snapshot.name)
+
+            # if we're already conflicted, no further processing
+            if self._config.list_conflicts_for(snapshot.name):
+                action.add_success_fields(is_conflict=True)
+                return
```
tomprince, line 311: I don't think this is correct.

There are at least three cases:
- we are in conflict with `device-a` and `device-b`
- we are in conflict with `device-a` but `device-b` has a descendant of our version of the file.
- we are in conflict with `device-a` but have already written a conflict file, and `device-a` now has a newer version

I think in the first two cases we still want to proceed with the logic of this function. In the third, we probably want to update the conflict file *only* if the file matches the file we wrote (which would require us to track the snapshot we wrote to the conflict file). 

We could also choose to *never* overwrite in the third case. But if we choose that option, we need to make sure that when resolving the conflict, we record as parent the version we wrote (so we'll need to write the snapshot cap to the db as well).

src/magic_folder/config.py
```diff
@@ -232,13 +233,20 @@
             [upload_duration_ns]  INTEGER        -- nanoseconds the last upload took
         )
         """,
+        """
+        --- This table represents our notion of conflicts (although they are also represented
+        --- on disk, our representation is canonical as the filesystem is part of the API)
+        CREATE TABLE [conflicted_files]
+        (
+            [name]             TEXT NOT NULL,    -- mangled name in UTF-8
+            [conflict_author]  TEXT NOT NULL     -- another participant who conflicts
```
tomprince, line 242: I think we also want to store the snapshot cap we wrote to the conflict file.

We may also want to record the pathstate of the conflict file here (particularly if we don't treat `.conflict-*` files as snapshots, see https://github.com/LeastAuthority/magic-folder/pull/539#discussion_r685525375 for discussion).

tomprince (changes requested):

src/magic_folder/downloader.py
meejah: We don't support multiple devices yet, so nothing to do here until then.
Thus, the only immediately interesting case is "in conflict with `device-a`" (which is equivalent to your first case above).

In this case we want do want to stop and do nothing else because we can't know what, if anything, the user has seen nor what they've done as far as resolving the conflict .. we only learn that when they use the HTTP API to say so or use the filesystem API to say so.

If we _update_ the conflict file after writing it the user may make a decision based on stale information (e.g. because they opened the conflict file when it first appeared). They still may take considerable time to decide what to do because this likely takes some out-of-band communication. This allows a huge window for them to "resolve" the conflict based on the stale information.

src/magic_folder/config.py
meejah: Yes, we may indeed -- but only when we provide a way to resolve conflicts (then we'll need the Snapshot capability to put in our parents list for the resolution snapshot).

src/magic_folder/downloader.py
meejah: That is to say: a) we don't support multiple users yet and; b) even if we did, we _don't_ want to ever "update" conflict-files that already got written.

If we did decide we want to give users a way to bump ahead our local notion of the conflict (i.e. update any existing conflict-files and possibly write new ones, if there are more) then it should be an explicit HTTP API along with corresponding CLI command(s).

src/magic_folder/config.py
tomprince: We want to at least store the snapshot cap now, since it isn't recoverable in the future, when we *can* resolve conflicts.

tomprince: Basically, if don't store the snapshot cap now, we won't have enough information to resolve the conflict in the future.

src/magic_folder/downloader.py
```diff
@@ -304,6 +304,12 @@ def _process(self, snapshot):
                           capability=snapshot.capability,
                           ) as action:
             relpath = magic2path(snapshot.name)
+
+            # if we're already conflicted, no further processing
+            if self._config.list_conflicts_for(snapshot.name):
```
tomprince, line 309: This should be `relpath` (similarly everywhere below).

PR #541, opened 2021-08-09
Fix failing Windows unit test.

tomprince:


src/magic_folder/test/test_config.py
```diff
@@ -839,6 +840,7 @@ def test_limit(self):
         Add 35 RemoteSnapshots and ensure we only get 30 back from
         'recent' list.
         """
+        self.patch(self.db, '_get_current_timestamp', itertools.count().next)
```
meejah, line 843: should this be "with self.patch()" or does it magically handle its own cleanup?

tomprince: It does handle its own cleanup (via `addCleanup`).

meejah (approved):
great, seems to work ..
(after linter etc happy plz merge)

PR #551, opened 2021-08-11
Add a service that calls a function periodically or on-demand.

### LINKED ISSUE #552: Add a service that calls a function periodically or on-demand.
In #534, we started calling `RemoteSnapshotCreator.upload_local_snapshots` for every local snapshot created, and protected it with a lock so it would only run one such call at a time. However, that function process multiple snapshots at a time, so we end up with a lot of unnecessary calls.

To fix this, we should replace the lock with a new primitive that will de-duplicate explicit calls to the function while it is already running, so it only gets called one additional time after the current call is finished.  It should also interact with the periodic calls to the function, so that they don't trigger an additional call to the function if it is running when they'd be triggered.

tomprince:
Fixes #552. 

Use this to ensure that we don't have duplicate calls to `UploaderService.perform_upload`.

### DIFF
--- a/src/magic_folder/test/test_util_twisted.py
+++ b/src/magic_folder/test/test_util_twisted.py
@@ -1,0 +1,201 @@
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+
+import attr
+from eliot import log_call, start_action
+from eliot.testing import UnflushedTracebacks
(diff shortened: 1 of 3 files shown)


tomprince:
I'm particularly interested in a review of `PeriodicService` design/code/tests.

PR #588, opened 2021-11-05
587.cancel behavior

### LINKED ISSUE #587: Leaving folders that are actively synchronizing becomes impossible
In the Gridsync integration test suite, I have a simple helper/teardown function that some tests use to leave all magic-folders for a given Magic-Folder instance, like so:

```
@inlineCallbacks
def leave_all_folders(magic_folder):
    folders = yield magic_folder.get_folders()
    for folder in list(folders):
        yield magic_folder.leave_folder(folder)
```

(where `leave_folder` sends an authenticated `DELETE /v1/magic-folder/<folder>` request)

Up until recently, this has worked fine: I could create/join numerous (~10) magic-folders and use the function above to leave them all in quick succession without problems. As of the current Magic-Folder main@HEAD (revision 8e44c20c8695069522abb34cd594875dd80b59dd), however, this function causes Magic-Folder to hang: typically the first three or four folders will leave successfully (as indicated by receiving a response body of `{}`), but the next subsequent request will hang and never return anything.

If I introduce a slight pause in between requests like so, however, the requests complete successfully:

```
@inlineCallbacks
def leave_all_folders(magic_folder):
    folders = yield magic_folder.get_folders()
    for folder in list(folders):
        yield deferLater(reactor, 0.1, lambda: None)
        yield magic_folder.leave_folder(folder)
```

Given this, I suspect that there's some deeper problem here with rapid configuration updates (maybe related to the `DeferredLock` that was introduced in #583?), and that this problem was introduced somewhere between revision ed4af492cf8ce3a76c48d429a44a510ba17ea816 and revision 8e44c20c8695069522abb34cd594875dd80b59dd (since if I roll back to ed4af492cf8ce3a76c48d429a44a510ba17ea816 I can leave all folders in quick succession without the hang).

I will try to get some logs shortly...

meejah:
fixes #587

### DIFF
--- a/integration/test_add.py
+++ b/integration/test_add.py
@@ -125,8 +125,8 @@ def test_leave_many(request, reactor, temp_filepath, alice, bob):
     # try and ensure that the folders are "doing some work" by adding
     # files to them all (sizes are in KiB)
     fake_files = (
-        ('zero', 100),
-        ('one', 10000),
+        ('zero', 100),  # 100K
+        ('one', 10000), # 10M
     )
     for fname, size in fake_files:
         for name in names:
--- a/integration/test_add.py
+++ b/integration/test_add.py
@@ -49,7 +49,7 @@ def test_leave(request, reactor, temp_filepath, alice, bob):
     """
     magic = temp_filepath
     original_folder = magic.child("cats")
-    recover_folder = magic.child("kitties")
+    recover_folder = magic.child("cats_recover")
     original_folder.makedirs()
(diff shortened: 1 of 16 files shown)


meejah:
Still needs proper tests

meejah:
once again, codecov appears to be "simply wrong". Locally on a clean build with coverage (using `cuv report`) I see `94.6%: 227 covered of 240 added lines (leaving 13 missing)`

Issue #591, opened 2021-11-22
magic-folder-api should load config read-only


meejah:
There are some occasional spurious integration test failure on Windows and MacOS to do with "database being locked".

My current theory is that this is due to running `magic-folder-api` subcommands when the "main" process is doing database-stuff (although I'm unsure why they don't just fail 100% of the time, unless sqlite3 only actually locks the database when there's a write transaction?)

In any case, it would be better practice to have `magic-folder-api` only open in read-only mode since it doesn't need write-access and -- ideally -- that would also fix these occasional problems.

(It would _also_ be possible to fix by just doing HTTP transaction instead of running `magic-folder-api` commands, but the latter is what most downstream CLI users would usually do so it's a better "integration test" IMO if the integration tests do that too...)

meejah:
See also https://github.com/LeastAuthority/magic-folder/issues/339

meejah:
Investigation required: how does locking work with sqlite, and is it platform-dependent?

PR #602, opened 2021-12-22
Depend on Tahoe-LAFS 1.17.0

### LINKED ISSUE #597: Upgrade Tahoe-LAFS dependency to 1.17.0
Tahoe 1.17.0 is out and includes a number of security-related fixes. We should make it possible to use magic-folder with this version of Tahoe so these fixes are available to magic-folder users.  Almost all of these fixes are for server-side Tahoe-LAFS functionality (in the introducer and storage nodes) so this is not an *urgent* upgrade for magic-folder from a security perspective.

A factor that does add some urgency is that ZKAPAuthorizer will soon need to require exactly Tahoe-LAFS 1.17.0 (support for multiple versions is difficult because of ZKAPAuthorizer's use of internal Tahoe-LAFS Python APIs) so that ZKAPAuthorizer-enabled servers can be upgraded to receive the security fixes.  Since ZKAPAuthorizer includes client and server code in the same Python package, both pieces are forced to have the same requirements.

There are other good (if also not urgent) reasons to upgrade to Tahoe-LAFS 1.17.0 - notably, since that release introduces Python 3 compatibility, it becomes possible to seriously consider porting magic-folder to Python 3.

exarkun:
Bump the dependency on Tahoe-LAFS to 1.17.0.  This is the library-dependency change.  This is different from the node-dependency change (which is widened to include 1.17.0 by #601).

Stacked on #601 
Fixes #597


exarkun:
The CI failures are different from each other but also both against Tahoe-LAFS 1.14.  I wonder if they're problems with 1.14 or just intermittent failures.

meejah:
> The CI failures are different from each other but also both against Tahoe-LAFS 1.14. I wonder if they're problems with 1.14 or just intermittent failures.

Do we care about 1.14? Maybe should just drop it now...

exarkun:
> > The CI failures are different from each other but also both against Tahoe-LAFS 1.14. I wonder if they're problems with 1.14 or just intermittent failures.
> 
> Do we care about 1.14? Maybe should just drop it now...

Seems reasonable to me.  It's over 20 months old and 1.15 has been available for over over 14 months.  That said, I don't know if 1.14 is really the problem here or if it just got unlucky.

meejah:
> > > The CI failures are different from each other but also both against Tahoe-LAFS 1.14. I wonder if they're problems with 1.14 or just intermittent failures.
> > 
> > 
> > Do we care about 1.14? Maybe should just drop it now...
> 
> Seems reasonable to me. It's over 20 months old and 1.15 has been available for over over 14 months. That said, I don't know if 1.14 is really the problem here or if it just got unlucky.

I think "unlucky" -- some of the `magic-folder-api` subcommands read the config-database and "sometimes" they get unlucky on mac or windows where you can't read the database like that (see https://github.com/LeastAuthority/magic-folder/issues/591 )

PR #611, opened 2022-01-19
196.cooperate

### LINKED ISSUE #196: Cooperatively copy content to the stash in `create_snapshot`
#141 introduces `create_snapshot` in `magic_folder/snapshot.py`.  In the code beneath the comment marked "2." it reads from a source file and writes to a stash file.  If the stash file is large, this may take a long time.  While this is happening, the magic-folder process is unable to perform other tasks (like reacting to filesystem change notifications or servicing API requests).

An early version of this function was written using `inlineCallbacks` and included a `yield` in the body of the loop.  However, `inlineCallbacks` does not context switch on a bare yield.  Thus the whole file is copied before any context switch can happen.

The smallest code change to address this would be to yield something like `twisted.internet.task.deferLater(reactor, 0.0, lambda: None)` which will allow a context switch (and then resume "very soon").  This may be a good first improvement.  There is room for further improvement, though:

  * `deferLater` after every iteration will be insensitive to the speed of the copying operation.  Compared to a slow disk, a fast disk may be able to do twice as many iterations in the same amount of time but both will suffer the same reactor-turn overhead with this implementation, disproportionately slowing down magic-folder on the system with the faster disk.   
  * `twisted.internet.task.cooperate` may be a solution to this, as it measures the passage of time and makes context switching decisions based on that.
  * However, `cooperate` uses the global reactor which makes it test-unfriendly.  Using it requires jumping through a few extra hoops to be able to write good tests.  The strategy I've used previously is to parameterize `cooperate` itself and then let tests supply their own implementation - either one tied to the test's reactor or one that just consumes the entire task synchronously.  Whether this makes sense here or not may depend on how many layers we have to thread `cooperate` through to reach `create_snapshot`.
  * We may want an even smarter scheduler - for example, one that takes into account new change notifications for files that are in the process of being copied.  If inotify tells us of a new write to a file that's being copied, there's a good chance the in-progress copy is now largely garbage.  We might not want to _unconditionally_ abandon that snapshot but abandoning it in some cases may make sense.

meejah:
fixes #196

### DIFF
--- a/integration/test_synchronize.py
+++ b/integration/test_synchronize.py
@@ -311,16 +311,17 @@ async def test_internal_inconsistency(request, reactor, temp_filepath, alice, bo
         timeout=25,
     )
 
-    await bob.stop_magic_folder()
+    await bob.stop_magic_folder()  # restarted in "finally" below
 
-    # update the file (so now there's two versions)
-    content1 = non_lit_content("one")
-    original_folder.child("fluffy").setContent(content1)
-    await take_snapshot(alice, "internal", "fluffy")
+    try:
+        # update the file (so now there's two versions)
+        content1 = non_lit_content("one")
+        original_folder.child("fluffy").setContent(content1)
+        await take_snapshot(alice, "internal", "fluffy")
+        await twisted_sleep(reactor, 2)
 
-    await twisted_sleep(reactor, 2)
-
-    await bob.start_magic_folder()
+    finally:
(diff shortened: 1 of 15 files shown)


meejah:
As per https://github.com/LeastAuthority/magic-folder/runs/5048853188?check_suite_focus=true#step:8:8 there is at least one vote that codecov is "just wrong".

PR #613, opened 2022-01-20
612.tahoe req

### LINKED ISSUE #612: The distutils metadata pinning Tahoe-LAFS precludes use reasonable, compatible versions
`requirements/base.in` declares a dependency on exactly Tahoe-LAFS 1.17.1 and `setup.py` loads this in and records it as the `install_requires` metadata.  If there were an upstream `1.17.1.postN` release (or if a downstream packager produced one) this would be unusable with magic-folder.

It would be nice to allow these postN releases.

meejah:
fixes #612


exarkun (approved):
Thanks.  This works for me.

PR #615, opened 2022-01-25
559.capabilities

### LINKED ISSUE #559: Protect capabilities from appearing in logs
Capabilities are secret and should never be revealed in logs. Users cannot be expected to know that logs contain secret information that could give someone access to all their files (e.g. if they post them on a help forum or send logs to a provider).

One way to achieve this could be to create an object to wrap capabilities instead of using a byte-string. The default `__str__` for such an object would output `[REDACTED]` or similar so that logs don't accidentally show capabilities.

Much of the Eliot logs, for example, are littered with capabilities. Currently it is up to "outside software" to ensure these are censored properly, which is obviously not ideal.

meejah:
fixes #559

### DIFF
--- a/integration/test_add.py
+++ b/integration/test_add.py
@@ -7,7 +7,9 @@ from eliot.twisted import (
     inline_callbacks,
 )
 
-from magic_folder.util.capabilities import to_readonly_capability
+from magic_folder.util.capabilities import (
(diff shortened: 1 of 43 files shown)


meejah:
codecov is wrong, again.

PR #666, opened 2022-09-01
shut down on stdin close

### LINKED ISSUE #665: Automatically exit
When running magic-folder as a subprocess the parent needs a way to ensure its children shut down.

This is, unfortunately, not a straightforward problem. Some support from the children can be useful.

Thus, we add a feature whereby if `stdin` is closed, magic-folder will exit. This means that when run as a child process, magic-folder will exit if the parent goes away (even if it can't have time to e.g. run its own cleanup / kill).

meejah:
fixes #665


meejah:
Can't get more than 100% patch coverage...sorry, codecov.

PR #667, opened 2022-09-06
test windows integration tests

meejah:
#666 seems to be suffering from some unrelated integration test failures. Do they pass on a very recent "main" branch?


meejah:
see #669 uncovered in the above test-run .. maybe i'll find more!

PR #669, opened 2022-09-06
Conflicts on Personal DMD / local-state mismatch

### LINKED ISSUE #668: Intermittant Windows failure: DMD vs. local state mismatch?
An intermittent Windows failure started happening again "somewhat" repeatably, so I decided to track it down. It's in the `test_internal_inconsistency` integration test.

The test involves "alice" and "bob" and a single folder.

- "alice" creates the folder, adds "bob" and puts a single file into it
- "bob" downloads the single update to the single file
- "bob" is stopped
- "alice" then updates the file (i.e. uploads a new version) before re-starting "bob"
- when bob re-starts, we (sometimes!) get a conflict .. and thus the test ends up failing (because the content isn't correct).
- (In correct runs, there is no conflict).

This only happens sometimes. Looking at the progress of the state-machines on good versus bad runs, I believe that what happens is that bob makes it far enough to update the database (`[snapshots]` table) but _not_ far enough to update bob's Personal DMD.

In such a case, when bob re-starts the new change from alice is noticed and downloaded. Then the "ancestor" checks fail, which I _believe_ is because of the mismatch between the local state and the Personal DMD. This results in a conflict.

meejah:
fixes #668


meejah:
Dead end

PR #671, opened 2022-09-14
670.pid files

### LINKED ISSUE #670: pid-file behavior
Currently, magic-folder produces a "pid-file".

When `magic-folder run` starts, this pid-file is checked.

- If there is no file, startup proceeds.

- If there is a file:
  - the PID is read from it
  - if no such process exists, startup proceeds
  - if the process exists:
    - ..and the command-line appears to be a magic-folder, that one is killed and startup proceeds
    - otherwise an error is produced and the process exits

We may wish to change the latter behavior. Another way to handle this would be to assume that the PID file is stale and some other process happened to get that PID (so, proceed with startup).

A port of BSD pid-file handling (https://linux.die.net/man/3/pidfile) additionally uses locking (we can't on Windows, but otherwise could do this too) and suggests the correct behavior is to exit if there is any process running with the PID in the pid-file.

The purpose of having a pid-file in the first place is to avoid running a second instance of magic-folder upon the same configuration (which could corrupt it or lead to other undefined behavior). We could simplify our behavior to the above-linked BSD-like behavior (leaving it up to the caller to decide if the pid-file is stale or not). That is, don't do the optional kill and always exit.

This would mean that any "parent" program such as Gridsync would be responsible for deciding "is this a magic-folder instance or not" .. that is, the "if the process exists" logic above. If a human is using the command-line to run the program (or some manager like systemd) then they must decide if it is stale or not.

meejah:
fixes #670


meejah:
Local tools say 98/103 or 95% line-coverage of the diff. Again unclear how codecov is reaching a different conclusion

PR #694, opened 2023-02-18
686.status updates

### LINKED ISSUE #686: "Changes only" status output
The status endpoint currently has the behavior of sending the entire JSON-serialized status state on every change to that state.

Instead, all clients could be kept in sync by using the following protocol:

- the first message is always a "full state" message
- any number of "update" messages may follow
- (think; optional? required?) any message may be a "full state" message, after which any "update" messages pertain to changes to the latest "full state" message

- on any incoming "full state" message, reset all state (e.g. GUI controls) to match this message
- on "update" messages, change the existing state as per the message

We should spec out all the messages; see e.g. `interface.rst` and the status-API (there's just one message currently). This would make the entire API `/v2` I guess? (Or, how do we tell existing clients what version this is?)

...and for compatibility the `/v1` endpoint would have to keep doing what it does now, until deprecation? Also we don't have a way to deprecate APIs.

meejah:
Fixes #686 adding an "events" based status output (instead of the "full status each time it changes" method before).


meejah:
I'd like to merge ... or change? .. this. @crwood did you have a look as to whether it serves Gridsync's purposes?

Another main question is: do we want backwards-compat etc or make this events-based API the _only_ API? (from the `magic-folder status` perspective this works fine).

docs/interface.rst
```diff
@@ -325,74 +325,50 @@ There is a WebSocket-based status API located at ``/v1/status``.
...
+Thereafter, most messages will include only a single event (although clients should handle any number).
+
+Every event has a ``"kind"`` key describing what sort of message it is.
+Events will contain other keys; clients should be tolerant of keys in the state they don't understand.
 
 The client doesn't send any messages to the server; it is an error to do so.
+
+The follow event kinds are understood (see ``status.py`` for more details on the sending side, and ``cli.py`` for an example of receiving them):
+
+- ``"scanner"``: has key ``last-scan`` which is a unix-timestamp saying when we last looked for local changes.
```
crwood, line 345: In keeping in line with the suggestion above, consider `scan-finished` or `scan-completed` here.

docs/interface.rst
```diff
@@ -325,74 +325,50 @@ There is a WebSocket-based status API located at ``/v1/status``.
...
+
+- ``"poller"``: has a key ``last-poll`` describing when we last asked for remote changes.
```
crwood, line 347: Suggestion: rename `poller` -> `poll-finished` (or `poll-completed`).

docs/interface.rst
```diff
@@ -325,74 +325,50 @@ There is a WebSocket-based status API located at ``/v1/status``.
...
+
+- ``"tahoe"``: describes the status of our connected Tahoe-LAFS client: ``connected`` and ``desired`` are the number of servers we are conencted to (and how many we want). Whether we are currently connected to enough is in a boolean ``happy``.
```
crwood, line 349: Suggestion: rename `tahoe` -> `connection-changed` (or `connection-updated`).

docs/interface.rst
```diff
@@ -325,74 +325,50 @@ There is a WebSocket-based status API located at ``/v1/status``.
...
+
+- ``"error"``: An error, with ``timestamp`` and ``summary`` (human-readable string).
```
crwood, line 351: Suggestion: rename `error` -> `error-occurred`.

docs/interface.rst
```diff
@@ -325,74 +325,50 @@ There is a WebSocket-based status API located at ``/v1/status``.
...
+
+- ``"folder-add"``: Key ``folder`` says which folder was added.
```
crwood, line 353: Suggestion: rename `folder-add` -> `folder-added`.

docs/interface.rst
```diff
@@ -325,74 +325,50 @@ There is a WebSocket-based status API located at ``/v1/status``.
...
+
+- ``"folder-delete"``: Key ``folder`` says which folder was deleted.
```
crwood, line 355: Suggestion: rename `folder-delete` -> `folder-removed` or even `folder-left` (since the term "delete" usually corresponds to a filesystem event, but removing/leaving a folder from magic-folder (the application) doesn't delete the corresponding directory from the filesystem).

docs/interface.rst
```diff
@@ -325,74 +325,50 @@ There is a WebSocket-based status API located at ``/v1/status``.
...
+...where the ``events`` list contains some non-zero number of event messages.
+The first message, upon connect, will likely contain many events: enough to give a consistent view of the current state.
+Thereafter, most messages will include only a single event (although clients should handle any number).
+
+Every event has a ``"kind"`` key describing what sort of message it is.
+Events will contain other keys; clients should be tolerant of keys in the state they don't understand.
 
 The client doesn't send any messages to the server; it is an error to do so.
+
+The follow event kinds are understood (see ``status.py`` for more details on the sending side, and ``cli.py`` for an example of receiving them):
```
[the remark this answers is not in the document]
meejah: These seem reasonable (some of the present-tense stuff is a holdover from when it was "state" I guess and that made sense).

Shall I fix these up and merge...?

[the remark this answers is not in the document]
crwood: >Shall I fix these up and merge...?

If you don't mind, please hold off for a little longer, as I'm still trying to unbreak my integration code/tests (which depended heavily on the old status message format) and narrow down what else might be missing or behaving differently...

One other issue I've noticed so far is that according to the docs, the new "error" event no longer provides the folder-name context. This prevents consumers of this new API from being able to tell which folder an error corresponds to. To elaborate, with the old "state"-based status API, "error" messages were scoped/nested beneath the name of folder that they corresponded to, like so:

```
{
  "state": {
    "synchronizing": False,
    "folders": {
      "Documents": {
        "uploads": [],
        "downloads": [],
        "errors": [
          {
            "timestamp": 1678901234,
            "summary": "Something went wrong!",
          }
        ],
        "recent": [],
        "tahoe": {"happy": True, "connected": 1, "desired": 1},
        "scanner": {"last-scan": 1678901234.5678},
        "poller": {"last-poll": 1678901234.5678},
        }
      }
   }
}
```

From this, it's easy/possible to tell that the "Something went wrong!" error occurred in the "Documents" folder and not some other folder. With the new "events"-based API though, we only get a `timestamp` and `summary` keys. It would be nice if the folder-name could be included in "error" event messages if/where the error is scoped to a particular folder (perhaps with a `None`/`null` value in the event that there is no corresponding folder?) or if there were two "error" message kinds -- for example a "folder-error-occurred" kind for errors that do correspond to a particular folder (and whose messages would always have a "folder" key/value) and a "general-error-occurred" kind for errors that do not.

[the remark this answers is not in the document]
meejah: Yes, might be just a docs thing then. I think it's also _optional_ actually: there are errors that are possible that aren't scoped to a folder (e.g. "can't connect to tahoe" could be one).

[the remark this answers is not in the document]
meejah: double-checking the code, the event API will only ever receive `error` events with the `folder` specified -- so it's just a doc issue then

meejah:
Okay, i updated the spelling of events (and other review things). Won't merge yet though

docs/interface.rst
```diff
@@ -325,74 +325,50 @@ There is a WebSocket-based status API located at ``/v1/status``.
...
+
+- ``"upload-finished"``: a file (``relpath``) in a folder (``folder``) has completed. An ``upload-started`` will always preceed this.
+
+- ``"download-queued"``: same as upload version.
+
+- ``"download-started"``: same as upload version.
+
+- ``"download-finished"``: same as upload version.
+
+All timestamps are "seconds since the Unix epoch", as floats.
```
crwood, line 369: This line is untrue. In the case of `error-occurred`, [`timestamp` is an integer](https://github.com/LeastAuthority/magic-folder/blob/d3cd024b7229475e0b5a63d05233e3837367b76a/src/magic_folder/status.py#L283), not a float. Consider either updating `error-occurred` to use a float too, or updating this line accordingly.

Further related to consistency: `timestamp` is used across three messages, but some other messages that effectively contain timestamps are calling those timestamps by other names (`queued-at`, `started-at`). Perhaps they should all just be `timestamp`? (Also, is there a reason why most of the messages have timestamps but some of them don't? Perhaps _every_ message should have a timestamp?)

exarkun: Note JSON has neither integers nor floats, [only numbers](https://www.rfc-editor.org/rfc/rfc8259#section-6).

meejah: I thought I changed them all to `timestamp` so yes can change `queued-at` and `started-at` too then.

Should I just say "number" there? (I think I put "float" because that's more accurate for Python as to what JSON really does). Or should we serialize timestamps differently altogether (probably the other contender is ISO8601 strings?)

Note that the timestamps represent the time when the thing happened, not when the message is sent. But, sure, if timestamps can benefit other place we could add them. (It might be better to do that as a separate PR, as this was _supposed_ to be simply mechanical: going from "here is the state" to "here are events" -- but containing the same information).

Implementation-wise, internally we do already have and record timestamps for the things that include them, but not for everything. "folder add" doesn't record when it was done and there's nowhere sensible to record folder-delete timestamps? (unless we just do it ephemerally). Those are the only two right?

exarkun: > Should I just say "number" there? (I think I put "float" because that's more accurate for Python as to what JSON really does). Or should we serialize timestamps differently altogether (probably the other contender is ISO8601 strings?)

Since this is for the JSON interface, I think speaking in terms of JSON makes the most sense.  It also means that if a particular JSON implementation decides to spit out an `int` then the docs have not been proven wrong. :)

ISO8601 is _probably_ a reasonably good idea for timestamp representation but I don't know if it makes that much difference here.  It is more easily human-readable so it would make manual testing against the API or debugging by reading raw logs a bit more convenient.  Of course it would also be very easy to build a simple tool to improve the presentation of raw logs for any human who wants to get the info.

I'd probably lean towards putting ISO8601 timestamps on the todo list for v2 of the API and then forgetting about it for a while.

meejah: Isn't `"foo": 0` still valid JSON, meaning "the float 0.0"...? (Similarly with `1` etc)

In any case, Python does indeed produce and consume JSON like that. (The RFC says "...which may be followed by a fraction part and/or an exponent part" so I think that's still standards-compliant).

Anyway, I'll change it to "number".

PR #696, opened 2023-03-06
Add script to create test directories & files

hacklschorsch:
Copy-pasted from https://whetstone.private.storage/privatestorage/privatestoragedesktop/-/issues/761#note_27299


meejah:
If we want to keep this ad-hoc test code, it needs linting fixes and resolution of the review comment.

PR #729, opened 2023-06-01
expose invites in status API

### LINKED ISSUE #708: Invite-related events
As a follow-up for https://github.com/LeastAuthority/magic-folder/issues/686#issuecomment-1398650148 and after https://github.com/LeastAuthority/magic-folder/pull/694, consider also emitting "events" messages for the recently-added operations relating to magic-folder invites. Examples of such messages might be something like:

```
{"event": "invite-created", "folder": "Cat Pics", "participant-name": "Bob's Laptop", "id": "40fe4a79-5012-40dc-9574-ae2160b57797", "wormhole-code": "7-guitarist-revenge" }
{"event": "invite-updated", "folder": "Cat Pics", "participant-name": "Bob's Laptop", "id": "40fe4a79-5012-40dc-9574-ae2160b57797", "wormhole-code": "7-guitarist-revenge", "state": "Offer sent; waiting for response" }
{"event": "invite-succeeded", "folder": "Cat Pics", "participant-name": "Bob's Laptop", "id": "40fe4a79-5012-40dc-9574-ae2160b57797", "wormhole-code": "7-guitarist-revenge" }
{"event": "invite-failed", "folder": "Cat Pics", "participant-name": "Bob's Laptop", "id": "40fe4a79-5012-40dc-9574-ae2160b57797", "wormhole-code": "7-guitarist-revenge", "reason": "Key-confirmation failed" }
{"event": "invite-rejected", "folder": "Cat Pics", "participant-name": "Bob's Laptop", "id": "40fe4a79-5012-40dc-9574-ae2160b57797", "wormhole-code": "7-guitarist-revenge" }
{"event": "invite-cancelled", "folder": "Cat Pics", "participant-name": "Bob's Laptop", "id": "40fe4a79-5012-40dc-9574-ae2160b57797", "wormhole-code": "7-guitarist-revenge" }
```

"Events" messages like these would prevent clients from having to manually query the `.../invites` endpoint in order to track state-related information and/or remove the need to await on (and, e.g., keep a reference to the Deferred for) the result of the `.../invite-wait` endpoint. In addition, having an "invite-updated" event for each "[phase](https://github.com/LeastAuthority/magic-folder/blob/main/docs/invite-diagram.seqdiag)" of the invite process would allow for finer-grained progress messages to be displayed to the user (since, currently, `.../invite-wait` will only show the "final result" of an invite; there is no way to get intermediate/progress information without polling).

meejah:
Fixes #708

### DIFF
--- a/docs/interface.rst
+++ b/docs/interface.rst
@@ -409,4 +409,5 @@ If one of these files is currently being uploaded, there will also be a ``upload
 To know the state of all files, use the other endpoints.
 
 
-The *invite events* follow an ordered pattern: ``invite-created`` will be followed by 3 ``invite-updated`` events (one each for ``welcome``, ``versions`` and ``code`` as those arrive) and then finalized by one of the "terminal" events (``invite-succeeded``, ``invite-failed``, ``invite-rejected`` or ``invite-cancelled``).
+The *invite events* follow an ordered pattern: ``invite-created`` will be followed by ``invite-created``, ``invite-welcomed``, ``invite-code-created`` and then finalized by one of the "terminal" events (``invite-succeeded``, ``invite-failed``, ``invite-rejected`` or ``invite-cancelled``).
+These can thus form the basis of a progress-indicator or other UX treatments.
--- a/docs/interface.rst
+++ b/docs/interface.rst
@@ -385,7 +385,7 @@ The following event kinds are understood (see ``status.py`` for more details on 
 
 - ``"download-finished"``: same as upload version.
 
-- ``"invite-created"``: A new invite is created. All invite events also have a ``folder`` indicating which folder they pertain to, a ``uuid``, ``participant-name`` indicating the invitee and ``mode`` (``"read-only"`` or ``"read-write"``).
+- ``"invite-created"``: A new invite is created. All invite events also have a ``folder`` indicating which folder they pertain to, a ``id``, ``participant-name`` indicating the invitee and ``mode`` (``"read-only"`` or ``"read-write"``).
 
 - ``"invite-welcomed"``: Connection to the mailbox has succeeded. As well as the keys from ``"invite-created"`` the welcome data is contained in ``welcome`` (you should show users the ``"motd"`` if there is one).
 
--- a/docs/interface.rst
+++ b/docs/interface.rst
@@ -403,3 +403,6 @@ Note that the first "update events" message received will _not_ contain all the 
 For example, if there are 50 files in the folder and 48 have already been uploaded, there will be just 2 ``upload-queued`` events (because the other 48 have all finished already).
 If one of these files is currently being uploaded, there will also be a ``upload-started`` event.
 To know the state of all files, use the other endpoints.
+
+
+The *invite events* follow an ordered pattern: ``invite-created`` will be followed by 3 ``invite-updated`` events (one each for ``welcome``, ``versions`` and ``code`` as those arrive) and then finalized by one of the "terminal" events (``invite-succeeded``, ``invite-failed``, ``invite-rejected`` or ``invite-cancelled``).
--- a/docs/interface.rst
+++ b/docs/interface.rst
@@ -393,7 +393,7 @@ All invite events also have a ``folder`` indicating which folder they pertain to
 - ``"invite-code-created"``: We have allocated what we need on the mailbox server and have a valid code.
 As well as the keys from ``"invite-created"``, a ``code`` key will be included indicating the wormhole code.
 
-- ``"invite-received-versions"``: We have received an encrypted message from our peer.
+- ``"invite-versions-received"``: We have received an encrypted message from our peer.
 As well as the keys from ``"invite-created"``, a ``versions`` key is included indicating the peer's version information.
 
 - ``"invite-succeeded"``: An invite is successful. All the keys from `"invite-created"`.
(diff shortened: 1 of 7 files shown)


docs/interface.rst
```diff
@@ -385,9 +385,24 @@ The follow event kinds are understood (see ``status.py`` for more details on the
 
 - ``"download-finished"``: same as upload version.
 
+- ``"invite-created"``: A new invite is created. All invite events also have a ``folder`` indicating which folder they pertain to, a ``uuid``, ``participant-name`` indicating the invitee and ``mode`` (``"read-only"`` or ``"read-write"``).
```
crwood, line 388: Other RESTful endpoints (beneath `/experimental/magic-folder/<folder>/invite-*`) use `id` for the UUID whereas these use `uuid`. For consistency's sake -- and to signal to consumers of these APIs that the values are interchangeable/compatible -- consider using the same key/field name throughout.

docs/interface.rst
```diff
@@ -385,9 +385,24 @@ The follow event kinds are understood (see ``status.py`` for more details on the
...
+
+- ``"invite-updated"``: An update to an invite has happened; as well as the keys from ``"invite-created"`` there may be one or all of: ``welcome``, the Welcome message from the Magic Wormhole server (you should show users the ``"motd"`` if there is one); a ``versions`` indicating the app-version support of the other side; or a ``code`` indicating the wormhole code.
```
crwood, line 390: I wonder if this event kind should be split up into (three?) separate ones -- e.g., "invite-welcome-received", "invite-versions-received", "invite-code-received" (or similar) -- since a) there are only a few such expected "subkinds", b) separating it this way would allow consumers to hook these messages up more directly (i.e., without the extra parsing), and c) this sort of separation would be more consistent with the "terminal" messages below (where you provide separate event  kinds -- "succeeded", "failed", "rejected", "cancelled" -- for all the different ways that an invite can be completed).

src/magic_folder/status.py
```diff
@@ -311,6 +348,57 @@ def _marshal_event_poll_completed(folder_name, poller):
...
+    msg = _invite_to_json(invite)
+    msg["folder"] = folder_name
+    msg["reason"] = str(reason)
+    return _marshal_event("invite-rejected", msg)
+
+
+def _marshal_event_invite_cancelled(folder_name, invite):
+    msg = _invite_to_json(invite)
+    msg["folder"] = folder_name
+    return _marshal_event("invite-canceled", msg)
```
crwood, line 399: Everywhere else this is `invite-cancelled` (with two lowercase "L"s) so this should be `invite-cancelled` too (or, alternatively, everything else should be updated to match this one).

crwood (changes requested):
Thanks a lot for this. A few inline comments below.

docs/interface.rst
```diff
@@ -385,9 +385,29 @@ The follow event kinds are understood (see ``status.py`` for more details on the
 
 - ``"download-finished"``: same as upload version.
 
+- ``"invite-created"``: A new invite is created. All invite events also have a ``folder`` indicating which folder they pertain to, a ``id``, ``participant-name`` indicating the invitee and ``mode`` (``"read-only"`` or ``"read-write"``).
+
+- ``"invite-welcomed"``: Connection to the mailbox has succeeded. As well as the keys from ``"invite-created"`` the welcome data is contained in ``welcome`` (you should show users the ``"motd"`` if there is one).
+
+- ``"invite-code-created"``: As well as the keys from ``"invite-created"``, a ``code`` key will be included indicating the wormhole code.
+
+- ``"invite-versions"``: As well as the keys from ``"invite-created"``, a ``versions`` key is included indicating the peer's version information.
```
crwood, line 394: This event-kind follows a different naming-convention than the others (which use `object-verb`, in past-tense, throughout). Consider updating/changing this to match the existing convention (though, in this case, I'm not sure what would make the most sense -- "invite-version-received"? "invite-version-negotiated"? "invite-version-accepted"?).

docs/interface.rst
```diff
@@ -385,9 +385,29 @@ The follow event kinds are understood (see ``status.py`` for more details on the
...
+
 All timestamps are "seconds since the Unix epoch", as numbers (JSON only has "numbers" and doesn't distinguish floats from ints).
 
 Note that the first "update events" message received will _not_ contain all the updates to that point; it will synthesize the correct events to communicate the current state.
 For example, if there are 50 files in the folder and 48 have already been uploaded, there will be just 2 ``upload-queued`` events (because the other 48 have all finished already).
 If one of these files is currently being uploaded, there will also be a ``upload-started`` event.
 To know the state of all files, use the other endpoints.
+
+
+The *invite events* follow an ordered pattern: ``invite-created`` will be followed by ``invite-created``, ``invite-welcomed``, ``invite-code-created`` and then finalized by one of the "terminal" events (``invite-succeeded``, ``invite-failed``, ``invite-rejected`` or ``invite-cancelled``).
```
crwood, line 412: >``invite-created`` will be followed by ``invite-created``

This seems like an erroneous duplicate. Should the version-related event-kind be included here?

docs/interface.rst
meejah: This just means "we got the `versions` message from the mailbox" -- does that help with naming?

docs/interface.rst
meejah: yes, that should be whatever you want to call `invite-versions`

docs/interface.rst
crwood: Perhaps "invite-version-received" then?

docs/interface.rst
```diff
@@ -385,9 +385,32 @@ The follow event kinds are understood (see ``status.py`` for more details on the
...
 
+- ``"invite-created"``: A new invite is created locally.
+All invite events also have a ``folder`` indicating which folder they pertain to, a ``id``, ``participant-name`` indicating the invitee and ``mode`` (``"read-only"`` or ``"read-write"``).
+
+- ``"invite-welcomed"``: Connection to the mailbox has succeeded. As well as the keys from ``"invite-created"`` the welcome data is contained in ``welcome`` (you should show users the ``"motd"`` if there is one).
+
+- ``"invite-code-created"``: We have allocated what we need on the mailbox server and have a valid code.
+As well as the keys from ``"invite-created"``, a ``code`` key will be included indicating the wormhole code.
+
+- ``"invite-received-versions"``: We have received an encrypted message from our peer.
```
crwood, line 396: This is extreme bike-shedding (so feel free to ignore!) but "invite-versions-received" would be even more consistent with the other existing event names/kinds -- all of which end with with the past-tense verb (i.e., "-ed").

crwood (approved):
LGTM. Just one very bikeshed-y suggestion below but... I don't think it should hold anything up; feel free to address or ignore at your discretion. :)

PR #732, opened 2023-06-26
fix status updates on conflicted files

### LINKED ISSUE #731: Conflicted files and downloads
According to https://whetstone.private.storage/privatestorage/privatestoragedesktop/-/issues/791#note_30691 it appears that conflicted files are queued for download, but never complete.

They should either complete successfully (i.e. because we concluded "yes, still conflicted") or just not start in the first place (i.e. because we know it's already conflicted, so why download a Snapshot at all).

Beware that care may be required here: a file may  become conflicted with Participant A and then some time later _also_ become conflicted with Participant B -- we still want to detect that second case (while still under conflict to A). That is, we shouldn't download any additional updates for "A" after detecting the conflict, but we probably _do_ want to download updates from other participants (so we can record more conflicting participants, if they do indeed conflict).

meejah:
Fixes #731

### DIFF
--- a/integration/test_multiuser.py
+++ b/integration/test_multiuser.py
@@ -11,6 +11,7 @@ from .util import (
     await_file_contents,
     find_conflicts,
 )
+from twisted.internet.defer import DeferredList
 
 
 def non_lit_content(s):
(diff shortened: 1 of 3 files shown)


crwood (approved):
Tested and confirmed: with these changes I no longer observe the additional (and never starting/completing) "download-queued" event messages for conflicted files.

PR #736, opened 2023-06-30
invites and readonly

### LINKED ISSUE #734: invite read-only vs read-write and errors
Currently, if a device is invited as `read-write` but acts as if it was `read-only` (i.e. by responding without a Personal readcap) the magic-folder side ends up returning "wormhole failed or other side declined".

It should have a more-specific error-message (i.e. something about read-only or a protocol-violation or ...)

It could instead allow such a response. That is, allow such a "read-only" response to a read-write invite and "gracefully downgrade" to a read-only participant anyway -- since the "hard work" was producing, relaying and consuming the invite code.

meejah:
Fixes #734


meejah:
Local tools say there's a single uncovered line; no idea what codecov is doing here...

hacklschorsch (approved):
Nice!

PR #752, opened 2024-03-27
725.conflict resolution  rebase

### LINKED ISSUE #725: Conflict resolution
Currently, it is possible to detect and produce conflicts. These produce a database entry, are listed by the API and make "conflict" files in the filesystem.

There is not an API to _resolve_ these conflicts. We likely need two:

- an HTTP API to ask for a particular resolution;
- and a way to determine from the filesystem a resolution (i.e. by removing the conflict files).

See also #102

meejah:
Fixes #725 

(Work-in-progress). API, CLI and documentation for performing conflict resolution.

### DIFF
--- a/docs/conflicts.rst
+++ b/docs/conflicts.rst
@@ -6,10 +6,10 @@ Magic Folder Conflicts and Resolution
 =====================================
 
 When we have two or more participants updating a folder, it can happen that a file is modified "at the same time" by two or more participants.
-More correctly, "at the same time" here means "between communications".
+More correctly, "at the same time" here means "between communications events".
 
-Effectively, participants are speaking a protocol by posting messages to their mutable Capabilities.
-Namely, updading which Snapshot object a particular (file) name points to.
+Effectively, participants are speaking a protocol by posting updates to their mutable Capabilities.
+Namely, updating which Snapshot object a particular (file) name points to (including adding a new name, or removing one).
 
 We do not know *when* another Participant has updated their notion of the current state.
 However, what we *can* notice is when they "commit" to that state (that is, they update publically their current Snapshot pointer).
@@ -49,11 +49,12 @@ Either our Snapshot appears in some parents (including grandparents, etc) of the
 
 If it does appear, this is an "update" (and we simply replace the local content with the incoming content).
 
-Instead if our Snapshot does not appear as any ancestor of the incoming Snapshot, a commit is determined.
+Instead if our Snapshot does not appear as any ancestor of the incoming Snapshot, a conflict is determined.
 
 This is because when the other device created their Snapshot, they didn't know about ours (or else it would appear as an ancestor) so we have made a change "at the same time".
 
 Note that unlike tools like Git, we do not examine the contents of the file or try to produce differences -- everything is determined from the metadata in the Snapshot.
+This means that even if we happened to make the very same edits "at the same time" it would still be a conflict.
 
 
 Showing Conflicts Locally
@@ -84,7 +85,7 @@ Resolving via the CLI
 ---------------------
 
 The subcommand ``magic-folder resolve`` may be used to specify a resolution.
-It allows you to choose ``--mine`` or ``--theirs``(if there is only one other conflict).
+It allows you to choose ``--mine`` or ``--theirs`` (if there is only one other conflict).
 Otherwise, you must use ``--use <name>`` to specify which version to keep.
 
 Currently there is no API for doing something more complex (e.g. simultaneuously replacing the latest version with new content).
--- a/docs/conflicts.rst
+++ b/docs/conflicts.rst
@@ -83,6 +83,18 @@ Resolving via Filesystem
 Resolving via the CLI
 ---------------------
 
+The subcommand ``magic-folder resolve`` may be used to specify a resolution.
+It allows you to choose ``--mine`` or ``--theirs``(if there is only one other conflict).
+Otherwise, you must use ``--use <name>`` to specify which version to keep.
+
+Currently there is no API for doing something more complex (e.g. simultaneuously replacing the latest version with new content).
+
+Complete example:
+
+.. code-block:: console
+
+    magic-folder resolve --mine ~/Documents/Magic/foo
+
 
 Resolving via the HTTP API
 --------------------------
--- a/docs/conflicts.rst
+++ b/docs/conflicts.rst
@@ -1,0 +1,89 @@
+.. -*- coding: utf-8 -*-
+
+.. _conflicts:
+
+Magic Folder Conflicts and Resolution
+=====================================
+
+When we have two or more participants updating a folder, it can happen that a file is modified "at the same time" by two or more participants.
+More correctly, "at the same time" here means "between communications".
+
+Effectively, participants are speaking a protocol by posting messages to their mutable Capabilities.
+Namely, updading which Snapshot object a particular (file) name points to.
+
+We do not know *when* another Participant has updated their notion of the current state.
+However, what we *can* notice is when they "commit" to that state (that is, they update publically their current Snapshot pointer).
+
+Each file is considered individually, and is represented by a tree of Snapshots of the contents (plus metadata).
+Included in the metadata are one or more "parent" Snapshots (a brand new file has zero parents).
+
+For more on this, the :ref:`datamodel` has a section on :ref:`datamodel_conflicts`.
+
+
+Communication Between Folders
+-----------------------------
+
+Consider a single file named ``"foo"``.
+When this file is created (for example on the device called Laptop) that device uploads a Snapshot (``S0``) with no parent Snapshots.
+
+By "upload" we mean to push the content and metadata to the Tahoe-LAFS client, receiving ultimately a Capability for the Snapshot ``S0``.
+The file entry for ``"foo"`` is then updated, visualized as: ``foo -> S0``
+
+Now, all other Participants can (and, eventually, will) notice this update when they poll the magic-folder.
+
+Each of these is a "communication event", so we talk about the regions between these as a cohesive state of that client.
+That is, anything that happens during that time is "at the same time" for the purposes of this protocol.
+
+Note that this time interval could be as short as a few seconds or as long as days (or more).
+
+Observe too that the "parents" of a particular Snapshot are a commitment as to the state visible to that client when creating the Snapshot.
+
+
+Detecting Conflicts
+-------------------
+
+A new update is communicated to us -- that is, we've downloaded a previously-unknown Snapshot from some other Participant's mutable Capbility.
+
+We now look at our own Snapshot for the corresponding file.
+Either our Snapshot appears in some parents (including grandparents, etc) of the new Snapshot, or it doesn't.
+
+If it does appear, this is an "update" (and we simply replace the local content with the incoming content).
+
+Instead if our Snapshot does not appear as any ancestor of the incoming Snapshot, a commit is determined.
+
+This is because when the other device created their Snapshot, they didn't know about ours (or else it would appear as an ancestor) so we have made a change "at the same time".
+
+Note that unlike tools like Git, we do not examine the contents of the file or try to produce differences -- everything is determined from the metadata in the Snapshot.
+
+
+Showing Conflicts Locally
+-------------------------
+
+Once a conflict is detected, "conflict marker" files are put into the local magic folder location (our local file remains unmodified, and something like ``foo.conflict-desktop`` will appear.
+The state database is also updated (conflicts can also be listed via the API and CLI)
+
+
+Resolving Conflicts
+-------------------
+
+We cannot "magically" solve a conflict: two devices produced new Snapshots while not communicating with each other.
+
+Thus, it is up to the humans using these devices to determine what happens.
+Once appropriate changes are decided upon, a new Snapshot is produced with *two or more parents*: one parent for each of the Snapshots involved in the conflict (we've only talked about one other participant so far, but there could be more).
+
+Such a Snapshot (with two or more parents) indicates to the other clients a particular resoltion to the conflict has been decided.
+
+So there's another case when we see an incoming new Snapshot: it may in fact *resolve an existing* conflict.
+
+
+Resolving via Filesystem
+------------------------
+
+
+Resolving via the CLI
+---------------------
+
+
+Resolving via the HTTP API
+--------------------------
+
--- a/docs/conflicts.rst
+++ b/docs/conflicts.rst
@@ -87,3 +87,4 @@ Resolving via the CLI
 Resolving via the HTTP API
 --------------------------
 
+See :ref:`api_resolve_conflict`
--- a/docs/interface.rst
+++ b/docs/interface.rst
@@ -262,6 +262,7 @@ Each item in the ``dict`` maps a relpath to a list of author-names.
 The author-names correspond to the device that conflicts with this file.
 There will also be a file named like ``<relpath>.conflict-<author-name>`` in the magic-folder whose contents match those of the conflicting remote file.
 
+.. _api_resolve_conflict:
 
 POST ``/v1/magic-folder/<folder-name>/resolve-conflict``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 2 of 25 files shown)


docs/conflicts.rst
```diff
@@ -0,0 +1,131 @@
...
+
+Communication Between Folders
+-----------------------------
+
+Consider a single file named ``"foo"``.
+When this file is created (for example on the device called Laptop) that device uploads a Snapshot (``S0``) with no parent Snapshots.
+
+By "upload" we mean to push the content and metadata to the Tahoe-LAFS client, receiving ultimately a Capability for the Snapshot ``S0``.
+The file entry for ``"foo"`` is then updated, visualized as: ``foo -> S0``
+Here, "updated" meanst that we use Tahoe-LAFS to edit to contents of our Personal directory to change "foo" to point at Snapshot ``S0``.
```
crwood, line 31: Typo: "meanst" -> "means"

docs/conflicts.rst
```diff
@@ -0,0 +1,131 @@
...
+
+Note that this time interval could be as short as a few seconds or as long as days (or more).
+
+Observe too that the "parents" of a particular Snapshot are a commitment to the state visible by a particular client when creating any new Snapshot.
+
+
+Detecting Conflicts
+-------------------
+
+A new update is communicated to us -- that is, we've downloaded a previously-unknown Snapshot from some other Participant's mutable Capbility.
```
crwood, line 46: Typo: "Capbility" -> "capability"

docs/conflicts.rst
```diff
@@ -0,0 +1,131 @@
...
+Either our Snapshot appears in some parents (including grandparents, etc) of the new Snapshot, or it doesn't.
+
+If it does appear, this is an "update" (and we simply replace the local content with the incoming content).
+
+Instead if our Snapshot does not appear as any ancestor of the incoming Snapshot, a conflict is determined.
+
+This is because when the other device created their Snapshot, they didn't know about ours (or else it would appear as an ancestor) so we have made a change "at the same time".
+
+Note that unlike tools like Git, we do not examine the contents of the file or try to produce differences -- everything is determined from the metadata in the Snapshot.
+This means that even if we happened to make the very same edits "at the same time" it would still be a conflict.
```
crwood, line 58: Curious: Is this true because, presumably, the capability strings would differ (on the assumption that the parties aren't sharing the convergence secret)? What if they are sharing the a convergence secret (and somehow manage to push at the same time)?

(I'm mostly just wondering whether this behavior -- of "it would still be a conflict" -- is intended or accidental here).

docs/conflicts.rst
```diff
@@ -0,0 +1,131 @@
...
+
+
+Showing Conflicts Locally
+-------------------------
+
+Once a conflict is detected, "conflict marker" files are put into the local magic folder location (our local file remains unmodified, and something like ``foo.conflict-desktop`` will appear.
+The state database is also updated (conflicts can also be listed via the API and CLI).
+
```
crwood, line 66: Will such detected conflicts also be announced via the "events" API? If so, consider mentioning it here.

docs/conflicts.rst
```diff
@@ -0,0 +1,131 @@
...
+Resolving Conflicts
+-------------------
+
+We cannot "magically" solve a conflict: two devices produced new Snapshots while not communicating with each other.
+
+Thus, it is up to the humans using these devices to determine what happens.
+Once appropriate changes are decided upon, a new Snapshot is produced with *two or more parents*: one parent for each of the Snapshots involved in the conflict (we've only talked about one other participant so far, but there could be more).
+
+Such a Snapshot (with two or more parents) indicates to the other clients a particular resoltion to the conflict has been decided.
```
crwood, line 78: Like the above: can this be announced via the "events" API?

docs/conflicts.rst
```diff
@@ -0,0 +1,131 @@
...
+If this is the case, conflict markers are removed and the local database is updated (i.e. removing the conflict).
+
+It is a human problem if this resolution is not to your particular liking; you can produce an edit again or talk to the human who runs the other computer(s) involved.
+The history of these changes *is available* in the parent (or grandparent) Snapshots if the UX you're using can view or restore these.
+
+
+Resolving via Filesystem
+------------------------
+
+Not currently possible.
```
crwood, line 90: This seems important. A lot of other folder-sync tools expect/require users to resolve conflicts via the filesystem (e.g., by renaming one of the local/marked copies to the "real" filename and deleting the other marked copies) and so users might be naturally inclined to attempt to resolve conflicts this way. Is there another good way/place to emphasize that this won't work for users who might not read/see this document?

Beyond that, are there any plans to support this method of conflict resolution? "Not _currently_ possible" might imply that it could be added later but there's no indication of importance, priority, etc. If there's a ticket relating to this, consider linking it here.

docs/conflicts.rst
```diff
@@ -0,0 +1,131 @@
...
+
+
+Resolving via the CLI
+---------------------
+
+The subcommand ``magic-folder resolve`` may be used to specify a resolution.
+It allows you to choose ``--mine`` or ``--theirs`` (if there is only one other conflict).
+Otherwise, you must apply the ``--use <name>`` option to specify which version to keep.
+
+Currently there is no API for doing something more complex (e.g. simultaneuously replacing the latest version with new content).
```
crwood, line 100: Maybe a link to the existing API docs would be useful here (e.g., "Developers can use the .../resolve-conflict endpoint for resolving simple cases like the above programmatically; see `interface.rst`)?

docs/conflicts.rst
```diff
@@ -0,0 +1,131 @@
...
+--------------------------
+
+See :ref:`api_resolve_conflict`
+
+
+Future Directions
+-----------------
+
+We do not consider the current conflict functionality "done".
+There are other features required to make this more robust and have a nicer user experience.
```
crwood, line 119: Minor: maybe that terminating period (".") should be a colon (":") to signify that you're hereby listing those other features.

docs/conflicts.rst
```diff
@@ -0,0 +1,131 @@
...
+
+*Viewing old data*: While it is currently possible in the datamodel to view past versions of the files, we do not know of any UI that does this (and the CLI currently cannot).
+
+*Restore old version*: Similarly, it is possible to produce a new Snapshot that effectively restores an older version of the same file.
+We do not know of any UI that can do this.
+
+*Completely new content*: As hinted above, it might be nice to be able to produce a resolution that is some combination of multiple versions (like one sometimes does with Git conflicts, for example).
```
crwood, line 126: In this context, I wonder also about pruning old snapshots (e.g., to save costs on metered/ZKAPs-enabled grids) and/or requesting deletion (e.g., in cases where a file might have been published unintentionally).

(These are somewhat of beyond just _conflict_ detection though...)

docs/interface.rst
```diff
@@ -262,6 +262,22 @@ Each item in the ``dict`` maps a relpath to a list of author-names.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Ask for a particular resolution to a conflicted file.
+The body is a JSON object containing the keys:
+
+* ``relpath``: the file name inside the folder
+* ``take``: ``"mine"`` or ``"theirs"`` indicating whose version to take
+* ``use``:  a Participant name indicating whose version to take
+
+Note that you can use only one of ``take`` or ``use``, and that ``take="theirs"`` only works when there is exactly one other conflict.
```
crwood, line 277: What actually happens (or is expected to happen) if both are specified? Will users of the API get a 500 error? A response containing an error message in the body? Something else?

docs/conflicts.rst
meejah: Doesn't matter if the convergent secret is shared or not, the Snapshot is a new immutable and will be different. Yes, it is intended; if a client makes a new Snapshot, it is a new version whether or not the actual content is identical.

docs/conflicts.rst
meejah: No, there are no events related to conflicts currently

docs/conflicts.rst
meejah: What exactly should be announced? (Probably best expressed as an enhancement ticket / use-case)

docs/conflicts.rst
meejah: There is no deletion API in Tahoe-LAFS.

The closest there is (which would work for "pruning" too) is to  simply stop refreshing leases on things you no longer care about.

docs/interface.rst
meejah: You get an `_InputError` which is rendered as a BAD_REQUEST like all other input validation errors in the API.

docs/conflicts.rst
meejah: (I think this is more directly related to "history" / "how the datamodel works" than anything about conflicts per se though)

docs/conflicts.rst
meejah: #725 mentions it, but there's not a separate ticket.

The note isn't meant to mean this _isn't_ important, but there's not even a great description of what it should do, and this PR is already well into "probably too big" territory so I certainly don't want to add a whole bunch more to it.

docs/conflicts.rst
meejah: The API also doesn't allow for the complex thing, but you mean pointing out that the HTTP API covers everything the CLI does?

meejah:
@crwood can you write your desired "event" behavior as a ticket, please?

Other than that I think this is mergeable?

docs/conflicts.rst
crwood: >I think this is more directly related to "history" / "how the datamodel works" than anything about conflicts per se though

Yes, agreed (and hence my note/admission above that this is "beyond just conflict detection" -- so no need to really do anything here/now); this was mostly just an observation about something that I -- as a hypothetical/imaginary end user of this software -- might wonder about when encountering such a list of missing features. To that end, a note that there is no way to prune/delete snapshots _might_ be nice here (depending on what level of knowledge the reader is assumed to have) but I also grant that a) reliable "deletion" isn't (ever?) possible, b) if there ever were deletion-like functionality, it would probably be a Tahoe-LAFS thing -- not a Magic-Folder thing, and c) this is more of a general concern (that isn't specific to _conflicts_). So feel free to ignore/disregard/resolve. :)

docs/conflicts.rst
crwood: Ticket filed: https://github.com/tahoe-lafs/magic-folder/issues/755

src/magic_folder/magic_file.py
```diff
@@ -73,7 +73,7 @@ class MagicFileFactory(object):
     _config = attr.ib()  # MagicFolderConfig
     _tahoe_client = attr.ib()
     _folder_status = attr.ib()
-    _local_snapshot_service = attr.ib()
+    _local_snapshot_service = attr.ib()#validator=attr.validators.instance_of(LocalSnapshotService))
```
crwood, line 76: Is this an actual pragma-like directive for `attrs`, or something commented-out (that should maybe be removed)? 

I haven't used `attrs` much, so not sure what the actual intent is here -- or why the validator is outside (and not inside) `attr.ib()`'s parens.

src/magic_folder/magic_file.py
```diff
@@ -194,6 +214,9 @@ class MagicFile(object):
     _queue_remote = attr.ib(default=attr.Factory(list))
     _is_working = attr.ib(default=None)
 
+    # XXX
```
crwood, line 217: What is this "XXX" supposed to signify? Should this be expanded into a comment? If not, remove.

src/magic_folder/magic_file.py
```diff
@@ -1286,15 +1389,28 @@ def _done_working(self):
...
+        outputs=[_working, _status_download_queued, _begin_download]
     )
+    # XXX this is trix-y .. the user could be changing the local
+    # contents because they'll ultimately be doing a "resolve" (with
+    # those contents) __OR__ they could be naively simply editing the
+    # file more.
+    #
+    # ...so I think we _do_ want to queue a local-update: an incoming
+    # resolution may happen, and if we have no local change: great,
+    # apply it. But if we _do_ have a local change, we'd want to
```
crwood, line 1408: This sounds about right to me.

In such instances, though, I wonder if would it be worth perhaps uniquely identifying each individual "instance" of conflict -- perhaps associating each instance with a UUID or something -- so that we know or can track from which instance of conflict was ultimately resolved (and "conflict again" only if the other instance(s) were resolved)?

I was thinking about this originally in the context of https://github.com/tahoe-lafs/magic-folder/issues/755, and wondering if, for example, it might make sense to have a "conflict-_updated_" event (in order to signal, say, that another change has happened on top of an already-conflicted file that did _not_ resolve the conflict, or maybe to signal that a two-participant conflict has become a three-participant one -- e.g., going from `"participants": ["Alice", "Bob"]` to `"participants": ["Alice", "Bob", "Charlie"]`) versus, instead, just emitting another "conflict-_detected_" message (and considering all conflicts associated with a given `relpath` as being "resolved" if/when one of them is).

(I don't have particularly strong opinions here as I haven't thought very far past the simpler "two participants, one conflict" case...)

src/magic_folder/magic_file.py
```diff
@@ -1286,15 +1389,28 @@ def _done_working(self):
...
+    # .. conflict again?
+    #
+    # need integration test here: have conflict; A changes local file
```
crwood, line 1411: Should such a test be added to this PR? (Or captured in a separate ticket?)

src/magic_folder/test/test_magic_file.py
```diff
@@ -324,10 +326,136 @@ def test_multiple_remote_updates(self):
...
+        yield DeferredList([d0, d1])
+
+        self.assertEquals(
+            self.config.list_conflicts_for("dual-conflict"),
+            [Conflict(snapshot_cap=cap1, participant_name="beth")]
+        )
+
+        self.uploader._uploads = ["dual-conflict"]
+        snap = yield mf.resolve_conflict(Conflict(snapshot_cap=cap1, participant_name="beth"))
+        print("XXX", snap)
```
crwood, line 390: Should this print statement be removed? If it's intended to be preserved, consider using more descriptive string here.

crwood (approved):
LGTM.

I left a few additional inline comments below -- mostly minor or speculative. Feel free to address as you see fit (or not!) and merge.

src/magic_folder/magic_file.py
meejah: there's no pragmas, that's like "note to self" (because circular import :/ )

src/magic_folder/magic_file.py
meejah: we do already track multiple "instances" in the sense that we track which other participants are currently conflicted (which can be 0..n)

the only case i can think of might be where e.g. "alice" is conflicted, and then updates her copy further (so on "our" side we'd still be conflicted).

Is there anything that should behave differently in this case? (Clients _could_ I think already figure out where the conflict "first" happened if they wanted, by examining the version tree).

crwood: >the only case i can think of might be where e.g. "alice" is conflicted, and then updates her copy further (so on "our" side we'd still be conflicted).

It also still seems possible (I believe?) to have participants resolve a conflict in diverging ways that would require additional intervention. For example, Alice might accept/prefer Bob's copy, while Charlie ("at the same time") accepts/prefers Alice's original copy (i.e., prior to her accepting of Bob's).

I suppose the questions here are more UX-oriented ones -- about whether and how users need to be exposed (and/or make decisions about) each "intermediate" instance in a multi-conflict scenario. Presumably, whichever snapshots are "the latest" probably matter most here (especially since past snapshots are preserved in the event they are needed).

Anyway, I don't think there's anything that needs to be done about this right now; sorry for the digression!

meejah: Yes, two resolutions "at the same time" are possible -- they should immediately produce another conflict (probably need _lots_ more tests like this :thinking: ). However, this isn't logically "another instance" of an existing conflict, but a new conflict (between the Snapshot that Alice produced and the Snapshot that Charlie produced as resolutions).

src/magic_folder/magic_file.py
meejah: https://github.com/tahoe-lafs/magic-folder/issues/756

PR #762, opened 2024-11-28
Poke CI (NOT FOR MERGE)

crwood:
Just poking CI to see whether the integration test failures seen in https://github.com/tahoe-lafs/magic-folder/pull/761 will  trigger independently...


meejah:
FYI, to upgrade to artifacts v4 we also need to make the artifact names unique.

crwood:
>FYI, to upgrade to artifacts v4 we also need to make the artifact names unique.

Yes, let me see if I can fix that..

crwood:
Okay, I've fixed the artifact name-conflicts in https://github.com/tahoe-lafs/magic-folder/pull/761

PR #763, opened 2024-12-03
update CI

meejah:


meejah:
Fun! That looks like a real (and very weird) bug on only-master: one of the 20 folders gets a differently-sized metadata Immutable.

```
E       testtools.matchers._impl.MismatchError: Differences: [
E       len({(416, 292), (416, 190)}) != 1: after <function <lambda>> on [[416, 800, 190], [416, 800, 190], [416, 800, 190], [416, 800, 190], [416, 800, 190], [416, 800, 190], [416, 800, 190], [416, 800, 190], [416, 800, 190], [416, 800, 190], [416, 800, 190], [416, 800, 190], [416, 800, 190], [416, 800, 190], [416, 800, 190], [416, 800, 190], [416, 800, 190], [416, 800, 190], [416, 800, 190], [416, 800, 292]]
E       ]
```

PR #768, opened 2025-01-24
README: `init` not `create`

### LINKED ISSUE #769: README refers to non-existing command
The README tells me to create a Magic Folder using the `create` command, but when I try that, MF tells me that command would not exist: `Error: Unknown command: create`

`magic-folder --version` returns "Magic Folder version 24.3.1.dev49+g993f1ab4".

hacklschorsch:
I guess this should be `magic-folder init` not `magic-folder create`, since the latter gives me `Error: Unknown command: create` (`magic-folder --version` returns "Magic Folder version 24.3.1.dev49+g993f1ab4")

Fixes #769.


meejah (approved):
Sounds plausible / LGTM (haven't used the CLI for a while).

Will need a "newsfragment" like in tahoe-lafs (is why the codechecks is failing).

